PR #36, opened 2025-02-01
Simplify an expression which triggers complexity errors in nightly toolchains

owenv:


owenv:
@swift-ci test

jakepetroules (approved):

owenv:
This change resolves the build issue, we fail later on in the process now. Merging

PR #39, opened 2025-02-01
Rely on autolinking for Swift Testing dependency of test support libraries

owenv:
Attempt to work around a build issue with the macOS nightlies


owenv:
@swift-ci test

owenv:
Confirmed this resolves macOS build issues, although we have test issues which still need to be resolved

PR #40, opened 2025-02-01
[doc] Fix SWBBuildSystem's hierarchy in the architecture doc

kateinoigakukun:
I think the layering hierarchy of `SWBBuildSystem` might be wrong in the architecture document.


owenv:
We are still in the process of sorting out some CI issues, but I'll merge this since it's limited to documentation changes

Issue #16, opened 2025-01-30
Remove appIDHasFeaturesEnabled


jakepetroules:
This property has not been needed for many years. Any support code necessary to compute that value can be removed from the codebase.

The appIDHasFeaturesEnabled in SWBProvisioningTaskInputsSourceData is public API however, and will need to remain as a deprecated property to maintain ABI compatibility.

Issue #2, opened 2025-01-28
Allow plugins to contribute BuiltinMacros


owenv:
Currently, all of the builtin macros used by SwiftBuild are centralized in BuiltinMacros.swift. We should update the build settings initialization process to allow defining these in plugins as well, so that macros associated with a feature can live in the plugin which implements this feature.

PR #41, opened 2025-02-01
Remove appIDHasFeaturesEnabled

ekeege:
Closes: https://github.com/swiftlang/swift-build/issues/16


Sources/SWBProtocol/PlanningOperationMessages.swift
```diff
@@ -120,16 +118,15 @@ public struct ProvisioningTaskInputsSourceData: Serializable, Equatable, Sendabl
...
+        self.init(configurationName: configurationName, provisioningProfileSupport: provisioningProfileSupport, provisioningProfileSpecifier: provisioningProfileSpecifier, provisioningProfileUUID: provisioningProfileUUID, provisioningStyle: provisioningStyle, teamID: teamID, bundleIdentifier: bundleIdentifier, productTypeEntitlements: productTypeEntitlements, productTypeIdentifier: productTypeIdentifier, projectEntitlementsFile: projectEntitlementsFile, projectEntitlements: projectEntitlements, signingCertificateIdentifier: signingCertificateIdentifier, signingRequiresTeam: signingRequiresTeam, sdkRoot: sdkRoot, sdkVariant: sdkVariant, supportsEntitlements: supportsEntitlements, wantsBaseEntitlementInjection: wantsBaseEntitlementInjection, entitlementsDestination: entitlementsDestination, localSigningStyle: localSigningStyle, enableCloudSigning: false)
     }
 
     public init(from deserializer: any Deserializer) throws {
         let count = try deserializer.beginAggregate(20...22)
         self.configurationName = try deserializer.deserialize()
-        self.appIDHasFeaturesEnabled = try deserializer.deserialize()
```
owenv, line 132: Here, we should check the count of fields received to determine whether appIDHasFeaturesEnabled is still present - if it is, we can immediately discard it

Sources/SWBProtocol/PlanningOperationMessages.swift
```diff
@@ -157,7 +154,6 @@ public struct ProvisioningTaskInputsSourceData: Serializable, Equatable, Sendabl
     public func serialize<T: Serializer>(to serializer: T) {
         serializer.serializeAggregate(21) {
```
owenv, line 155: This count should be updated since one of the serialized fields was deleted

Sources/SWBProtocol/ProjectModel/ProvisioningSourceData.swift
```diff
@@ -67,7 +49,6 @@ extension ProvisioningSourceData: PendingSerializableCodable {
     public func legacySerialize<T: Serializer>(to serializer: T) {
         serializer.serializeAggregate(4) {
```
owenv, line 50: Similarly, the field count here should be updated

Sources/SWBProtocol/ProjectModel/ProvisioningSourceData.swift
```diff
@@ -76,7 +57,6 @@ extension ProvisioningSourceData: PendingSerializableCodable {
     public init(fromLegacy deserializer: any Deserializer) throws {
         let count = try deserializer.beginAggregate(4...5)
         self.configurationName = try deserializer.deserialize()
-        self.appIDHasFeaturesEnabled = try deserializer.deserialize()
```
owenv, line 79: Similar to the above, we should check the count here to determine if we need to deserialize and then discard the now-unused field

owenv (changes requested):
Thank you, this is a nice cleanup! I'll leave a few comments below where I think this change needs a few compatibility fixes to the message serialization format.

For context:
Generally speaking there are two types of compatibility we care about in Swift Build:
1. ABI compatibility of the SwiftBuild module, which clients link directly
2. Serialization format compatibility for messages passed between the SwiftBuild module and SWBBuildService, which may either be linked into the client or running in a separate process.

For case #2, we want to ensure that older versions of SwiftBuild are still able to communicate with SWBBuildService. That means that SWBBuildService should still tolerate receiving messages with the appIDHasFeaturesEnabled field, even though newer versions of the API will not include it.

Since the message in this PR is used by the client to communicate code-signing information to the build system, the `swift package launch-xcode --disable-sandbox` command described in the README is a good way to test it manually - serialization issues should generally appear as failures in the build log.

jakepetroules:
@ekeege Are you still planning to finish this or would you prefer someone else take it up?

owenv:
thanks!
@swift-ci test

owenv:
@swift-ci test macOS

owenv:
@swift-ci test Windows

owenv:
@swift-ci test Windows

owenv (approved):

jakepetroules (approved):

PR #44, opened 2025-02-01
Windows Build Fixes

compnerd:
Add a few module build fixes for Windows allowing this code to build on Windows with SPM.

### DIFF
--- a/Sources/SWBCore/KnownFolders.swift
+++ b/Sources/SWBCore/KnownFolders.swift
@@ -16,6 +16,14 @@ import Foundation
 #if os(Windows)
 import WinSDK
 
+private var KF_FLAG_DEFAULT: DWORD {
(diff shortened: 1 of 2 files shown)


owenv:
Interesting, I haven't seen this when doing my own windows builds, but I may have been using a different compiler / visual studio (I don't have the version numbers at the moment). That said, this looks like a safe and uncontroversial change

owenv (approved):

jakepetroules (commented):
Same as Owen, any idea why this would work for us but not for you?

compnerd:
Interesting, I suspect it is related to compiler versions. This enum has always been strongly typed. I was building with 6.2-dev I think?

jakepetroules:
Ah, I think the latest I had tried was 6.1.

Issue #3, opened 2025-01-28
Allow using swiftc as the Swift Build linker driver


owenv:
Swift Build currently always uses clang as the linker driver, while SwiftPM has historically used swiftc. We should add a build setting clients can use to opt into using swiftc as the linker driver to ensure consistency where the two drivers have divergent behavior / command line flags. This will likely require some changes throughout the build system where we currently make assumptions about flags the linker driver accepts, and especially how we escape flags to pass them to the underlying linker.

compnerd:
@jakepetroules I think that we should continue to use `clang` as the linker driver for the ASM case. However, the important thing is the middle two cases you identified - the failure of SPM to do so has been the reason that we need to overlink on Windows and explicitly link the swift standard library on all targets as `swiftc` will force the dependency on the swift standard library.

jakepetroules:
Why clang for the pure ASM case? Wouldn't that potentially pull in unexpected dependencies?

compnerd:
It is actually the inverse - it drops unnecessary dependencies - use of `swiftc` would pull in the swift standard library.

jakepetroules:
Sorry, I meant why use clang instead of ld for the pure ASM case? Wouldn't that pull in libc when you might not want it?

PR #51, opened 2025-02-02
Add WebAssembly platform to Swift build system based on Swift SDK

kateinoigakukun:
This change adds a new platform, `webassembly`, to the Swift build system. The platform is based on the existing Swift SDK artifact bundles, which can be installed by SwiftPM.
To build your Xcode project for WebAssembly, you need to install a Swift SDK for WebAssembly and Swift OSS toolchain that supports WebAssembly compilation target, then set the `SDKROOT`, `SUPPORTED_PLATFORMS` and `TOOLCHAINS` macros in your Xcode project file like below:

```
SDKROOT = DEVELOPMENT-SNAPSHOT-2025-01-11-a-wasm32-unknown-wasip1-threads
SUPPORTED_PLATFORMS = webassembly
TOOLCHAINS = org.swift.62202501101a
```


Sources/SWBWebAssemblyPlatform/SwiftSDK.swift
MaxDesiatov, line 1: This is not Wasm-specific, so maybe it belongs somewhere else? IIUC it literally vendors SwiftPM code.

MaxDesiatov (approved):

Sources/SWBWebAssemblyPlatform/SwiftSDK.swift
owenv: SWBCore might be a good place for this. It would be nice to have these models in a shared location that Swift Build and SwiftPM could share, but it might be a bit early to attempt that

Tests/SWBWebAssemblyPlatformTests/SWBWebAssemblyPlatformTests.swift
```diff
@@ -0,0 +1,119 @@
...
+                        dependencies: [
+                            "staticlib",
+                        ]
+                    ),
+                    TestStandardTarget(
+                        "staticlib",
+                        type: .staticLibrary,
+                        buildConfigurations: [
+                            TestBuildConfiguration("Debug", buildSettings: [
+                                // FIXME: Find a way to make these default
```
owenv, line 70: WebAssembly should be able to provide spec files for the library product types which override the generic copies and give this a default value. But I think that would be fine to do as a follow up later

owenv (commented):
Thanks, this is really great! I left some comments below, most of them are pretty minor.

One thing I want to explore is if we can generate the Swift Build `SDK` and `Platform` in an entirely data-driven manner from the Swift SDK, so that this support that you've added can be generalized to other Swift SDKs. But https://github.com/swiftlang/swift-build/issues/3 and a few other issues will need to be sorted out before that's possible, and I don't want to block your work here in the meantime. Maybe it would make sense to leave comments where we're currently hardcoding information about WebAssembly that's not in the Swift SDK, and as we eliminate this cases we could rename this plugin to something like `SWBSwiftSDKPlatform`?

Sources/SWBTestSupport/RunDestinationTestSupport.swift
```diff
@@ -268,6 +268,11 @@ extension _RunDestinationInfo {
     package static var qnx: Self {
         return .init(platform: "qnx", sdk: "qnx", sdkVariant: "qnx", targetArchitecture: "undefined_arch", supportedArchitectures: ["aarch64", "x86_64"], disableOnlyActiveArch: true)
     }
+
+    /// A run destination targeting WebAssembly/WASI generic device, using the public SDK.
+    package static var wasm: Self {
+        return .init(platform: "webassembly", sdk: "webassembly", sdkVariant: "webassembly", targetArchitecture: "wasm32", supportedArchitectures: ["wasm32"], disableOnlyActiveArch: true)
```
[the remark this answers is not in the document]
jakepetroules: There's always an implicit variant; sdkVariant follows the SDK name for everything except Mac Catalyst. So keeping it as `webassembly` would be consistent with what we have already.

Sources/SWBCore/Specs/Tools/LinkerTools.swift
```diff
@@ -293,7 +293,7 @@ public final class LdLinkerSpec : GenericLinkerSpec, SpecIdentifierType {
...
         let supportsSwiftConcurrencyNatively = cbc.producer.platform?.supportsSwiftConcurrencyNatively(cbc.scope, forceNextMajorVersion: false, considerTargetDeviceOSVersion: false) ?? true
         let shouldEmitRPathForSwiftConcurrency = UserDefaults.allowRuntimeSearchPathAdditionForSwiftConcurrency && !supportsSwiftConcurrencyNatively
-        if (cbc.producer.platform?.supportsSwiftInTheOS(cbc.scope, forceNextMajorVersion: true, considerTargetDeviceOSVersion: false) != true || cbc.producer.toolchains.usesSwiftOpenSourceToolchain || shouldEmitRPathForSwiftConcurrency) && isUsingSwift {
+        if (cbc.producer.platform?.supportsSwiftInTheOS(cbc.scope, forceNextMajorVersion: true, considerTargetDeviceOSVersion: false) != true || cbc.producer.toolchains.usesSwiftOpenSourceToolchain || shouldEmitRPathForSwiftConcurrency) && isUsingSwift && cbc.producer.platform?.identifier != "webassembly" {
```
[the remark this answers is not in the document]
jakepetroules: But there's also `|| cbc.producer.toolchains.usesSwiftOpenSourceToolchain`

Sources/SWBWebAssemblyPlatform/Plugin.swift
```diff
@@ -0,0 +1,144 @@
...
+    manager.register(WebAssemblySDKRegistryExtension(), type: SDKRegistryExtensionPoint.self)
+}
+
+struct WebAssemblyPlatformSpecsExtension: SpecificationsExtension {
+    func specificationFiles() -> Bundle? {
+        .module
+    }
+
+    func specificationDomains() -> [String : [String]] {
+        ["webassembly": ["generic-unix"]]
```
jakepetroules, line 30: generic-unix is the base domain for non-Apple POSIX-y/UNIX-y ELF-based types of platforms. The main intent here is to share compiler and linker specs for clang and GNU/BSD linkers where the flags are generally the same across any platforms which use it (e.g. Linux and FreeBSD both inherit this).

I'm not sure if WASI fits that bill (I'm not saying it's wrong either), just want to get a sense of how similar the compiler and linker invocations are, vs whether it makes more sense to inherit from the global domain instead, and define the bits from scratch.

Sources/SWBWebAssemblyPlatform/Plugin.swift
```diff
@@ -0,0 +1,144 @@
...
+        let wasmSwiftSDKs = (try? SwiftSDK.findSDKs(
+            targetTriples: Array(supportedTriples.keys),
+            fs: localFS
+        )) ?? []
+
+        var wasmSDKs: [(path: Path, platform: SWBCore.Platform?, data: [String: PropertyListItem])] = []
+
+        for wasmSDK in wasmSwiftSDKs {
+            for (triple, tripleProperties) in wasmSDK.targetTriples {
+                guard let (arch, os, env) = supportedTriples[triple] else {
```
jakepetroules, line 104: Interesting approach, defining a dedicated SDK for each triple. I wonder if we should handle this somewhat differently, by adding all the triples to a single SDK and then having dedicated build settings that control which ABI variant you're building against?

@owenv, any thoughts here?

Tests/SWBWebAssemblyPlatformTests/SWBWebAssemblyPlatformTests.swift
```diff
@@ -0,0 +1,119 @@
...
+        guard let sdkroot = getEnvironmentVariable("WASM_SDKROOT") else { return }
+        guard let toolchain = getEnvironmentVariable("WASM_TOOLCHAINS") else { return }
+
+        try await withTemporaryDirectory { (tmpDir: Path) in
+            let testProject = TestProject(
+                "TestProject",
+                sourceRoot: tmpDir,
+                groupTree: TestGroup(
+                    "SomeFiles",
+                    children: [
```
jakepetroules, line 32: Would be great to have this test cover C/C++ as well, not just Swift.

Sources/SWBWebAssemblyPlatform/Plugin.swift
owenv: It's an interesting question, since I think right now WebAssembly doesn't have anything like a 'universal' binary. I'm not sure I have a good answer right now

kateinoigakukun: > by adding all the triples to a single SDK and then having dedicated build settings that control which ABI variant you're building against?

I agree that it's more straightforward mapping from Swift SDK to SwiftBuild SDK.

However, what is holding me back from adopting this approach is that I don't know what to assign to the variant's name. In the Swift SDK, target triple is the key for each variant,  but I am wondering if using target triple as SDK_VARIANT is the natural way for SwiftBuild as other variants are `macosx`, `iosmac`, and so on.

Sources/SWBWebAssemblyPlatform/SwiftSDK.swift
kateinoigakukun: Moved it to `SWBCore` module

Sources/SWBWebAssemblyPlatform/Plugin.swift
owenv: SDK_VARIANT is not very flexible today, but it we're able to lift some of its limitations it could be a good fit

Package.swift
```diff
@@ -191,6 +191,10 @@ let package = Package(
             name: "SWBUniversalPlatform",
             dependencies: ["SWBCore"],
             swiftSettings: swiftSettings),
+        .target(
+            name: "SWBWebAssemblyPlatform",
+            dependencies: ["SWBCore"],
+            swiftSettings: swiftSettings),
```
owenv, line 197: Could you please rebase this and change `swiftSettings` to `swiftSettings(.v6)`? I made some build changes that interfered, sorry

owenv (commented):
I think this needs a rebase and small change to pickup some Package.swift changes, but otherwise this looks ready to merge

neonichu (approved):

owenv (approved):

owenv:
@swift-ci test

owenv:
CI is an active WIP, so the jobs will fail, but give some signal on the health of the macOS build

owenv:
This looks good to merge now

kateinoigakukun:
Thanks!

Sources/SWBWebAssemblyPlatform/Plugin.swift
jakepetroules: Seems fine for now, we can always revisit as we get a better feel for how this and Swift SDKs fit together with Swift Build

PR #54, opened 2025-02-02
[SWBCore] Enhance LibSwiftDriver quality

MojtabaHs:
This PR is backward compatible.
Note that the changed variable is local and not visible from the outside of the function body. So it is safe to change.


jakepetroules (approved):

owenv:
@swift-ci test

owenv (approved):

PR #56, opened 2025-02-02
Clean up legacy ReStructuredText in existing XCSpec documentation

heckj:
I tried to convert the bare text `eval_rst` over to DocC format to support viewing it through DocC. DocC's table support is more limited than what ReStructured Text provides though (no embedded tables), so I've broken out the (few) tables that were embedded and referenced them below the main table.

The viewing experience in DocC itself leaves the content fairly "squashed", so this may not be desirable as an end state, but wanted to offer it up.


jakepetroules:
Great cleanup, thanks for this!

owenv (approved):
thanks!

PR #58, opened 2025-02-02
Narrow import access level in a few places

owenv:


owenv:
@swift-ci test

compnerd (commented):
I have a slightly different variant of this that I can put up. I've been trying to clean up the warnings as it has made dealing with the Windows build very painful.

jakepetroules (approved):

owenv:
@compnerd I can hold this open for a bit if you'd like to try a different change. Just FYI I'm about to land https://github.com/swiftlang/swift-build/pull/60 which impacts various public/package usage to work around a compiler bug.

compnerd:
I've put up https://github.com/swiftlang/swift-build/pull/62. I think that I will leave it to you to figure out how to merge the two (you want to merge your changes into that, merge my changes into this change). I'd like to get enough of the build sorted out so that I can take a look at what we can do to make progress towards the CMake build so that we can start integrating swift-build into the Windows toolchains.

PR #60, opened 2025-02-02
Access level changes to work around thunk generation bug

owenv:
Avoid calls to superclass package initializers from final internal subclass initializers so the compiler doesn't generate broken, unused thunks


owenv:
@swift-ci test

jakepetroules (approved):

PR #62, opened 2025-02-02
SWBCore,SWBTaskConstruction,SWBTaskExecution,SWBTestSupport,SWBUtil: …

compnerd:
…narrow some imports

Narrow the scope of imports for some of the import statements across the package. This avoids some unnecessary warnings.

### DIFF
--- a/Sources/SWBCore/Extensions/ToolchainRegistryExtension.swift
+++ b/Sources/SWBCore/Extensions/ToolchainRegistryExtension.swift
@@ -11,7 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 public import SWBUtil
-public import SWBMacro
+private import SWBMacro
 
 public struct ToolchainRegistryExtensionPoint: ExtensionPoint, Sendable {
     public typealias ExtensionProtocol = ToolchainRegistryExtension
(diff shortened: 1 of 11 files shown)


owenv:
@swift-ci test

owenv:
I'm expecting the macOS build should pass, but the tests will fail. The linux build will fail. There are a few issues with the configuration of the CI environment currently which I'm following up on.

owenv:
```
115 | extension llbuild_pid_t {
116 |     public var pid: pid_t {
    |                |- error: property cannot be declared public because its type uses a private type
    |                `- note: type alias 'pid_t' is imported by this file as 'private' from 'SWBLibc'
```

I think this is a difference in how the type is defined on Windows vs macOS

compnerd:
Okay, should we just conditionally change the access level then?

owenv:
I think that's probably best, I don't see a better way to avoid the inconsistency here

jakepetroules:
@swift-ci test macOS

jakepetroules (approved):

compnerd:
@swift-ci please test macOS platform

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #64, opened 2025-02-03
Stop using static wrapped properties for feature flags

owenv:
Replace them with immutable properties so they can be accessed concurrently


owenv:
@swift-ci test

jakepetroules (approved):

PR #65, opened 2025-02-03
Tests: simplify some expectations (NFCI)

compnerd:
These expectations and requirements are unnecessary. The expectations are provably tautological. Simply remove them to reduce clutter and warnings from the compiler.

### DIFF
--- a/Tests/SWBTaskExecutionTests/InfoPlistProcessorTaskTests.swift
+++ b/Tests/SWBTaskExecutionTests/InfoPlistProcessorTaskTests.swift
@@ -377,7 +377,7 @@ fileprivate struct InfoPlistProcessorTaskTests: CoreBasedTests {
             let platformName = "macosx"
             let executionDelegate = MockExecutionDelegate(core: try await getCore())
 
-            let action = InfoPlistProcessorTaskAction(try prepareContext(InfoPlistProcessorTaskActionContext(scope: scope, productType: nil, platform: try #require(core.platformRegistry.lookup(name: platformName)), sdk: try #require(core.sdkRegistry.lookup(platformName)), sdkVariant: nil, cleanupRequiredArchitectures: []), fs: executionDelegate.fs))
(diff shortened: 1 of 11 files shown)


jakepetroules (approved):

Tests/SWBTaskExecutionTests/InfoPlistProcessorTaskTests.swift
```diff
@@ -1582,7 +1582,7 @@ fileprivate struct InfoPlistProcessorTaskTests: CoreBasedTests {
         let platform = try #require(core.platformRegistry.lookup(name: platformName), "invalid platform name '\(platformName)'")
         let executionDelegate = MockExecutionDelegate(core: try await getCore())
 
-        let action = InfoPlistProcessorTaskAction(try prepareContext(InfoPlistProcessorTaskActionContext(scope: scope, productType: nil, platform: platform, sdk: try #require(core.sdkRegistry.lookup(platformName)), sdkVariant: nil, cleanupRequiredArchitectures: []), fs: executionDelegate.fs))
```
owenv, line 1585: This and a few similar examples should be valid - `try #require(core.sdkRegistry.lookup(platformName))` looks up an SDK, assert that it is non-nil, and then passes it to the InfoPlistProcessorTaskActionContext via a T-to-T? conversion. I'm also not seeing this warn locally - perhaps the implementation of the #require macro has changed?

I think I'm ok taking this change anyway, since a nil SDK here should cause downstream failures, they just won't be quite as close to where the problem occurred

owenv:
@swift-ci test

owenv (approved):

Tests/SWBTaskExecutionTests/InfoPlistProcessorTaskTests.swift
compnerd: The compiler was pointing out that the return type doesn't indicate optionality and so the require is unnecessary.

owenv: Huh, `lookup` is defined as `package func lookup(_ name: String) -> SDK?`. But I'm fine with landing this

compnerd:
@owenv is this safe to merge? Seems that there is more work needed on the CI before that is able to give valuable signal.

owenv:
@compnerd yes, go ahead and merge, I did some manual verification and everything looks ok. I'm continuing to work on the CI situation

PR #70, opened 2025-02-03
Remove a Swift parseable output test case which is obsolete

owenv:
We no longer attempt to parse any of the command fields, so let's stop asserting their presence in compiler parseable output


owenv:
@swift-ci test

PR #76, opened 2025-02-03
build: initial support for building with CMake

compnerd:
Add support for building with CMake which is required to get the toolchain bootstrapped. This partially addresses #7.


Sources/SWBMacro/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
+
+add_library(SWBMacro
```
etcwilde, line 2: Is this macro support so that Swift-Build can build macros, or are these macros used in the building of Swift-Build? If the former, this is fine. If the latter, this should be split out into a separate ExternalProject so that cross-compiling will work.

Sources/SwiftBuild/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
+
```
etcwilde, line 1: leading newline

Sources/SWBUtil/CMakeLists.txt
```diff
@@ -0,0 +1,101 @@
+
```
etcwilde, line 1: leading newline

Sources/SWBMacro/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+  MacroType.swift
+  MacroValueAssignmentTable.swift)
+set_target_properties(SWBMacro PROPERTIES
+  Swift_LANGUAGE_VERSION 6)
+target_link_libraries(SWBMacro PUBLIC
+  SWBUtil
+  SwiftDriver
+  TSCBasic)
+
+target_include_directories(SWBMacro PUBLIC
```
etcwilde, line 25: What is going on here? What bits aren't getting picked up by the linking?

etcwilde (commented):
Lots of leading newlines that should probably be the license header assuming those are still required.

Do you know if the macro bits are macro-macros, or are they support within Swift-Build for  building macros for projects that Swift-Build is building?

Sources/SWBMacro/CMakeLists.txt
compnerd: I do not understand why these are not getting picked up. There are two cases where I had to explicitly workaround the `target_link_libraries` being insufficient.

Sources/SWBMacro/CMakeLists.txt
heckj: SWBMacro is a "macro expansion inside swift-build" thing, not macro-as-compiler-extension. Details about what this is and how it works at https://github.com/swiftlang/swift-build/blob/main/SwiftBuild.docc/Core/macro-evaluation.md

compnerd:
I've tested this only on Windows so far with:
```
cmake -B out -G Ninja -S . -D CMAKE_BUILD_TYPE=Releaes -D SwiftSystem_DIR=S:\b\21\cmake\modules -D ArgumentParser_DIR=S:\b\14\cmake\modules -D TSC_DIR=S:\b\12\cmake\modules -D SwiftDriver_DIR=S:\b\15\cmake\modules -D LLBuild_DIR=S:\b\13\cmake\modules -D SQLite3_LIBRARY=S:\Library\sqlite-3.46.0\usr\lib\SQLite3.lib -D SQLite3_INCLUDE_DIR=S:\Library\sqlite-3.46.0\usr\include\ -D CMAKE_STATIC_LIBRARY_PREFIX_Swift=lib -D BUILD_SHARED_LIBS=YES -D "CMAKE_Swift_FLAGS=`"-LS:\b\5\lib`""
```

Sources/SWBMacro/CMakeLists.txt
owenv: Yeah, this is a module that contains the implementation of the "xcconfig language" which is also used for internal rule specifications, etc. No relation to swift macros

etcwilde: okay, no changes needed.

owenv:
Thanks, really appreciate your work here!  I'll see if I can test this on some of the other platforms

compnerd:
@owenv I can guarantee that it won't work on Linux at least! The missing piece is the `find_package(SwiftCrypto)` to wire up the swift-crypto dependency.

CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+                    "$<$<AND:$<COMPILE_LANGUAGE:Swift>,$<VERSION_LESS:$<TARGET_PROPERTY:Swift_LANGUAGE_VERSION>,6>>:SHELL:-enable-upcoming-feature IsolatedDefaultValues>"
+                    # rdar://127809703
+                    # "$<$<COMPILE_LANGUAGE:Swift>:SHELL:-enable-upcoming-feature RegionBasedIsolation>"
+                    "$<$<COMPILE_LANGUAGE:Swift>:SHELL:-enable-upcoming-feature ExistentialAny>"
+                    "$<$<COMPILE_LANGUAGE:Swift>:SHELL:-enable-upcoming-feature InternalImportsByDefault>")
+
+find_package(ArgumentParser)
+find_package(LLBuild)
+find_package(SwiftDriver)
+find_package(SwiftSystem)
```
owenv, line 42: One thing to consider here is that the sources are currently setup to allow using CryptoKit and System from Darwin SDKs in place of swift-crypto and swift-system. Currently, the OS copies are preferred when they exist.

Sources/SWBBuildService/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+  ClientExchangeDelegate.swift
+  DependencyGraphMessages.swift
+  DocumentationInfo.swift
+  LocalizationInfo.swift
+  Messages.swift
+  PlanningOperation.swift
+  PreviewInfo.swift
+  ProjectDescriptor.swift
+  Session.swift
+  Tools.swift)
```
iThinker, line 25: Any chance to use glob instead of manually listing all sources? Current approach introduces a lot of code to maintain, and extra work when files are added or renamed.

CMakeLists.txt
compnerd: Not sure what you have in mind. None of these are marked as `REQUIRED`. The system versions on Darwin are the preferred ones and will be given preference and if the dependencies are not found, the build would not be blocked. Of course, that is all Darwin specific.

Sources/SWBBuildService/CMakeLists.txt
compnerd: No, that doesn't work as you would expect. The glob is applied at configure time and so any changes after the initial configuration would be ignored.

etcwilde: No, globbing source files does not work and frequently causes issues with files being mysteriously omitted from the library: https://discourse.cmake.org/t/cmake-doesnt-register-new-files-when-i-add-new-files/13390

Updating the CMakeLists.txt when you add a file is part of the incremental build flow, invalidating that dependency.

CMakeLists.txt
owenv: Oh right, of course. I'd forgotten these were not required by default

jakepetroules:
I know it's not typical CMake convention, but could we consider using some kind of wildcard pattern to pick up the swift files instead of listing them individually?

compnerd:
> I know it's not typical CMake convention, but could we consider using some kind of wildcard pattern to pick up the swift files instead of listing them individually?

We can't do that reliably. The wild-carding (glob) would be applied at configure time. Any changes after the initial configuration (renaming, removal, addition) would be ignored.

etcwilde:
> I know it's not typical CMake convention, but could we consider using some kind of wildcard pattern to pick up the swift files instead of listing them individually?

[As noted on the comment above](https://github.com/swiftlang/swift-build/pull/76#discussion_r1940276982):

Globbing source files does not work and frequently causes issues with files being mysteriously omitted from the library when people try: https://discourse.cmake.org/t/cmake-doesnt-register-new-files-when-i-add-new-files

Dirtying the CMakeLists.txt file is what tells Ninja/underlying build system that it needs to re-run CMake to get the new definition of the target when doing incremental builds.

CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+                    "$<$<AND:$<COMPILE_LANGUAGE:Swift>,$<VERSION_LESS:$<TARGET_PROPERTY:Swift_LANGUAGE_VERSION>,6>>:SHELL:-enable-upcoming-feature ForwardTrailingClosures>"
+                    "$<$<AND:$<COMPILE_LANGUAGE:Swift>,$<VERSION_LESS:$<TARGET_PROPERTY:Swift_LANGUAGE_VERSION>,6>>:SHELL:-enable-upcoming-feature GlobalConcurrency>"
+                    "$<$<AND:$<COMPILE_LANGUAGE:Swift>,$<VERSION_LESS:$<TARGET_PROPERTY:Swift_LANGUAGE_VERSION>,6>>:SHELL:-enable-upcoming-feature ImplicitOpenExistentials>"
+                    "$<$<AND:$<COMPILE_LANGUAGE:Swift>,$<VERSION_LESS:$<TARGET_PROPERTY:Swift_LANGUAGE_VERSION>,6>>:SHELL:-enable-upcoming-feature ImplicitOpenExistentialsImportObjcForwardDeclarations>"
+                    "$<$<AND:$<COMPILE_LANGUAGE:Swift>,$<VERSION_LESS:$<TARGET_PROPERTY:Swift_LANGUAGE_VERSION>,6>>:SHELL:-enable-upcoming-feature InferSendableFromCaptures>"
+                    "$<$<AND:$<COMPILE_LANGUAGE:Swift>,$<VERSION_LESS:$<TARGET_PROPERTY:Swift_LANGUAGE_VERSION>,6>>:SHELL:-enable-upcoming-feature IsolatedDefaultValues>"
+                    # rdar://127809703
```
etcwilde, line 34: Where is this radar number coming from and how is it related to this option?

jakepetroules:
> > I know it's not typical CMake convention, but could we consider using some kind of wildcard pattern to pick up the swift files instead of listing them individually?
> 
> [As noted on the comment above](https://github.com/swiftlang/swift-build/pull/76#discussion_r1940276982):
> 
> Globbing source files does not work and frequently causes issues with files being mysteriously omitted from the library when people try: https://discourse.cmake.org/t/cmake-doesnt-register-new-files-when-i-add-new-files
> 
> Dirtying the CMakeLists.txt file is what tells Ninja/underlying build system that it needs to re-run CMake to get the new definition of the target when doing incremental builds.

I haven't done any serious CMake work in 10+ years and had some slight hope the situation with wildcards might've improved since then, but I guess we'll just have to deal with the static lists for now. Thanks for confirming.

CMakeLists.txt
jakepetroules: Looks like a 3 got transposed to a 2; it was referenced in Package.swift as `rdar://137809703` and this option is disabled because the Swift compiler crashes on Windows with the Swift 6.0.0 release when enabling RegionBasedIsolation without StrictConcurrency.

Sorry, I should have included a textual summary of the issue when adding this.

Sources/SWBBuildService/CMakeLists.txt
iThinker: Thanks for the answer.
Small question.
Is this list present (or will be) in any other configs nearby?
Should it have a separate single source of truth that is imported by all other build systems?

iThinker: Cmake documentation mentions `CONFIGURE_DEPENDS` that was added in 3.12 to make globing work correctly. Although it may depend on generator. https://cmake.org/cmake/help/latest/command/file.html

CMakeLists.txt
etcwilde: I was mostly poking fun at @compnerd for using numbers that have no meaning to him. 🙂 I like giving him a hard time for pasting around radar numbers that may or may not be relevant at all.

compnerd: @etcwilde obviously, I should use `radr://1` 😂

Sources/SWBBuildService/CMakeLists.txt
iThinker: Agreed. If there is no known other use for these lists they can stay as is.

out/CMakeCache.txt
```diff
@@ -0,0 +1,534 @@
+# This is the CMakeCache file.
```
owenv, line 1: Has this cache file been modified at all? If so, should we check these into a cmake/caches dir instead?

compnerd: eek, the `out` directory shouldn't be there!

owenv (approved):
I won't claim to be a CMake expert, but this set of changes lgtm

compnerd:
This change now depends on #95. I believe that we are building everything but the tests now.

compnerd:
@swift-ci please smoke test

CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+set(CMAKE_CXX_VISIBILITY hidden)
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED YES)
+set(CMAKE_CXX_EXTENSIONS NO)
+set(CMAKE_VISIBILITY_INLINES_HIDDEN YES)
+
+option(SwiftBuild_USE_LLBUILD_FRAMEWORK "Use LLBuild Framework" NO)
+
+add_compile_options("$<$<COMPILE_LANGUAGE:Swift>:SHELL:-package-name SwiftBuild>"
+                    "$<$<AND:$<COMPILE_LANGUAGE:Swift>,$<VERSION_LESS:$<TARGET_PROPERTY:Swift_LANGUAGE_VERSION>,6>>:SHELL:-enable-upcoming-feature ConciseMagicFile>"
```
etcwilde, line 25: I think in this case, it might be reasonable to factor out the version check a bit?
```cmake
set(swift6_version_check "$<VERSION_LESS:$<TARGET_PROPERTY:Swift_LANGUAGE_VERSION>,6>")

add_compile_options("$<$<COMPILE_LANGUAGE:Swift>:SHELL:-package-name SwiftBuild>"
  "$<$<AND:$<COMPILE_LANGUAGE:Swift>,${swift6_version_check}>:SHELL:-enable-upcoming-feature ConciseMagicFile>"
  "$<$<AND:$<COMPILE_LANGUAGE:Swift>,${swift6_version_check}>:SHELL:-enable-upcoming-feature DeprecateApplicationMain>"
  ...)

File: CMakeLists.txt [hunk 10]
```diff
@@ -0,0 +1,51 @@
...
+                    "$<$<AND:$<COMPILE_LANGUAGE:Swift>,$<VERSION_LESS:$<TARGET_PROPERTY:Swift_LANGUAGE_VERSION>,6>>:SHELL:-enable-upcoming-feature GlobalConcurrency>"
+                    "$<$<AND:$<COMPILE_LANGUAGE:Swift>,$<VERSION_LESS:$<TARGET_PROPERTY:Swift_LANGUAGE_VERSION>,6>>:SHELL:-enable-upcoming-feature ImplicitOpenExistentials>"
+                    "$<$<AND:$<COMPILE_LANGUAGE:Swift>,$<VERSION_LESS:$<TARGET_PROPERTY:Swift_LANGUAGE_VERSION>,6>>:SHELL:-enable-upcoming-feature ImplicitOpenExistentialsImportObjcForwardDeclarations>"
+                    "$<$<AND:$<COMPILE_LANGUAGE:Swift>,$<VERSION_LESS:$<TARGET_PROPERTY:Swift_LANGUAGE_VERSION>,6>>:SHELL:-enable-upcoming-feature InferSendableFromCaptures>"
+                    "$<$<AND:$<COMPILE_LANGUAGE:Swift>,$<VERSION_LESS:$<TARGET_PROPERTY:Swift_LANGUAGE_VERSION>,6>>:SHELL:-enable-upcoming-feature IsolatedDefaultValues>"
+                    # rdar://137809703
+                    # "$<$<COMPILE_LANGUAGE:Swift>:SHELL:-enable-upcoming-feature RegionBasedIsolation>"
+                    "$<$<COMPILE_LANGUAGE:Swift>:SHELL:-enable-upcoming-feature ExistentialAny>"
```
etcwilde, line 36: What is this indentation?

compnerd: My guess is Mayonnaise 😆

compnerd: Hmm, seems proper to me? It is the C++ style:

```cmake
function_name(parameter1
              parameter2)
```

compnerd:
@swift-ci please test

CMakeLists.txt
compnerd: Nope, but I swear I saw the same thing in the GH diff view (hence the previous joke). I think that it was a rendering issue (I've experienced that previously a few times with the web view).

jakepetroules (approved):

compnerd:
@swift-ci please test

compnerd:
Okay, this is finally ready. I've built with CMake and SPM locally. SPM's warnings about the unrecognized files are also gone. I think that we can go ahead and get this merged. We cannot enable the CI checks for this yet as this requires an updated toolchain on the Linux builders. That issue can be followed along at https://github.com/swiftlang/swift/pull/79117

owenv (approved):

owenv:
@swift-ci please test macOS

PR #89, opened 2025-02-05
Remove broken asserts in PIFLoadingTests

owenv:
These asserts never worked and aren't useful, but we're inactive until recently due to an erroneous nil check. Drop them from the test entirely

### DIFF
--- a/Tests/SWBCoreTests/PIFLoadingTests.swift
+++ b/Tests/SWBCoreTests/PIFLoadingTests.swift
@@ -1147,48 +1147,8 @@ private final class ProjectModelItemClass: ProjectModelItem {
         }
     }
 
-    @Test(.requireXcode16())
+    @Test
     func loadingStandardTarget() throws {
-        let classOneFileRef: FileReference? = try
-        {
-            let fileRefPIF: [String: PropertyListItem] = [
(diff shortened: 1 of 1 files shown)


owenv:
@swift-ci test

mhrawdon:
I think the intent of these assertions was to check that the build phases were being loaded as the correct types, and that they contained the correct build files. Could they be re-worked to check that more reliably? Or are the `PIFLoadingTests` generally not very useful at this point?

owenv:
I couldn't figure out the intent here, because we're comparing BuildFiles to FileReferences, which doesn't seem like it should ever work

jakepetroules (approved):

jakepetroules:
Yeah, this definitely seems like it was always broken.

mhrawdon:
After discussing with Owen, I'm going to take a crack at resolving this in a different way. I may do a pass through this file (which is one of the oldest files in the code base) to make it more robust (e.g., to eliminate force-unwraps).

mhrawdon:
I think [this PR](https://github.com/swiftlang/swift-build/pull/100) should fix these issues (and clean up a bunch of code besides).

PR #91, opened 2025-02-05
Update CODEOWNERS for String Catalog integration

matthewseaman:
Adding myself and @kulpreetchilana to CODEOWNERS for files involving String Catalog integration.
Also keeping the existing owners.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,3 @@
 * @jakepetroules @mirza-garibovic @mhrawdon @neonichu @owenv
+*XCStrings* @matthewseaman @kulpreetchilana @jakepetroules @mirza-garibovic @mhrawdon @neonichu @owenv
+*StringCatalog* @matthewseaman @kulpreetchilana @jakepetroules @mirza-garibovic @mhrawdon @neonichu @owenv


owenv (approved):

owenv:
@swift-ci test

jakepetroules (commented):
I don't think you need to add the rest of us here; global CODEOWNERS should be cumulative with topic-specific ones, isn't it?

matthewseaman:
> I don't think you need to add the rest of us here; global CODEOWNERS should be cumulative with topic-specific ones, isn't it?

Sadly according to the GitHub docs the last matching line takes precedence over anything before it.
https://docs.github.com/en/enterprise-cloud@latest/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

```
# Order is important; the last matching pattern takes the most
# precedence. When someone opens a pull request that only
# modifies JS files, only @js-owner and not the global
# owner(s) will be requested for a review.
*.js    @js-owner #This is an inline comment.
```

neonichu (approved):

jakepetroules:
> > I don't think you need to add the rest of us here; global CODEOWNERS should be cumulative with topic-specific ones, isn't it?
> 
> Sadly according to the GitHub docs the last matching line takes precedence over anything before it. https://docs.github.com/en/enterprise-cloud@latest/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
> 
> ```
> # Order is important; the last matching pattern takes the most
> # precedence. When someone opens a pull request that only
> # modifies JS files, only @js-owner and not the global
> # owner(s) will be requested for a review.
> *.js    @js-owner #This is an inline comment.
> ```

Huh, that's annoying.

jakepetroules (approved):

PR #93, opened 2025-02-05
Track the reason for constructing the build graph.

shawnhyam:
If the reason for the graph is 'dependencyGraph', we don't want to error out immediately when a target is not approvedByUser. We need to get further in the process so a proper error diagnostic gets created.


shawnhyam:
@swift-ci test

owenv:
@swift-ci test macOS

jakepetroules (approved):

owenv:
@swift-ci test macOS

PR #94, opened 2025-02-05
SWBBuildService: add a missing `try`

compnerd:
When building with a recent toolchain, this seems to not compile due to a missing `try`.


compnerd:
@swift-ci please test

owenv:
@compnerd This is throwing/non-throwing based on whether SWIFT_PACKAGE is defined. I think this change is reasonable, but we may want to replace some of the `#if SWIFT_PACKAGE` conditions in plugin loading with something like `#if USE_STATIC_PLUGIN_INITIALIZATION` which can be shared by the CMake build as well, as I think we will want it to use the same loading mechanism. I can take a quick look at making that change

owenv:
(https://github.com/swiftlang/swift-build/pull/95)

PR #95, opened 2025-02-05
Allow using static plugin initialization outside the package build

owenv:
We may want to adopt this style of plugin loading in the CMake build, which does not define SWIFT_PACKAGE


owenv:
@swift-ci test

neonichu (approved):

compnerd (approved):
Thanks! This seems reasonable to me, I'll need to make some tweaks to the CMake build, but this seems like it was hiding some behavioural differences! (Still haven't gotten to the running part)

jakepetroules (approved):

PR #96, opened 2025-02-05
Fetching built target descriptions can take much longer than necessary <rdar://143802224>

jonathanpenn:
Higher layer APIs properly request built target descriptions as a batch, but down here in the build system it still does a serialized fetch for each target ID that kicks off a build description request and rewalking all the targets to find the right one. When the build system is under contention (when indexing, for instance) this can cause target description requests to take much longer than necessary.

This change plumbs the batch target fetching API all the way down to the `BuildDescriptionManager`. 20x speedups observed in average cases!

### DIFF
--- a/Sources/SWBBuildService/Messages.swift
+++ b/Sources/SWBBuildService/Messages.swift
@@ -1027,47 +1027,15 @@ extension MessageHandler {
             do {
                 var responses: [PreviewInfoMessagePayload] = []
 
-                for targetID in message.targetIDs {
-                    do {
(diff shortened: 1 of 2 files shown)


jonathanpenn:
@swift-ci

owenv:
@swift-ci test macOS

owenv (approved):

jakepetroules (approved):

jonathanpenn:
@swift-ci test macOS

PR #97, opened 2025-02-05
build: update the dependencies for the plugins

compnerd:
The plugins now depend on a wider variety of modules. Reflect that into the build definition. SPM gets away with it as it emits the modules into a common directory and the current scheduling seems to allow it to build. The CMake based build system identified the missing dependencies.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -190,31 +190,31 @@ let package = Package(
 
         .target(
             name: "SWBAndroidPlatform",
-            dependencies: ["SWBCore"],
+            dependencies: ["SWBCore", "SWBMacro", "SWBUtil"],
             swiftSettings: swiftSettings(languageMode: .v6)),
         .target(
             name: "SWBApplePlatform",
-            dependencies: ["SWBCore", "SWBTaskConstruction"],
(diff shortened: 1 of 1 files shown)


owenv:
@swift-ci test

owenv (approved):
thanks - yeah this all looks correct

jakepetroules (approved):

mirza-garibovic:
I'm not opposed to listing transitive deps explicitly, but I'm not sure that we have to do that here for any real reason. `SWBMacro` and `SWBUtil` are dependencies of `SWBCore`, and SPM guarantees that you can access dependencies transitively (and a lot of packages rely on that), so we shouldn't need to list them redundantly here. If you have to list transitive dependencies in the CMake build, I would consider that a bug (or at least divergence) of the CMake Swift rules.

jakepetroules:
Why would that be a bug in CMake's Swift rules? Seems like a valid design decision to me either way.

compnerd:
@swift-ci please test

mirza-garibovic:
> Why would that be a bug in CMake's Swift rules? Seems like a valid design decision to me either way.

Yeah that's why I said "or at least divergence", meaning divergence from SPM.

PR #100, opened 2025-02-05
Fix PIFLoadingUnitTests.loadingStandardTarget() and clean up other tests in that suite.

mhrawdon:
`loadingStandardTarget()` had several checks which were failing, and had been broken for a long time but had been masked by other behavior which was recently fixed.

Many other tests were written early in the project's history, and could be updated to use newer paradigms. In particular, many unnecessary optionals and force-unwraps have been removed or converted.

One of the principles applied here is that tests should strive to perform as much as their testing as is reasonable, and not simply stop on first failure. Even lines which fail an `#expect()` can often move on to the section block of the test rather than skipping everything afterwards.

### DIFF
--- a/Tests/SWBCoreTests/PIFLoadingTests.swift
+++ b/Tests/SWBCoreTests/PIFLoadingTests.swift
@@ -630,7 +630,7 @@ private final class ProjectModelItemClass: ProjectModelItem {
             Issue.record("property list is not a dictionary")
             return
         }
-        let fileRef = try GroupTreeReference.parsePIFDictAsReference( fileRefPIF, pifLoader: pifLoader ) as! FileReference
+        let fileRef = try #require(GroupTreeReference.parsePIFDictAsReference(fileRefPIF, pifLoader: pifLoader) as? FileReference)
 
         // Examine the file reference.
         #expect(fileRef.guid == "some-fileReference-guid")
@@ -673,7 +673,7 @@ private final class ProjectModelItemClass: ProjectModelItem {
             Issue.record("property list is not a dictionary")
             return
         }
-        let fileGroup = try GroupTreeReference.parsePIFDictAsReference( fileGroupPIF, pifLoader: pifLoader ) as! FileGroup
+        let fileGroup = try #require(GroupTreeReference.parsePIFDictAsReference(fileGroupPIF, pifLoader: pifLoader) as? FileGroup)
 
         // Examine the file group.
         #expect(fileGroup.guid == "some-fileGroup-guid")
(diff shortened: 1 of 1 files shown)


mhrawdon:
@swift-ci test

owenv (approved):

neonichu (approved):

jakepetroules (approved):

Tests/SWBCoreTests/PIFLoadingTests.swift
```diff
@@ -683,24 +683,28 @@ private final class ProjectModelItemClass: ProjectModelItem {
...
             #expect(fileRef.sourceTree == SourceTree.groupRelative)
             #expect(fileRef.path.stringRep == "ClassOne.m")
             #expect(fileRef.fileTypeIdentifier == "sourcecode.c.objc")
             #expect(fileRef.regionVariantName == nil)
         }
-        do
-        {
-            let fileRef = fileGroup.children[1] as! FileReference
+        catch {
+            Issue.record(error)
```
plemarquand, line 695: This is going to produce two issues, one captured by the #expect macro failing and then the other captured manually here, which probably isn't the intention. 

Could these tests be decomposed in to a swift-testing parameterized test instead? You could create a struct where all the info in each `do` block is captured, and then provide an array of these as parameters to the test. That way each file gets its own test run.

grynspan: Or just plain allow the test to fail if `try #require()` fails.

Or use `try?` instead of `do`/`catch`, since presumably the expression being passed to `#require()` does not throw an error:

```swift
if let fileRef = try? #require(...) {
  #expect(fileRef.guid == ...)
  ...
}
```

mhrawdon: Hmm, I didn't realize it would _both_ throw an error and record an issue. I think I like the `if-let` model that @grynspan suggests in that case, since the only reason I was catching the exception was to record the issue. Will do a pass through the file to make that change. Thanks for the tip!

mhrawdon: Okay, I've updated with those changes. The changes introduce some new build warnings due to [this compiler bug](https://github.com/swiftlang/swift/issues/79202), which hopefully can be fixed before too long. But I don't think it's worth working around with more cumbersome code just to avoid those warnings.

mhrawdon:
@swift-ci test

Tests/SWBCoreTests/PIFLoadingTests.swift
```diff
@@ -683,18 +683,14 @@ private final class ProjectModelItemClass: ProjectModelItem {
         #expect(fileGroup.children.count == 2)
 
         // Examine its children
-        do
-        {
-            let fileRef = fileGroup.children[0] as! FileReference
+        if let fileRef = try? #require(fileGroup.children.first as? FileReference) {
```
jakepetroules, line 686: Unfortunately this pattern introduced some warnings and also unintentionally changed behavior -- since there is no else case, a previous crash is now an unexpected pass rather than a failure. Fixing in #135

PR #103, opened 2025-02-06
Force write updates to preparation nodes

hamishknight:
Make sure we always bump the mod time of a preparation node when it gets run since clients rely on the timestamp to know when the preparation has changed.

rdar://143757173


hamishknight:
@swift-ci please test

owenv (approved):

PR #109, opened 2025-02-06
SWBTestSupport,SWBUtil,Tests: avoid some won't be executed warnings

compnerd:
This cleans up some warnings on Windows about code that will never be executed due to the earlier returns in the Windows paths.

### DIFF
--- a/Sources/SWBTestSupport/Misc.swift
+++ b/Sources/SWBTestSupport/Misc.swift
@@ -221,8 +221,10 @@ extension ProcessInfo {
                 return p[0] == 1
             }
         }
-        #endif
         return false
+        #else
+        return false
+        #endif
     }
 
     // Get memory usage of current process in bytes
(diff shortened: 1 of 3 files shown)


compnerd:
@swift-ci please test

Sources/SWBTestSupport/Misc.swift
```diff
@@ -221,8 +221,8 @@ extension ProcessInfo {
                 return p[0] == 1
             }
         }
-        #endif
```
owenv, line 224: This needs a #else branch to continue building on non-Windows

compnerd: Ah, right, Linux won't be able to `import Darwin`.

compnerd:
@swift-ci please test

owenv (approved):

jakepetroules (approved):

PR #115, opened 2025-02-07
Add SWIFT_WARNINGS_AS_ERROR_GROUPS

swiftlysingh:
Implements SWIFT_WARNINGS_AS_ERROR_GROUPS as discussed in #8.

Please note that while I believe the core implementation is complete, I haven't finished developer testing, and the test cases might still need adjustments. I'm actively working on these and will update the pull request once it's ready for thorough review

### DIFF
--- a/Tests/SWBTaskConstructionTests/SwiftTaskConstructionTests.swift
+++ b/Tests/SWBTaskConstructionTests/SwiftTaskConstructionTests.swift
@@ -3917,6 +3917,94 @@ fileprivate struct SwiftTaskConstructionTests: CoreBasedTests {
             results.checkNoDiagnostics()
         }
     }
+
+    @Test(.requireSDKs(.host))
+    func warningAsWarningsGroups() async throws {
+        try await withTemporaryDirectory { tmpDir in
+            let testProject = try await TestProject(
+                "ProjectName",
+                sourceRoot: tmpDir.join("srcroot"),
+                groupTree: TestGroup(
+                    "SomeFiles",
+                    children: [
+                        TestFile("File1.swift"),
+                    ]),
+                targets: [
+                    TestStandardTarget(
+                        "TargetName",
+                        type: .framework,
+                        buildConfigurations: [
+                            TestBuildConfiguration("Debug", buildSettings: [
+                                "SWIFT_WARNINGS_AS_WARNINGS_GROUPS": "Unsafe DeprecatedDeclaration",
+                                "SWIFT_EXEC": swiftCompilerPath.str,
+                                "CODE_SIGN_IDENTITY": ""
+                            ]),
+                        ],
+                        buildPhases: [
+                            TestSourcesBuildPhase([
+                                TestBuildFile("File1.swift"),
+                            ]),
+                        ])
+                ])
+
+            let tester = try await TaskConstructionTester(getCore(), testProject)
+
+            try await tester.checkBuild(BuildParameters(configuration: "Debug", overrides: ["SWIFT_VERSION": swiftVersion])) { results in
+                results.checkTarget("TargetName") { target in
+                    results.checkTask(.matchTarget(target), .matchRuleType("SwiftDriver Compilation")) { task in
+                        task.checkCommandLineContains([
+                            "-Wwarning", "Unsafe",
+                            "-Wwarning", "DeprecatedDeclaration"
+                        ])
+                    }
+                }
+            }
+        }
+    }
+
+    @Test(.requireSDKs(.host))
+    func warningAsErrorsGroups() async throws {
+        try await withTemporaryDirectory { tmpDir in
+            let testProject = try await TestProject(
+                "ProjectName",
+                sourceRoot: tmpDir.join("srcroot"),
+                groupTree: TestGroup(
+                    "SomeFiles",
+                    children: [
+                        TestFile("File1.swift"),
+                    ]),
+                targets: [
+                    TestStandardTarget(
+                        "TargetName",
+                        type: .framework,
+                        buildConfigurations: [
+                            TestBuildConfiguration("Debug", buildSettings: [
+                                "SWIFT_WARNINGS_AS_ERRORS_GROUPS": "UnknownWarningGroup PreconcurrencyImport",
+                                "SWIFT_EXEC": swiftCompilerPath.str,
+                                "CODE_SIGN_IDENTITY": ""
+                            ]),
+                        ],
+                        buildPhases: [
+                            TestSourcesBuildPhase([
+                                TestBuildFile("File1.swift"),
+                            ]),
+                        ])
+                ])
+
+            let tester = try await TaskConstructionTester(getCore(), testProject)
+
+            try await tester.checkBuild(BuildParameters(configuration: "Debug", overrides: ["SWIFT_VERSION": swiftVersion])) { results in
+                results.checkTarget("TargetName") { target in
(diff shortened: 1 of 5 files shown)


swiftlysingh:
@swift-ci test

owenv:
Thanks! Just as a general note, I wonder if the UX here might be a bit nicer if -Werror and -Wwarning were split into two separate settings, since the `:error/:warning` suffix may not be as discoverable to users. Feel free to ping when this is ready for review.

Tests/SWBTaskConstructionTests/SwiftTaskConstructionTests.swift
```diff
@@ -3917,6 +3917,61 @@ fileprivate struct SwiftTaskConstructionTests: CoreBasedTests {
             results.checkNoDiagnostics()
         }
     }
+    @Test(.requireSDKs(.macOS))
```
jakepetroules, line 3920: Nothing about this test looks macOS-specific, so we should use `.requireSDKs(.host)` in order to maintain coverage across all host platforms including Window, Linux, etc.

swiftlysingh: The existing tests were non-functional, so I removed them. I'm currently facing challenges with developer testing these changes and writing appropriate unit tests. Any guidance you can provide on the best approach, or suggestions on references within the codebase, would be greatly appreciated.

cc: @owenv @jakepetroules

swiftlysingh:
> Thanks! Just as a general note, I wonder if the UX here might be a bit nicer if -Werror and -Wwarning were split into two separate settings, since the `:error/:warning` suffix may not be as discoverable to users. Feel free to ping when this is ready for review.

Thanks for the feedback. I’ve updated the code to separate the settings into two distinct options: `SWIFT_WARNINGS_AS_WARNINGS_GROUPS` and `SWIFT_WARNINGS_AS_ERROR_GROUPS`. 

Could you please let me know if the naming aligns with standards and if the implementation looks solid? 

@owenv

owenv:
The naming and implementation here looks good to me. One tiny change I'd recommend is renaming SWIFT_WARNINGS_AS_ERROR_GROUPS to SWIFT_WARNINGS_AS_ERRORS_GROUPS just so both settings are consistently plural

Tests/SWBTaskConstructionTests/SwiftTaskConstructionTests.swift
owenv: For tests I'd recommend starting out with a copy of SwiftTaskConstructionTests.upcomingFeatures(). This test is testing passing `-enable-upcoming-feature` to the Swift compiler, which is pretty similar to the new settings you're adding here. Then, I would make two changes:

1. Update the test project to use the new settings you added
2. Update the `tester.checkBuild` call at the bottom of the test to assert the Swift command line has the right flags corresponding to those settings

swiftlysingh:
> For tests I'd recommend starting out with a copy of SwiftTaskConstructionTests.upcomingFeatures(). This test is testing passing `-enable-upcoming-feature` to the Swift compiler, which is pretty similar to the new settings you're adding here. Then, I would make two changes:
> 1. Update the test project to use the new settings you added
> 2. Update the `tester.checkBuild` call at the bottom of the test to assert the Swift command line has the right flags corresponding to those settings



I have updated name and added tests. This was very helpful, thanks! 

Upon further looking into `-enable-upcoming-feature`, I was wondering if the new settings would be better as a `Bool` instead of current `StringList`? What do you all think?

Tests/SWBTaskConstructionTests/SwiftTaskConstructionTests.swift
```diff
@@ -3917,6 +3917,88 @@ fileprivate struct SwiftTaskConstructionTests: CoreBasedTests {
...
+                    children: [
+                        TestFile("File1.swift"),
+                    ]),
+                targets: [
+                    TestStandardTarget(
+                        "TargetName",
+                        type: .framework,
+                        buildConfigurations: [
+                            TestBuildConfiguration("Debug", buildSettings: [
+                                "SWIFT_WARNINGS_AS_WARNINGS_GROUPS": "-Wwarning",
```
owenv, line 3938: Instead of the flag itself, the argument to this setting should be something like `DeprecatedDeclaration` or `Unsafe` as a way of controlling the behavior of the diagnostic groups defined here: https://github.com/swiftlang/swift/blob/main/include/swift/AST/DiagnosticGroups.def. It's important that it's a StringList typed setting so that clients can specify an arbitrary list of groups

Tests/SWBTaskConstructionTests/SwiftTaskConstructionTests.swift
```diff
@@ -3917,6 +3917,88 @@ fileprivate struct SwiftTaskConstructionTests: CoreBasedTests {
...
+                            ]),
+                        ])
+                ])
+
+            let tester = try await TaskConstructionTester(getCore(), testProject)
+
+            await tester.checkBuild(BuildParameters(configuration: "Debug", overrides: ["SWIFT_VERSION": "5.0"])) { results in
+                results.checkTarget("TargetName") { target in
+                    results.checkTask(.matchTarget(target), .matchRuleType("SwiftDriver Compilation")) { task in
+                        task.checkCommandLineContains(["-Wwarning"])
```
owenv, line 3955: Similarly, this check should assert that we see both the `-Wwarning` option and its argument on the resulting command line. So assuming the test used the `Unsafe` group, we'd check for `task.checkCommandLineContains(["-Wwarning", "Unsafe"])`

Tests/SWBTaskConstructionTests/SwiftTaskConstructionTests.swift
swiftlysingh: This makes a lot of sense. Thanks for the clarification! 

Just to make sure, that these diagnostic groups would be same for both the new settings?

Tests/SWBTaskConstructionTests/SwiftTaskConstructionTests.swift
```diff
@@ -3917,6 +3917,88 @@ fileprivate struct SwiftTaskConstructionTests: CoreBasedTests {
...
+                            ]),
+                        ])
+                ])
+
+            let tester = try await TaskConstructionTester(getCore(), testProject)
+
+            try await tester.checkBuild(BuildParameters(configuration: "Debug", overrides: ["SWIFT_VERSION": swiftVersion])) { results in
+                results.checkTarget("TargetName") { target in
+                    results.checkTask(.matchTarget(target), .matchRuleType("SwiftDriver Compilation")) { task in
+                        task.checkCommandLineContains(["-Werror", "UnknownWarningGroup", "PreconcurrencyImport"])
```
owenv, line 3996: The compiler flag accepts a single argument, so here we should check for the pattern `-Werror UnknownWarningGroup -Werror PreconcurrencyImport`

owenv (approved):

owenv:
@swift-ci test

owenv:
@swiftlysingh could you please squash these changes into a single commit before I merge?

swiftlysingh:
Hey, @owenv, I have  squashed the commits. 

I saw the tests were failing for Linux host, are these settings specific for MacOS? Should I update the tests to run on MacOS host only? I am unable to find the log files for the failing tests that ran before.

owenv:
There are a few known issues on Linux atm, we're waiting to pick up a new nightly toolchain with fixes. The failures are unrelated to your change here

owenv:
@swift-ci test

PR #121, opened 2025-02-08
Complete Sendability audit of Spec class hierarchy

owenv:
This mostly consists of more aggressively marking classes as final, applying unchecked Sendable where needed, and auditing some more of the macro infrastructure for Sendability.


owenv:
@swift-ci test

jakepetroules (approved):

owenv:
yeah, we should look at migrating it to a hierarchy of protocols at some point

PR #122, opened 2025-02-08
Refactor WorkspaceContext so it can be Sendable

owenv:
Only the last commit is new. Refactor SDKRegistry and WorkspaceContext so they can be annotated Sendable


jakepetroules (approved):

owenv:
@swift-ci test

PR #123, opened 2025-02-08
Fix Sendable warnings in BuildDescriptionManager

owenv:
Guard some state with mutexes which could be accessed concurrently bu build description requests


owenv:
@swift-ci test

mirza-garibovic (approved):

jakepetroules (approved):

PR #129, opened 2025-02-09
Add NSCache wrapper to resolve Sendability warnings in Cache impl

owenv:


owenv:
@swift-ci test

Sources/SWBUtil/Cache.swift
```diff
@@ -41,49 +41,58 @@ public protocol CacheableValue {
     var cost: Int { get }
 }
 
+// NSCache insert, remove, and lookup operations are documented as thread safe: https://developer.apple.com/documentation/foundation/nscache
```
jakepetroules, line 44: @parkera Should NSCache be formally marked Sendable?

jakepetroules (approved):

Sources/SWBUtil/Cache.swift
parkera: It would require the objects themselves to also be Sendable, which we cannot guarantee.

parkera: Cc @jmschonfeld if he had any other thoughts on this. N

mirza-garibovic: Couldn't we do `extension NSCache: Sendable where KeyType: Sendable, ObjectType: Sendable`?

owenv: I think the delegate would also need to be constrained to be Sendable, which isn't enforced today. That's partially why I used the narrowly-scoped wrapper here.

jmschonfeld: Yeah making `NSCache` `Sendable` would require it to be conditional on the `KeyType` and `ObjectType` being `Sendable` which we can do, but other things it would require would be:

- All of the other operations on `NSCache` that set/get mutable state (the name, eviction policies, etc.) would need to be guarded by a lock to be synchronized. Some are on Darwin (like the `name` property) but that isn't guarded on non-Darwin and even Darwin doesn't synchronize the other properties like eviction policies
- The delegate protocol (and all delegates) would need to be `Sendable`
- We'd need to declare & require that all subclasses in any client must also be `Sendable` and need to synchronize any mutable state that they have

For those reasons, we haven't made `NSCache` `Sendable` today, but you can do what you have here as a "workaround" if you know you're using it in a safe manner. Rather than trying to retrofit `Sendable` onto `NSCache`, I think it'd be better to invest the energy into what a made-for-swift cache API would look like that doesn't necessarily bring along Objective-C semantics like subclassing, the delegate pattern, the requirement that the keys/values be `NSObject` types, etc. and the workaround can be used with `NSCache` for cases where it's safe in the meantime

mhrawdon (approved):

owenv:
@swift-ci test macOS

owenv:
@swift-ci test

PR #130, opened 2025-02-09
Remove delegate customization in TargetBuildGraphFactory delegate

owenv:
The test harnesses don't need anything other than a EmptyTargetDependencyResolverDelegate, which is trivially Sendable


owenv:
@swift-ci test

jakepetroules (approved):

PR #133, opened 2025-02-10
Eliminate nearly all remaining force casts in tests

jakepetroules:
Instead, make use of #require/throws to defend against runtime crashes.


jakepetroules:
@swift-ci test

PR #135, opened 2025-02-10
Fix accidentally-introduced warnings and regression in test behavior in PIFLoadingTests

jakepetroules:
Some of the conditional checks had redundant uses of #require, and regressed previous behavior by failing to include an else case to induce a test failure when the condition was not satisfied.

### DIFF
--- a/Tests/SWBCoreTests/PIFLoadingTests.swift
+++ b/Tests/SWBCoreTests/PIFLoadingTests.swift
@@ -683,19 +683,24 @@ private final class ProjectModelItemClass: ProjectModelItem {
         #expect(fileGroup.children.count == 2)
 
         // Examine its children
-        if let fileRef = try? #require(fileGroup.children.first as? FileReference) {
+        if let fileRef = fileGroup.children.first as? FileReference {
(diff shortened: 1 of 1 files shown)


jakepetroules:
@swift-ci test

grynspan:
There is no need for an else case though. An issue is recorded by `#require` on failure even if the error is caught.

grynspan:
(And the warnings were compiler bugs. You missed the discussion I think?)

mhrawdon:
Right, as @grynspan says, these changes aren't necessary:

1. `#require` will already generate an issue in addition to throwing, [which is documented](https://developer.apple.com/documentation/testing/require(_:_:sourcelocation:)-5l63q).
2. The warnings emitted by this use are a compiler bug, [tracked here](https://github.com/swiftlang/swift/issues/79202), but we shouldn't work around non-blocking compiler bugs just to sidestep some temporary errors.

We should undo these changes as the old version was more compact and ergonomic. I'll put up a PR.

mhrawdon:
Reverting in https://github.com/swiftlang/swift-build/pull/138

PR #136, opened 2025-02-10
Mark simulatedSingleInputlessOutputlessCommand as requireThreadSafeWorkingDirectory

jakepetroules:
This will skip the test on Amazon Linux 2 where posix_spawn_file_actions_addchdir_np is unavailable.


jakepetroules:
@swift-ci test

PR #138, opened 2025-02-10
Revert "Fix accidentally-introduced warnings and regression in test behavior in PIFLoadingTests"

mhrawdon:
The original code was correct and these changes overlooked some context as to why the original code was the way it was (e.g., the warnings are due to a compiler bug, `#require` will already emit an issue).

This reverts commit a6a6bb748e130ef75b39d5d2bc039108d82497b8.

### DIFF
--- a/Tests/SWBCoreTests/PIFLoadingTests.swift
+++ b/Tests/SWBCoreTests/PIFLoadingTests.swift
@@ -683,24 +683,19 @@ private final class ProjectModelItemClass: ProjectModelItem {
         #expect(fileGroup.children.count == 2)
 
         // Examine its children
-        if let fileRef = fileGroup.children.first as? FileReference {
+        if let fileRef = try? #require(fileGroup.children.first as? FileReference?) {
             #expect(fileRef.guid == "first-fileReference-guid")
             #expect(fileRef.sourceTree == SourceTree.groupRelative)
             #expect(fileRef.path.stringRep == "ClassOne.m")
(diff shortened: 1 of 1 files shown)


Tests/SWBCoreTests/PIFLoadingTests.swift
```diff
@@ -683,24 +683,19 @@ private final class ProjectModelItemClass: ProjectModelItem {
         #expect(fileGroup.children.count == 2)
 
         // Examine its children
-        if let fileRef = fileGroup.children.first as? FileReference {
+        if let fileRef = try? #require(fileGroup.children.first as? FileReference) {
```
grynspan, line 686: I bet you can suppress the warning with:

```suggestion
        if let fileRef = try? #require(fileGroup.children.first as? FileReference?) {
```

neonichu (approved):

Tests/SWBCoreTests/PIFLoadingTests.swift
mhrawdon: That works, and seems a simple way to suppress all but one of these warnings (the outlier isn't downcasting), while keeping the general approach. so I've incorporated it into the diffs here. Thanks!

mhrawdon:
@swift-ci test

jakepetroules:
> #require will already emit an issue

True, I missed that, thanks. Let's figure out a workaround for the remaining warning. It might be a compiler issue but it could be a while before a fix is in hand.

grynspan:
I can offer a workaround, it's just unsightly :)

PR #140, opened 2025-02-11
[DO NOT MERGE] Demonstrate SwiftTesting bug

jakepetroules:
This will unexpectedly pass without recording an issue, even though it's supposed to


jakepetroules:
@swift-ci test

PR #143, opened 2025-02-11
Remove optional casts in PIFLoadingTests.

mhrawdon:
This was an attempt to suppress an incorrect build warning, but it is functionally wrong as it results in a double optional, so restore the earlier version.


mhrawdon:
@swift-ci test

owenv (approved):

jakepetroules:
Thanks!

PR #144, opened 2025-02-11
Add some clarifying comments to XCFramework and related structs.

mhrawdon:
The Mac Catalyst properties in an XCFramework's Info.plist are a little surprising ("why is it iOS and not macOS?") until you understand that it's capturing exactly the target triple it was compiled with, so I wanted to make it explicit.

And the serialization struct is pretty bare bones, so I wanted to add a pointer to the `XCFramework` struct itself which is where almost all of the documentation is.


mhrawdon:
@swift-ci test

owenv (approved):

mirza-garibovic (approved):

PR #145, opened 2025-02-12
Wire up support for module wrapping on non-Darwin platforms

owenv:


owenv:
@swift-ci test

owenv:
fyi @cmcgee1024 since I think we discussed this at some point

Sources/SWBAndroidPlatform/Plugin.swift
```diff
@@ -83,9 +83,6 @@ struct AndroidSDKRegistryExtension: SDKRegistryExtension {
         let defaultProperties: [String: PropertyListItem] = [
             "SDK_STAT_CACHE_ENABLE": "NO",
 
-            // Workaround to avoid `-add_ast_path` on Linux, apparently this needs to perform some "swift modulewrap" step instead.
```
jakepetroules, line 86: So nice to see my hacks go away :)

Sources/SWBUtil/ProcessInfo.swift
```diff
@@ -200,6 +200,15 @@ extension ImageFormat {
             return false
         }
     }
+
+    public var requiresSwiftModulewrap: Bool {
+        switch self {
+        case .macho:
+            return false
+        default:
```
jakepetroules, line 208: nit: can we prefer to list the cases explicitly? I like to stay away from `default` since it forces re-evaluation of use sites when new cases are added. Not that adding new formats is particularly likely, but I could at least see a new one for wasm being added in the future.

Tests/SWBBuildSystemTests/BuildOperationTests.swift
```diff
@@ -222,6 +222,180 @@ fileprivate struct BuildOperationTests: CoreBasedTests {
...
+                            "staticlib",
+                        ]
+                    ),
+                    TestStandardTarget(
+                        "dynamiclib",
+                        type: .dynamicLibrary,
+                        buildConfigurations: [
+                            TestBuildConfiguration("Debug", buildSettings: [
+                                "DYLIB_INSTALL_NAME_BASE": "@rpath",
+                                "DYLIB_INSTALL_NAME_BASE[sdk=linux*]": "$ORIGIN",
```
jakepetroules, line 279: nit: I'd do:

```swift
"DYLIB_INSTALL_NAME_BASE": "$ORIGIN",
"DYLIB_INSTALL_NAME_BASE[sdk=macosx*]": "@rpath",
```

so this doesn't need to be adjusted for BSD, etc., in future.

jakepetroules (approved):

Sources/SWBUtil/ProcessInfo.swift
owenv: This is really an is-not-macho check, I don't expect we will ever move away from module wrap for any future object formats

owenv:
@swift-ci test

Sources/SWBCore/Specs/Tools/SwiftCompiler.swift
```diff
@@ -2127,6 +2127,11 @@ public final class SwiftCompilerSpec : CompilerSpec, SpecIdentifierType, SwiftDi
                 return (inputs, outputs)
             }()
 
+            if cbc.scope.evaluate(BuiltinMacros.PLATFORM_REQUIRES_SWIFT_MODULEWRAP) && cbc.scope.evaluate(BuiltinMacros.GCC_GENERATE_DEBUGGING_SYMBOLS) {
+                let moduleWrapOutput = Path(moduleFilePath.withoutSuffix + ".o")
```
compnerd, line 2131: Windows really should use `.obj` instead of `.o`.

owenv: I filed https://github.com/swiftlang/swift-build/issues/149 to track this because it needs to be done in a few other places at the same time to avoid a regression

jakepetroules: @compnerd I think it's more like "MSVC ABI should use .obj and gcc/mingw ABI should use .o", right?

owenv:
@swift-ci test

owenv:
Now that linux CI is up and running we can exercise this

PR #146, opened 2025-02-12
Make task actions extensible via plugins

jakepetroules:
This is the first step towards decoupling task action implementations from being hardcoded in the task execution framework. It doesn't move any implementations yet, just adds the hooks to unblock doing so.


jakepetroules:
What do folks think about the shape of this? The task locals are a bit wonky, but the polymorphically serialized task actions class hierarchy would be a lot more complex to unwind and thread down the right context to the serialization point otherwise, and this approach gets us closer to decoupling more quickly.

neonichu (approved):

owenv (approved):

PR #147, opened 2025-02-12
Install name mapping not specifying the correct platform number (rdar://144577495)

jonathanpenn:
The `$ld$previous` syntax used to map a renamed debug dylib for previews does not properly set the platform field.

This plumbs them through as a new computed build macro and checks it in the previews settings test when `LD_CLIENT_NAME` is used.


jonathanpenn:
@swift-ci test

jakepetroules (approved):

neonichu (approved):

owenv (approved):

jonathanpenn:
The ubuntu CI had issues prior to my change. The failure I see is unrelated.

PR #150, opened 2025-02-12
Strip #if directives from xcspecs and strings files (rdar://144731866)

matthewseaman:
Some build configurations can contain `#if` directives in xcspecs and strings files.

Such configurations are responsible for doing their own unifdefing if they want but we still need to strip those out of the final files so that they can be properly parsed as plists.


matthewseaman:
@swift-ci test

matthewseaman:
@swift-ci test

owenv (approved):

neonichu (approved):

jakepetroules (approved):

matthewseaman:
@swift-ci test Windows

PR #151, opened 2025-02-12
Track cache purpose to avoid collisions between incompatible plugins

owenv:
Avoid CAS plugin errors when different versions are used by different categories of tools.


owenv:
@swift-ci test

jakepetroules (approved):

neonichu (approved):

PR #152, opened 2025-02-13
Cache setup errors should be non-fatal

owenv:
We currently rely on this error being recoverable in various contexts, especially in build operations tests. I accidentally upgraded it to a user facing build error in a previous PR


owenv:
@swift-ci test

mirza-garibovic (approved):

mirza-garibovic:
Does it get logged somewhere?

owenv:
An error will be diagnosed later on in the build if a task attempts to access the cache and it's not available

neonichu (approved):

PR #159, opened 2025-02-14
Fix some issues with extended attributes

jakepetroules:
On Linux, the name of an extended attribute "must be a null-terminated string prefixed by a namespace identifier and a dot character", and only the "user" namespace is available for unrestricted use.

This changes our use of extended attributes to prefer a Linux-compatible name where possible, and updates the unit tests to account for this restriction and enables that test on Linux.

This also clarifies the support situation for extended attributes on Windows and OpenBSD. The logical equivalent to xattrs on Windows is perhaps ADS; add a comment to that effect and a reference to the Foundation issue tracking potentially adding the support at that level. Compile out xattrs support on OpenBSD as they don't exist there at all.

### DIFF
--- a/Sources/SWBCore/WorkspaceContext.swift
+++ b/Sources/SWBCore/WorkspaceContext.swift
@@ -384,7 +384,12 @@ public final class WorkspaceContext: Sendable {
 
 extension FSProxy {
     private static var CreatedByBuildSystemAttribute: String {
+        #if os(Linux)
+        // On Linux, "the name [of an extended attribute] must be a null-terminated string prefixed by a namespace identifier and a dot character" and only the "user" namespace is available for unrestricted access.
+        "user.org.swift.swift-build.CreatedByBuildSystem"
+        #else
         "com.apple.xcode.CreatedByBuildSystem"
+        #endif
     }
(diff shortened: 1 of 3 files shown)


Sources/SWBCore/WorkspaceContext.swift
```diff
@@ -384,7 +384,12 @@ public final class WorkspaceContext: Sendable {
 
 extension FSProxy {
     private static var CreatedByBuildSystemAttribute: String {
+        #if os(Linux)
+        // On Linux, "the name [of an extended attribute] must be a null-terminated string prefixed by a namespace identifier and a dot character" and only the "user" namespace is available for unrestricted access.
+        "user.com.apple.xcode.CreatedByBuildSystem"
```
owenv, line 389: we should switch this over to `user.org.swift.swift-build` where there's no backwards compatibility concern

jakepetroules: Heh, I thought about this too. Done.

jakepetroules:
@swift-ci test

mhrawdon (approved):

PR #161, opened 2025-02-14
Fallback toolchain: locate in /usr/local prefixes in addition to /usr

jakepetroules:
It's common for some platforms (like FreeBSD, and possibly some Linux distributions) to place non-system content at a "local" subdirectory of "usr". Allow finding the host toolchain there as well. This code path currently only applies to Linux.


jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

PR #162, opened 2025-02-14
Add a convenience API for tests to get the image format of a run destination

jakepetroules:
This is often semantically more appropriate than checking the platform, and helps reduce the amount of platform-specific checks needed as we port to more platforms. Immediately make use of this in a test.


jakepetroules:
@swift-ci test

owenv (approved):

PR #163, opened 2025-02-14
Enable Windows linker discovery

kcieplak:
- Enable Windows linker discovery to get linker type and version
- Make Platform Registry initialization async
- Change the function signature of additionalPlatformExecutableSearchPaths to allow passing of a filesystem for discovery.    Plugins will need to be updated to match new signatures
- Add the visual studio install directory to platform search paths for executables.
- Add Windows platform plugin extension to get install directory
- Add new internal build setting _LINKER_EXEC for testing purposes.
- Change the search paths of a Platform to a StackedSearchPath, as done with the Toolchain Registry
- Add test for discovery of windows linker

### DIFF
--- a/Sources/SWBCore/Core.swift
+++ b/Sources/SWBCore/Core.swift
@@ -100,9 +100,10 @@ public final class Core: Sendable {
 
             await core.initializeToolchainRegistry()
 
+            await core.initializePlatformRegistry()
+
             // Force loading SDKs.
             let sdkRegistry = core.sdkRegistry
-
             for `extension` in await pluginManager.extensions(of: SDKRegistryExtensionPoint.self) {
                 await sdkRegistry.registerSDKs(extension: `extension`, platformRegistry: core.platformRegistry)
             }
@@ -178,6 +179,7 @@ public final class Core: Sendable {
     public let connectionMode: ServiceHostConnectionMode
 
     @_spi(Testing) public init(delegate: any CoreDelegate, hostOperatingSystem: OperatingSystem, pluginManager: PluginManager, developerPath: String, inferiorProductsPath: Path?, additionalContentPaths: [Path], environment: [String:String], buildServiceModTime: Date, connectionMode: ServiceHostConnectionMode) async throws {
+
         self.delegate = delegate
         self.hostOperatingSystem = hostOperatingSystem
         self.pluginManager = pluginManager
@@ -301,31 +303,10 @@ public final class Core: Sendable {
     @_spi(Testing) public var toolchainPaths: [(Path, strict: Bool)]
 
     /// The platform registry.
-    public lazy var platformRegistry: PlatformRegistry = {
-        // FIXME: We should support building the platforms (with symlinks) locally (for `inferiorProductsPath`).
-
-        // Search the default location first (unless directed not to), then search any extra locations we've been passed.
-        var searchPaths: [Path]
-        if let onlySearchAdditionalPlatformPaths = getEnvironmentVariable("XCODE_ONLY_EXTRA_PLATFORM_FOLDERS"), onlySearchAdditionalPlatformPaths.boolValue {
-            searchPaths = []
-        }
-        else {
-            let platformsDir = self.developerPath.join("Platforms")
-            searchPaths = [platformsDir]
-            if hostOperatingSystem == .windows {
-                for dir in (try? localFS.listdir(platformsDir)) ?? [] {
-                    searchPaths.append(platformsDir.join(dir))
-                }
-            }
-        }
-        if let additionalPlatformSearchPaths = getEnvironmentVariable("XCODE_EXTRA_PLATFORM_FOLDERS") {
-            for searchPath in additionalPlatformSearchPaths.split(separator: ":") {
-                searchPaths.append(Path(searchPath))
-            }
-        }
-        searchPaths += UserDefaults.additionalPlatformSearchPaths
-        return PlatformRegistry(delegate: self.registryDelegate, searchPaths: searchPaths, hostOperatingSystem: hostOperatingSystem)
-    }()
+    let _platformRegistry: UnsafeDelayedInitializationSendableWrapper<PlatformRegistry> = .init()
+    public var platformRegistry: PlatformRegistry {
+        _platformRegistry.value
+    }
 
     @PluginExtensionSystemActor public var loadedPluginPaths: [Path] {
         pluginManager.pluginsByIdentifier.values.map(\.path)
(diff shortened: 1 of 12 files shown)


kcieplak:
@swift-ci test

owenv:
@swift-ci test

Sources/SWBCore/Specs/Tools/LinkerTools.swift
```diff
@@ -1549,14 +1555,20 @@ public func discoveredLinkerToolsInfo(_ producer: any CommandProducer, _ delegat
...
                 }
-
                 let goLD = [
                     #/GNU gold version (?<version>[\d.]+)-.*/#,
                     #/GNU gold \(GNU Binutils.*\) (?<version>[\d.]+)/#,
                 ]
                 if let match = try goLD.compactMap({ try $0.firstMatch(in: String(decoding: executionResult.stdout, as: UTF8.self)) }).first {
                     return DiscoveredLdLinkerToolSpecInfo(linker: .gold, toolPath: toolPath, toolVersion: try Version(String(match.output.version)), architectures: Set())
                 }
+                // link.exe has no option to simply dump the version, but if you pass an invalid argument, the program will dump a header that contains the version.
```
owenv, line 1565: This is not great, but I can't think of anything better

owenv (approved):

Sources/SWBWindowsPlatform/Plugin.swift
```diff
@@ -25,22 +26,40 @@ struct WindowsPlatformSpecsExtension: SpecificationsExtension {
...
+        return [vcToolsInstallDir: dir.str]
     }
 }
+
+struct WindowsPlatformExtension: PlatformInfoExtension {
+    public func additionalPlatformExecutableSearchPaths(platformName: String, platformPath: Path, fs: any FSProxy) async -> [Path] {
+        guard let dir = try? await findLatestInstallDirectory(fs: fs) else {
+            return []
+        }
+        return [dir.join("bin/Hostarm64/arm64"), dir.join("bin/Hostx64/x64") ]
```
jakepetroules, line 62: While the VS installer won't place Hostarm64 on an x86 system, it might be a good idea to check `Architecture.host.stringValue == "aarch64"` anyways before adding it to the search list just because of things like network drives, etc.

Sources/SWBCore/Specs/Tools/LinkerTools.swift
jakepetroules: FWIW, invoking with no arguments will also print the version header and is a valid invocation (prints help info), so it is maybe "cleaner" than intentionally passing an invalid argument.

jakepetroules (changes requested):

Sources/SWBTestSupport/CoreBasedTests.swift
```diff
@@ -232,6 +232,28 @@ extension CoreBasedTests {
             return try #require(defaultToolchain.executableSearchPaths.findExecutable(operatingSystem: core.hostOperatingSystem, basename: "llvm-lib") ?? (localFS.exists(fallbacklibtool) ? fallbacklibtool : nil), "couldn't find llvm-lib in default toolchain")
         }
     }
+    
+    // The path to ld in default toolchain or plaftform.
+    package var ldPath: Path {
+        get async throws {
+            let (core, defaultToolchain) = try await coreAndToolchain()
+            let fallbacklibtool = Path("/usr/bin/ld")
```
compnerd, line 240: This path is a relative path on Windows. I think that we should conditionalise this on platform that we are building swift build for.

Sources/SWBTestSupport/CoreBasedTests.swift
```diff
@@ -232,6 +232,28 @@ extension CoreBasedTests {
...
+            let toolName = core.hostOperatingSystem == .windows ? "link" : "ld"
+
+            // Look in the default toolchain.
+            if let executable = defaultToolchain.executableSearchPaths.findExecutable(operatingSystem: core.hostOperatingSystem, basename: toolName) {
```
compnerd, line 244: For Windows, I think that we also want to consider `lld-link`.

Sources/SWBCore/Core.swift
```diff
@@ -423,6 +404,29 @@ public final class Core: Sendable {
...
+            let platformsDir = self.developerPath.join("Platforms")
+            searchPaths = [platformsDir]
+            if hostOperatingSystem == .windows {
+                for dir in (try? fs.listdir(platformsDir)) ?? [] {
+                    searchPaths.append(platformsDir.join(dir))
+                }
+            }
+        }
+        if let additionalPlatformSearchPaths = getEnvironmentVariable("XCODE_EXTRA_PLATFORM_FOLDERS") {
+            for searchPath in additionalPlatformSearchPaths.split(separator: ":") {
```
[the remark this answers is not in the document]
jakepetroules: Ah, good catch -- this should be `Path.pathEnvironmentSeparator`

Sources/SWBCore/Specs/Tools/LinkerTools.swift
```diff
@@ -1232,11 +1231,18 @@ public final class LdLinkerSpec : GenericLinkerSpec, SpecIdentifierType, @unchec
 
     override public func discoveredCommandLineToolSpecInfo(_ producer: any CommandProducer, _ scope: MacroEvaluationScope, _ delegate: any CoreClientTargetDiagnosticProducingDelegate) async -> (any DiscoveredCommandLineToolSpecInfo)? {
         let alternateLinker = scope.evaluate(BuiltinMacros.ALTERNATE_LINKER)
+        var linker = if alternateLinker != "" { Path(alternateLinker) } else { producer.hostOperatingSystem == .windows ? Path("link.exe") : Path("ld") }
```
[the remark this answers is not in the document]
jakepetroules: Do you mean use `clang` as the linker? Or something else?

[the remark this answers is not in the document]
compnerd: Yes, using `clang` as the linker driver (it knows what to do to locate and construct the invocation).

Sources/SWBCore/Specs/Tools/LinkerTools.swift
```diff
@@ -1540,7 +1546,7 @@ public final class LibtoolLinkerSpec : GenericLinkerSpec, SpecIdentifierType, @u
 public func discoveredLinkerToolsInfo(_ producer: any CommandProducer, _ delegate: any CoreClientTargetDiagnosticProducingDelegate, at toolPath: Path) async -> (any DiscoveredCommandLineToolSpecInfo)? {
     do {
         do {
-            let commandLine = [toolPath.str, "-version_details"]
+            let commandLine = [toolPath.str,  producer.hostOperatingSystem == .windows ? "/VERSION" : "-version_details"]
```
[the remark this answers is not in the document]
kcieplak: Scratch that.  Looking closer the logic seems off there.

Sources/SWBCore/Core.swift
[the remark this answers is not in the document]
kcieplak: Fixed

Sources/SWBWindowsPlatform/Plugin.swift
kcieplak: Done

kcieplak:
@swift-ci test

Sources/SWBCore/Specs/Tools/LinkerTools.swift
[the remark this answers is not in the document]
kcieplak: clang is used to invoke the linker.  The discovery of the linker via discoveredCommandLineToolSpecInfo attempts to find out details about the linker that clang would invoke.  Internally depending on the linker type or the version discovered different linker options can be provided.

The ALTERNATE_LINKER build setting sets the `-fuse-ld` option forcing clang to use a different named linker.   I will verify if on  Windows clang will use link.exe by default.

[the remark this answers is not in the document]
kcieplak: Verified that indeed clang on windows by default will use link.exe.

Sources/SWBTestSupport/CoreBasedTests.swift
kcieplak: Yes, the current swift package manager on windows uses lld-link.  I will change the default to that for windows platforms.

jakepetroules: @compnerd lld-link is an independent implementation which is compatible with link.exe's command line, but doesn't use link.exe -- right?

kcieplak:
@swift-ci  test

shahmishal:
@swift-ci test

Sources/SWBTestSupport/CoreBasedTests.swift
compnerd: @jakepetroules right - it is compatible with `link.exe` for invocation, but is a complete implementation. At this point, on main, the toolchain should be relatively complete. We have the last piece enabled now which was `llvm-mt` which replaces `mt`.

For users of Swift, the only thing that they should be depending on from Microsoft would be the headers and import libraries. We have equivalent tools to allow them to hopefully perform all necessary operations.

kcieplak:
I am going to rebase on main with a new change as lots of https://github.com/swiftlang/swift-build/pull/207 has been added already

PR #164, opened 2025-02-14
Infer [system] based on install path when generating clang modulemaps

owenv:
rdar://144797825

Modules intended to be installed to system locations should annotate their modules as [system]. Do this automatically when the module map is generated by the build system


owenv:
@swift-ci test

jakepetroules (approved):
Looks good, but it would be great if you could add a little more context to the PR description describing the motivation for the change.

neonichu (approved):

PR #170, opened 2025-02-16
Use llvm-ar as the linux archiver

owenv:
This is consistent with SwiftPM, and llvm-ar, unlike libtool/ar, is included in Swift.org toolchains


owenv:
@swift-ci test

mirza-garibovic (approved):

jakepetroules:
Does this solve an actual compatibility issue or just speculative?

Seems good to use LLVM toolchain components all the way rather than mixing and matching with GNU tools though.

owenv:
It's needed in the docker images where libtool & ar aren't installed

jakepetroules:
Makes sense, thanks!

PR #173, opened 2025-02-16
fix: a typo in `dependencyDataFormat`

MojtabaHs:
I couldn't find any references to this property, so it should be safe to change. but please run @swift-ci test.
Please let me know if it needs a deprecation phase or maybe removing the variable entirely. 🤷🏻‍♂️
Thanks.

### DIFF
--- a/Sources/SWBBuildSystem/BuildOperation.swift
+++ b/Sources/SWBBuildSystem/BuildOperation.swift
@@ -1078,7 +1078,7 @@ private class InProcessCommand: SWBLLBuild.ExternalCommand, SWBLLBuild.ExternalD
         return signature
     }
 
-    var depedencyDataFormat: SWBLLBuild.DependencyDataFormat {
+    var dependencyDataFormat: SWBLLBuild.DependencyDataFormat {
         switch task.dependencyData {
         case .makefile?, .makefiles?:
             return .makefile
--- a/Sources/SWBBuildSystem/BuildOperation.swift
+++ b/Sources/SWBBuildSystem/BuildOperation.swift
@@ -1078,6 +1078,11 @@ private class InProcessCommand: SWBLLBuild.ExternalCommand, SWBLLBuild.ExternalD
         return signature
     }
 
+    // temporary for compatibility
+    var depedencyDataFormat: SWBLLBuild.DependencyDataFormat {
+        dependencyDataFormat
+    }
+
     var dependencyDataFormat: SWBLLBuild.DependencyDataFormat {
         switch task.dependencyData {
         case .makefile?, .makefiles?:


mirza-garibovic:
This is a requirement of a protocol in llbuild, so I don't think this will build anymore, you have to stage a change starting from llbuild

@swift-ci test

MojtabaHs:
@mirza-garibovic Do I have access to the llbuild? 🤔

There is also a an issue in [line 1088](https://github.com/MojtabaHs/swift-build/blob/51a470b1fdd600c410efcefc8649e3fd66bb68fe/Sources/SWBBuildSystem/BuildOperation.swift#L1088)
```
case .dependencyInfo?:
    return .dependencyinfo // 👈 Here
```

But It seems I have no access to those internals to fix it. (Or Im missing something here)

mirza-garibovic:
It's a package dependency and you should have access to it, try https://github.com/apple/swift-llbuild.git

jakepetroules:
I think the order you'll want to do this in is to implement both spellings here in Swift Build, and then afterwards change llbuild's spelling (assuming there are no other clients of this API in the Swift toolchain).

MojtabaHs:
I have opened a [PR on the dependency repo](https://github.com/swiftlang/swift-llbuild/pull/975)

About the `dependencyinfo`, it looks like a pattern that uses all lowercased for all options. Otherwise this:
```
llb_buildsystem_dependency_data_format_dependencyinfo
```
should be:
```
llb_buildsystem_dependency_data_format_dependency_info
```
Which is some sort of odd. So I leave this change for the original repository for now.

jakepetroules:
In order to properly stage this breaking change across multiple repositories, you'll need to land this PR _first_. Please add the following code next to the renamed API:

```swift
    // temporary for compatibility
    var depedencyDataFormat: SWBLLBuild.DependencyDataFormat {
        dependencyDataFormat
    }
```

...and then you can land this change. Subsequently, we can land the swiftlang/swift-llbuild#975 change. Once both are in place, you can open a third PR to remove the mis-spelled API usage from swift-build.

MojtabaHs:
@jakepetroules how about this instead?
```
@available(*, deprecated, renamed: "dependencyDataFormat", message: "This property has a typo. Please use the correct one.")
var depedencyDataFormat: SWBLLBuild.DependencyDataFormat {
    dependencyDataFormat
}
```

So the compiler can automatically highlight the change and even suggest adopting the new property name.

jakepetroules:
Not really necessary to add the deprecation notice because this is just satisfying a protocol requirement. No one's calling it directly.

Also, to be clear, I think landing the three PRs to make the transition is something that should be able to be completed on the order of _days_, so it wouldn't need to stick around for very long at all.

jakepetroules:
@swift-ci test

jakepetroules (approved):

PR #180, opened 2025-02-17
Skip the body of extendedAttributesSupport when running on tmpfs

jakepetroules:
Extended attributes in the user namespace aren't supported on tmpfs, on Linux.


jakepetroules:
@swift-ci test

owenv (approved):

PR #181, opened 2025-02-17
On non-Darwin platforms, infer the default toolchain when there is only one

owenv:
Right now, this is primarily relevant on Windows, which has no fallback system toolchain and now provides a ToolchainInfo.plist with a non-default identifier

### DIFF
--- a/Sources/SWBCore/ToolchainRegistry.swift
+++ b/Sources/SWBCore/ToolchainRegistry.swift
@@ -389,6 +389,7 @@ extension Array where Element == Toolchain {
 /// The ToolchainRegistry manages the set of registered toolchains.
 public final class ToolchainRegistry: @unchecked Sendable {
     let fs: any FSProxy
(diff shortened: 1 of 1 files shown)


owenv:
@swift-ci test

owenv:
This brings us from ~1500 to ~100 failures, once libclang.dll is fixed that should address a lot of the rest

neonichu (approved):

jakepetroules:
So this is needed because ToolchainInfo.plist now exists in nightlies, but didn't before?

jakepetroules (approved):

jakepetroules:
You might want to adjust the comment:

> // Windows toolchains do not have any metadata files that define a version, so the version needs to be derived from the path.

to say "Prior to Swift 6.2..." or something, so it's clear that we can remove that fallback path soon.

jakepetroules:
#185 for the comment adjustment

PR #184, opened 2025-02-17
Address a number of warnings on Windows

compnerd:
This set cleans up most of the non-Concurrency related warnings on Windows.

### DIFF
--- a/Sources/SWBCore/LibclangVendored/CStringArray.swift
+++ b/Sources/SWBCore/LibclangVendored/CStringArray.swift
@@ -12,6 +12,12 @@
 
 import Foundation
 
+#if os(Windows)
+private func strdup(_ s: UnsafePointer<CChar>) -> UnsafeMutablePointer<CChar> {
+    return _strdup(s)
+}
+#endif
+
 /// `CStringArray` represents a C null-terminated array of pointers to C strings.
 ///
 /// The lifetime of the C strings will correspond to the lifetime of the `CStringArray`
--- a/Sources/SWBCore/ProjectModel/Reference.swift
+++ b/Sources/SWBCore/ProjectModel/Reference.swift
@@ -324,7 +324,7 @@ public final class VariantGroup: GroupTreeReference, BuildFileRepresentable, @un
 
 
 /// A ProductReference represents the product of a StandardTarget object.  It acts as a placeholder so that product can be represented in other targets, but it contains no meaningful information itself; rather, it vends information about itself by querying its target for that information.  A ProductReference object is not part of a product's group tree and has no parent property; rather, it is owned by and has an unowned backpointer to its target.
-public final class ProductReference: Reference, BuildFileRepresentable
+public final class ProductReference: Reference, BuildFileRepresentable, @unchecked Sendable
 {
     /// The name of this reference - primarily for debugging purposes.
     public let name: String
--- a/Sources/SWBUtil/FSProxy.swift
+++ b/Sources/SWBUtil/FSProxy.swift
@@ -562,9 +562,13 @@ class LocalFS: FSProxy, @unchecked Sendable {
     }
 
     func remove(_ path: Path) throws {
+        #if os(Windows)
+        try fileManager.removeItem(atPath: path.str)
+        #else
         guard unlink(path.str) == 0 else {
             throw POSIXError(errno, context: "unlink", path.str)
         }
+        #endif
     }
 
     func removeDirectory(_ path: Path) throws {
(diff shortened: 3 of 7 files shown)


compnerd:
@swift-ci please test

Sources/SWBCore/ProjectModel/Reference.swift
```diff
@@ -324,7 +324,7 @@ public final class VariantGroup: GroupTreeReference, BuildFileRepresentable, @un
 
 
 /// A ProductReference represents the product of a StandardTarget object.  It acts as a placeholder so that product can be represented in other targets, but it contains no meaningful information itself; rather, it vends information about itself by querying its target for that information.  A ProductReference object is not part of a product's group tree and has no parent property; rather, it is owned by and has an unowned backpointer to its target.
-public final class ProductReference: Reference, BuildFileRepresentable
```
owenv, line 327: This isn't safe yet, I need to refactor how it initializes the `public unowned var target: Target?` backreference before it's safe to apply `@unchecked Sendable`

Tests/SWBCoreTests/PIFLoadingTests.swift
```diff
@@ -683,14 +683,14 @@ private final class ProjectModelItemClass: ProjectModelItem {
         #expect(fileGroup.children.count == 2)
 
         // Examine its children
-        if let fileRef = try? #require(fileGroup.children.first as? FileReference) {
```
owenv, line 686: We keep having to flip these #requires back and forth due to compiler bugs, I don't know that there's a way to avoid the warnings on every platform

compnerd: Is it a platform thing or a compiler version thing? I'm building with a relatively recent compiler.

owenv: As I understand it it's a compiler version thing: https://github.com/swiftlang/swift/issues/79202

Sources/SWBCore/LibclangVendored/CStringArray.swift
```diff
@@ -12,6 +12,12 @@
 
 import Foundation
 
+#if os(Windows)
```
jakepetroules, line 15: This brings back memories :)

Sources/SWBUtil/FSProxy.swift
```diff
@@ -562,9 +562,13 @@ class LocalFS: FSProxy, @unchecked Sendable {
     }
 
     func remove(_ path: Path) throws {
+        #if os(Windows)
```
jakepetroules, line 565: I don't know that there's any value in this being platform-conditional. I think this is only using unlink in the first place out of some early attempt to avoid Foundation for whatever reason at the time which is long since irrelevant. And we're using FileManager in the implementation of most of these FSProxy these functions anyways, so we may as well get to benefit from Foundation's work.

Sources/SWBUtil/POSIX.swift
```diff
@@ -15,17 +15,22 @@ import SWBLibc
...
 #if os(Windows)
-#if canImport(System)
-import System
-#else
-import SystemPackage
-#endif
+public import Foundation
+
+private func strerror(_ code: CInt) -> String {
+    withUnsafeTemporaryAllocation(of: CChar.self, capacity: 95) {
```
jakepetroules, line 21: 95?

Sources/SWBUtil/POSIX.swift
```diff
@@ -15,17 +15,22 @@ import SWBLibc
...
+        }
+        return String(cString: $0.baseAddress!)
+    }
+}
 #endif
 
 public enum POSIX: Sendable {
     public static func getenv(_ name: String) throws -> String? {
         #if os(Windows)
-        try name.withCString(encodedAs: CInterop.PlatformUnicodeEncoding.self) { wName in
```
jakepetroules, line 28: Did this actually warn, or just stylistic?

Sources/SWBUtil/POSIX.swift
compnerd: 94 + 1; the strerror message is limited to 94 characters.

jakepetroules: Thanks, could you encode that in a comment (with a link to docs?) so it doesn't look like an arbitrary magic to future readers?

Sources/SWBUtil/POSIX.swift
jakepetroules: Aha

PR #189, opened 2025-02-18
SWBUtil: use `FileManager` for more operations

compnerd:
`unlink` is deprecated on Windows, and `RemoveFileW` should be preferred. However, that would limit the path to `MAX_PATH` (261) characters. Prefer to use `FileManager to remove the file to avoid the path limit.

Adopt the fileManager path in more locations.

### DIFF
--- a/Sources/SWBUtil/FSProxy.swift
+++ b/Sources/SWBUtil/FSProxy.swift
@@ -362,7 +362,6 @@ class LocalFS: FSProxy, @unchecked Sendable {
     /// Check whether a given path is a symlink.
     /// - parameter destinationExists: If the path is a symlink, then this `inout` parameter will be set to `true` if the destination exists.  Otherwise it will be set to `false`.
     func isSymlink(_ path: Path, _ destinationExists: inout Bool) -> Bool {
-        #if os(Windows)
         do {
             let destination = try fileManager.destinationOfSymbolicLink(atPath: path.str)
             destinationExists = exists((path.isAbsolute ? path.dirname : Path.currentDirectory).join(destination))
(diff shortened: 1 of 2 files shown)


compnerd:
This is an extended change split out from the #184 based on @jakepetroules' comment that the use of raw C APIs is no longer a design constraint. This generally reduces complexity as `FileManager` already handles these error cases. On Windows, this also makes code more robust as this allows us to work with long paths.

compnerd:
@swift-ci please test

jakepetroules:
> On Windows, this also makes code more robust as this allows us to work with long paths.

That alone is a big reason to do this, I completely forgot about the POSIX APIs not handling long paths!

jakepetroules (approved):

neonichu (approved):

jakepetroules:
@swift-ci test

Sources/SWBUtil/FSProxy.swift
```diff
@@ -397,70 +380,11 @@ class LocalFS: FSProxy, @unchecked Sendable {
...
-            var destinationExists = false
-            if isDirectory(path) {
-                // If the item at the path is a directory, then we're good.  This includes if it's a symlink which points to a directory.
-                return
-            }
-            else if isSymlink(path, &destinationExists) {
-                // If the item at the path is a symlink, then we check whether it's a broken symlink or points to something that is not a directory.
-                if destinationExists {
-                    // The destination does exist, so it's not a directory.
-                    throw StubError.error("File is a symbolic link which references a path which is not a directory: \(path.str)")
```
jakepetroules, line 425: Test expectations need to be updated to accommodate these specific error strings going away.

jakepetroules (changes requested):

Sources/SWBUtil/FSProxy.swift
compnerd: Ah, thanks!

compnerd: Hmm, I don't see a good way to handle the check. There is no guarantee that the message is going to be identical across platforms and Foundation implementations :(

jakepetroules: I think it's fine if we just check that it _does_ fail in the expected cases without checking the specific error string.

compnerd:
@swift-ci please test

jakepetroules:
Closing because these changes have already landed via other PRs.

PR #191, opened 2025-02-19
[SWBCore][ObjC] Automatically enable ObjC ARC support for Apple platforms

### LINKED ISSUE #25: Enable Objective-C ARC support automatically for Apple platforms
There is a build setting called "CLANG_ENABLE_OBJC_ARC" that should be enabled on Apple platforms, but not for any others since that supports doesn't exist in swift toolchains there. When building PIFs the client shouldn't need to be aware of this implementation detail.

Once this is in place then SwiftPM won't need the following code in the PIF Builder:

```
#if os(macOS)
        // Objective-C support only for macOS
        settings[.CLANG_ENABLE_OBJC_ARC] = "YES"
#endif
```

Rajveer100:
Resolves #25

This toggles the `CLANG_ENABLE_OBJC_ARC` boolean macro.

### DIFF
--- a/Sources/SWBCore/Settings/Settings.swift
+++ b/Sources/SWBCore/Settings/Settings.swift
@@ -2436,6 +2436,26 @@ private class SettingsBuilder {
             platformTable.push(macro, realArchs)
         }
 
+        // Add ObjC ARC support for Apple platforms.
+        if let platform {
+            switch platform.name {
+            case "macosx",
+                "iphoneos",
+                "iphonesimulator",
+                "appletvos",
+                "appletvsimulator",
+                "watchos",
+                "watchsimulator",
+                "xros",
+                "xrsimulator":
+                if let project, project.isPackage {
+                    platformTable.push(BuiltinMacros.CLANG_ENABLE_OBJC_ARC, literal: true)
+                }
+            default:
+                break
+            }
+        }
+
         // Push the table.
         push(platformTable, .exported)
     }


Sources/SWBCore/Settings/Settings.swift
```diff
@@ -1573,6 +1573,12 @@ private class SettingsBuilder {
             addPlaygroundSettings(specLookupContext.platform)
         }
 
+        if let project, project.isPackage {
```
owenv, line 1576: We should restrict this to Apple platforms, since ARC is not available on Windows/Linux/etc. I think it would also make sense to push it somewhere user-supplied configuration is still able to override it - maybe in `addPlatformSettings` or similar

Sources/SWBCore/Settings/Settings.swift
```diff
@@ -2436,6 +2436,13 @@ private class SettingsBuilder {
             platformTable.push(macro, realArchs)
         }
 
+        // Add ObjC ARC support for Apple platforms.
+        if core.hostOperatingSystem == .macOS {
```
owenv, line 2440: Thanks for moving this into the platform settings! This check should use the targeted platform instead of the host platform, since the build system may be cross-compiling from macOS to a non-Apple platform. We also want to ensure this applies when the destination is iOS/watchOS/tvOS/visionOS

Rajveer100: Sure.

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv (approved):
This looks good to me, thanks!

Rajveer100:
Could you land this for me?!

PR #193, opened 2025-02-19
Set TOOLCHAIN_VERSION appropriately when a toolchain is primary

owenv:
This allows us to pick up the right value in the Windows Swift spec needed to set PATH


owenv:
@swift-ci test

neonichu (approved):

PR #160, opened 2025-02-14
Add a LIBTOOL_USE_RESPONSE_FILE build setting

jakepetroules:
This will be used to control whether response files are used when invoking libtool, as some platforms' implementations (like FreeBSD) may not support response files.

PR #194, opened 2025-02-19
fix llvm-lib response file configuration on Windows

owenv:
Otherwise we'll fail to create the response file


owenv:
@swift-ci test

neonichu (approved):

jakepetroules:
Thanks, I missed this in #160

PR #195, opened 2025-02-19
Adjust non-darwin default toolchain lookup

owenv:
Improve the non-darwin fallback behavior of default toolchain lookup to fix a few more windows tests


owenv:
@swift-ci test

PR #198, opened 2025-02-20
Remove customized missing framework diagnostics

jakepetroules:
This is some very old code that modifies the wording of some diagnostics from the compiler and linker that are emitted when a framework or module is not found. It's not that useful, adds significant complexity and overhead, and would require significant refactoring to make some of the call sites properly asynchronous. We don't really need this functionality anymore, so remove it.


jakepetroules:
@swift-ci test

owenv (approved):

jakepetroules:
@swift-ci test

PR #201, opened 2025-02-20
Remove bogus categories from LD_NO_PIE build settings

jakepetroules:
The actual category is "Linking - General" as defined in CoreBuildSystem. The conflict blocks build settings documentation from being generated.


jakepetroules:
@swift-ci test

PR #202, opened 2025-02-20
Remove swift-crypto version constraint

jakepetroules:
Now that the fix for apple/swift-crypto#262 has been included in a tagged release, we can safely upgrade to newer versions. This also pulls in fixes for FreeBSD that will unblock that platform bringup.


jakepetroules:
@swift-ci test

owenv (approved):

PR #203, opened 2025-02-20
Make PlannedNode conform to CustomStringConvertible.

mhrawdon:
Some checking logic in the test infrastructure reports errors with lists of `PlannedNode`s, and it is much easier to triage those errors if the identity of the nodes are present in the error message.

Also changed one such method to include this information rather than raw counts.


mhrawdon:
@swift-ci test

owenv (approved):

jakepetroules (approved):

PR #204, opened 2025-02-20
Fix Swift AST path computation in the presence of compatibility archs

owenv:
When computing paths for linker args propagating Swift debug info when statically linking, ensure we do so for both archs and compatibility archs if needed


owenv:
@swift-ci test

jakepetroules:
Should this have some tests?

jakepetroules (approved):

PR #207, opened 2025-02-21
Change Windows linker to lld(llvm linker)

kcieplak:
Force ALTERNATE_LINKER to lld-link for the Windows platform.

* Add linker tests for setting ALTERNATE_LINKER.  Test a few different combinations, default, invalid and specific linkers.
* Tests expose some linking issues (Windows) that will need to be fixed.
* Fix the regular expression for detecting lld linker version as windows does not have a space. i.e. 
`LLD 17.0.6` 
`LLD 17.0.0 (compatible with GNU linkers)`

### DIFF
--- a/Sources/SWBCore/Specs/Tools/LinkerTools.swift
+++ b/Sources/SWBCore/Specs/Tools/LinkerTools.swift
@@ -1234,17 +1234,44 @@ public final class LdLinkerSpec : GenericLinkerSpec, SpecIdentifierType, @unchec
 
     override public func discoveredCommandLineToolSpecInfo(_ producer: any CommandProducer, _ scope: MacroEvaluationScope, _ delegate: any CoreClientTargetDiagnosticProducingDelegate) async -> (any DiscoveredCommandLineToolSpecInfo)? {
         let alternateLinker = scope.evaluate(BuiltinMacros.ALTERNATE_LINKER)
-
         // The ALTERNATE_LINKER is the 'name' of the linker not the executable name, clang will find the linker binary based on name passed via -fuse-ld, but we need to discover
-        // its properties by executing the actual binary. On unix based oses the linkers are installed as ld.<name> on windows its <name>.exe
+        // its properties by executing the actual binary. There is a common filename when the linker is not "ld" across all platforms using "ld.<ALTERNAME_LINKER>(.exe)"
+        // macOS (Xcode SDK)
+        // -----------------
+        // ld
+        // ld-classic
+        //
+        // macOS (Open Source)
+        // -----------
+        // ld.lld -> lld
+        // ld64.lld -> lld
+        // lld
+        // lld-link -> lld
+        //
+        // Linux
+        // ------
+        // /usr/bin/ld -> aarch64-linux-gnu-ld
+        // /usr/bin/ld.bfd -> aarch64-linux-gnu-ld.bfd
+        // /usr/bin/ld.gold -> aarch64-linux-gnu-ld.gold
+        // /usr/bin/ld.lld -> lld
+        // /usr/bin/ld64.lld -> lld
+        // /usr/bin/lld
+        // /usr/bin/lld-link -> lld
+        // /usr/bin/gold -> aarch64-linux-gnu-gold
+        //
+        // Windows
+        // -------
+        // ld.lld.exe
+        // ld64.lld.exe
+        // lld-link.exe
+        // lld.exe
+        // link.exe //In Visual Studio
+        //
+        // Note: On Linux you cannot invoke the llvm linker by the direct name for determining the version,
+        // you need to use ld.<ALTERNATE_LINKER>
         var linkerPath = Path("ld")
-        if alternateLinker != "" {
-            linkerPath =
-                switch producer.hostOperatingSystem {
-                        case .linux: Path("ld.\(alternateLinker)")
-                        case .windows: Path("\(alternateLinker).exe")
-                        default : Path("\(alternateLinker)")
-                }
+        if alternateLinker != "" && alternateLinker != "ld" {
+            linkerPath = Path(producer.hostOperatingSystem.imageFormat.executableName(basename: "ld.\(alternateLinker)"))
         }
         // Create the cache key.  This is just the path to the linker we would invoke if we were invoking the linker directly.
         guard let toolPath = producer.executableSearchPaths.lookup(linkerPath) else {
--- a/Sources/SWBCore/Specs/Tools/LinkerTools.swift
+++ b/Sources/SWBCore/Specs/Tools/LinkerTools.swift
@@ -177,6 +177,7 @@ public struct DiscoveredLdLinkerToolSpecInfo: DiscoveredCommandLineToolSpecInfo 
     let undefinedSymbolCountLimit = 100
 
     override func parseLine<S: Collection>(_ lineBytes: S) -> Bool where S.Element == UInt8 {
+
         // Create a string that we can examine.  Use the non-failable constructor, so that we are robust against potentially invalid UTF-8.
         let lineString = String(decoding: lineBytes, as: Unicode.UTF8.self)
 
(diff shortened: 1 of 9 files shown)


kcieplak:
@swift-ci test

Tests/SWBBuildSystemTests/LinkerTests.swift
```diff
@@ -153,4 +156,184 @@ fileprivate struct LinkerTests: CoreBasedTests {
...
+    /// There is no reliable way to determine from a final linked binary which
+    /// linker was used, so the test enables some verbosity to see which linker
+    /// clang invokes.
+    /// Note: There is an output parser in the LinkerTool spec that does
+    ///       error parsing and creates a new build error diagnostic with
+    ///       a capaitalized error snippet, so this needs to be handled.
+    @Test(.requireSDKs(.host))
+    func alternateLinkerSelection() async throws {
+
+        // struct Linker {
```
owenv, line 174: Can this be dropped?

Tests/SWBBuildSystemTests/LinkerTests.swift
```diff
@@ -153,4 +156,184 @@ fileprivate struct LinkerTests: CoreBasedTests {
...
+        //     let name: String
+        //     let exec: Path
+        //     let validOn: [RunDestinationInfo]
+        // }
+
+        let runDestination: RunDestinationInfo = .host
+        // let defaultLinker = try await runDestination == .windows ? Linker(name: "link", exec: self.linkExePath, validOn: [.windows]) : Linker(name: "ld", exec: self.ldPath, validOn: [.macOS, .linux])
```
owenv, line 181: same here

kcieplak:
@swift-ci test

kcieplak:
Linux leg hung in testing.
Retriever
@swift-ci test

Sources/SWBCore/Specs/Tools/LinkerTools.swift
```diff
@@ -1240,9 +1241,18 @@ public final class LdLinkerSpec : GenericLinkerSpec, SpecIdentifierType, @unchec
...
-        let linkerPath = if alternateLinker != "" { Path(alternateLinker) } else { Path("ld") }
-
-        // Create the cache key.  This is just the path to the ld linker we would invoke if we were invoking the linker directly.
+        // The ALTERNATE_LINKER is the 'name' of the linker not the executable name, clang will find the linker binary based on name passed via -fuse-ld, but we need to discover
+        // its properties by executing the actual binary. On unix based oses the linkers are installed as ld.<name> on windows its <name>.exe
+        var linkerPath = Path("ld")
+        if alternateLinker != "" {
+            linkerPath =
+                switch producer.hostOperatingSystem {
+                        case .linux: Path("ld.\(alternateLinker)")
```
jakepetroules, line 1250: This looks like it shouldn't be Linux-specific. This doesn't handle macOS, FreeBSD, etc.

Should it just be if-windows-else?

Tests/SWBBuildSystemTests/LinkerTests.swift
```diff
@@ -153,4 +156,175 @@ fileprivate struct LinkerTests: CoreBasedTests {
...
+                    results.checkError(.contains("Unable to execute command: program not executable"))
+                    results.checkError(.contains("unable to execute command: program not executable"))
+                    results.checkError(.contains("Linker command failed with exit code 1"))
+                }
+                results.checkNoDiagnostics()
+            }
+
+
+            // lld - llvm linker
+            if runDestination != .macOS {
```
jakepetroules, line 269: Why excluding this check on macOS?

Presently Xcode toolchains don't ship LLD, but I think swift.org toolchains do. Could the check exclude XcodeDefault toolchain only?

jakepetroules (changes requested):

compnerd (commented):
How does one switch the linker after this change? This is roughly what we are currently doing with SPM specifically. We have a SDKSettings.plist that adds the `-use-ld=lld` option with the default flags, so I'm not conceptually against the change. I just want to ensure that we still have the means to change to link to test/workaround linker issues.

jakepetroules:
> How does one switch the linker after this change? This is roughly what we are currently doing with SPM specifically. We have a SDKSettings.plist that adds the `-use-ld=lld` option with the default flags, so I'm not conceptually against the change. I just want to ensure that we still have the means to change to link to test/workaround linker issues.

Raw linker flags should be removed from SDKSettings.plist later, and the linker can be overridden if needed by setting this new `ALTERNATE_LINKER` build setting in the PIF that SwiftPM will generate to communicate the project model to Swift Build.

compnerd:
> Raw linker flags should be removed from SDKSettings.plist later, and the linker can be overridden if needed by setting this new ALTERNATE_LINKER build setting in the PIF that SwiftPM will generate to communicate the project model to Swift Build.

I think that "later" is going to be significantly later. That really shouldn't happen until after the removal of current build engine. IMO, if SPM is using swift-build as the build engine, it should not consult `SDKSettings.plist` at all. As long as we surface a means for the user to allow the setting to be changed in the PIF, that is sufficient. When the old engine is removed, we would drop `SDKSettings.plist` entirely as it currently serves only to ship the default settings for the platform as we did not have the configuration knobs with the old build system.

jakepetroules:
> I think that "later" is going to be significantly later. That really shouldn't happen until after the removal of current build engine. IMO, if SPM is using swift-build as the build engine, it should not consult SDKSettings.plist at all.

Agree, SwiftPM shouldn't directly consult SDKSettings once the new engine is in place.

> When the old engine is removed, we would drop SDKSettings.plist entirely as it currently serves only to ship the default settings for the platform as we did not have the configuration knobs with the old build system.

Yeah, probably. We still need to design how non-Apple platforms will be brought into Swift Build in the long term, as the current SDK/platforms infrastructure is fairly tied to the SDKSettings.plist model for Apple platforms. Also ties in to Swift SDKs support (#6).

dschaefer2:
> Also ties in to Swift SDKs support

Yes, I'd like to see some convergence on how we handle the Windows SDK and Swift SDKs. I'm not sure what that means yet but it's hard to understand what's happening under the covers to set up the Windows SDK.

jakepetroules:
```sh
✘ Test alternateLinkerSelection() recorded an issue at CoreBasedTests.swift:264:24: Expectation failed: (defaultToolchain.executableSearchPaths → SWBCore.StackedSearchPath).findExecutable(operatingSystem: (core.hostOperatingSystem → .linux), basename: "ld") → nil
↳ couldn't find ld in default toolchain
✘ Test alternateLinkerSelection() failed after 2.029 seconds with 1 issue.
```

Sources/SWBCore/Specs/Tools/LinkerTools.swift
kcieplak: The naming of the linker is not consistent across platforms, so going from the value of ALTERNATE_LINKER (clang uses the term 'linker flavor')to the binary is tricky.  

I added a large comment, but we can use "ld.<ALTERNATE_LINKER>" as the file name in all cases except where the ALTERNATE_LINKER is "ld"

This logic eliminates the check for running platform os.

Tests/SWBBuildSystemTests/LinkerTests.swift
kcieplak: I updated the tests to determine the linker binary path and use that as the criteria to run the test, so on the OSS version will will find usr/bin/ld.ldd and the test will run.

jakepetroules: Resolved in the latest commit

kcieplak:
> ```shell
> ✘ Test alternateLinkerSelection() recorded an issue at CoreBasedTests.swift:264:24: Expectation failed: (defaultToolchain.executableSearchPaths → SWBCore.StackedSearchPath).findExecutable(operatingSystem: (core.hostOperatingSystem → .linux), basename: "ld") → nil
> ↳ couldn't find ld in default toolchain
> ✘ Test alternateLinkerSelection() failed after 2.029 seconds with 1 issue.
> ```

Yes in swift-ci we do not seem to have /usr/bin/ld.  Not sure why as running container from docker "swiftlang/swift:nightly-jammy", we have a multitude of linkers installed.

I have a new patch up that checks which linkers are installed and that governs which tests can run.

Tests/SWBBuildSystemTests/LinkerTests.swift
```diff
@@ -229,22 +229,33 @@ fileprivate struct LinkerTests: CoreBasedTests {
                 """
             }
 
+            // // Try to find the installed linkers
```
jakepetroules, line 232: nit: double `//`

Tests/SWBBuildSystemTests/LinkerTests.swift
```diff
@@ -229,22 +229,33 @@ fileprivate struct LinkerTests: CoreBasedTests {
...
+            let ldLinkerPath = try? await self.ldPath
```
jakepetroules, line 233: I'd leave the `try?` as `try` -- if this throws, the Core failed to initialize, and the whole test is bogus.

Sources/SWBCore/Specs/Tools/LinkerTools.swift
```diff
@@ -1240,17 +1240,44 @@ public final class LdLinkerSpec : GenericLinkerSpec, SpecIdentifierType, @unchec
...
+        //
+        // Windows
+        // -------
+        // ld.lld.exe
+        // ld64.lld.exe
+        // lld-link.exe
+        // lld.exe
+        // link.exe //In Visual Studio
+        //
+        // Note: On Linux you cannot invoke the llvm linker by the direct name for determining the version,
```
jakepetroules, line 1276: On Linux? Or on any Unix-like?

kcieplak: I only tried ubuntu, I did not try others

```
/usr/bin/lld
lld is a generic driver.
Invoke ld.lld (Unix), ld64.lld (macOS), lld-link (Windows), wasm-ld (WebAssembly) instead
```

kcieplak:
@swift-ci test

kcieplak:
Still have issues with tests on swift-ci Linux environment.   Seems we download a toolchains then setup some environment variables. I will have to replicate that to see what is going on.  

` export SWIFT_DRIVER_SWIFT_EXEC=/home/build-user/latest_toolchain/swift-DEVELOPMENT-SNAPSHOT-2025-02-20-a-ubuntu20.04/usr/bin/swiftc; export SWIFT_DRIVER_SWIFT_FRONTEND_EXEC=/home/build-user/latest_toolchain/swift-DEVELOPMENT-SNAPSHOT-2025-02-20-a-ubuntu20.04/usr/bin/swift-frontend;`

kcieplak:
@swift-ci test

kcieplak:
@swift-ci test linux

kcieplak:
@swift-ci  test linux

kcieplak:
@swift-ci test linux

kcieplak:
@swift-ci test linux

jakepetroules (approved):

PR #210, opened 2025-02-22
fix: `belongsToPreferredArch` anywhere in the package

MojtabaHs:
The root property is `package` protected and may need introducing a temporary alias. Please let me know if we should
- Add a soft-deprecated property with the old name and `@available` [like this](https://github.com/swiftlang/swift-build/pull/173#issuecomment-2673601114)
- Add a computed property with the old name and a basic `\\ comment` [like this](https://github.com/swiftlang/swift-build/pull/173#issuecomment-2673260832)
- Or if it’s fine as is.

Please don’t forget to run the @swift-ci test and let me know if any further changes are needed.
Thanks!


jakepetroules (approved):
@swift-ci test

No deprecation warnings needed; these are all internal-only APIs.

jakepetroules:
@swift-ci test

PR #213, opened 2025-02-23
Fix all remaining spelling errors throughout the codebase

jakepetroules:
...to the best of my knowledge.

I used cspell (`brew install cspell`) by running `cspell lint -u --words-only --no-progress "**/*" | sort > words.txt`, and then manually scanned through the list of around 2000 derived errors, skipping past technical terms or compound words that were valid but not in cspell's dictionaries.

This also standardized the spelling of a handful of words to US English, and tweaked the grammar of the surrounding sentence structure in a few cases where obscure constructed or slang words were used.

The majority of changes are in comments only, and most others in local variable names or test function names.

PR #215, opened 2025-02-24
ci: add auto spell check for any PRs and main branch changes

MojtabaHs:
Based on the discussion in [this PR](https://github.com/swiftlang/swift-build/pull/213#discussion_r1968123920), I’ve added cspell as a GitHub Action to automatically check for typos in PRs, making them harder to miss.

For now, it uses the tool’s default configuration, but we can adjust it as needed. It runs only on changed files, ignoring pre-existing issues, which should be reviewed manually to avoid unintended breaking changes (IMO).

### DIFF
--- a/.github/workflows/spellcheck.yaml
+++ b/.github/workflows/spellcheck.yaml
@@ -1,0 +1,11 @@
+name: 'Check spelling'
+on: # rebuild any PRs and main branch changes
+  pull_request:
+  push:
+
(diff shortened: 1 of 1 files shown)


jakepetroules:
Thanks for taking the time to look into this, but at this time we think the reward may not justify the maintenance burden and we'd prefer not to take this change, so I'm going to close this for now.

PR #216, opened 2025-02-24
Don't crash when attempting single-file builds with multiple targets in the build request

jakepetroules:
Instead, when a single file preprocess or assemble request is made, incorporate the results across multiple architectures and target contexts instead of erroring out. The client can then decide how to present the results. Previously, the client was expected to limit the request to a single target.

rdar://145323187

### DIFF
--- a/Sources/SWBBuildSystem/BuildManager.swift
+++ b/Sources/SWBBuildSystem/BuildManager.swift
@@ -59,9 +59,22 @@ package actor BuildManager {
 
         let buildOutputMap: [String:String]?
         if let buildOnlyThesePaths {
-            precondition(request.buildTargets.count == 1, "the client should ensure we have one target only")
-            let buildOnlyTheseOutputs = buildRequestContext.computeOutputPaths(for: buildOnlyThesePaths, workspace: workspaceContext.workspace, target: request.buildTargets.first!, command: request.buildCommand, parameters: request.parameters)
-            buildOutputMap = Dictionary(uniqueKeysWithValues: zip(buildOnlyTheseOutputs, buildOnlyThesePaths.map{ $0.str }))
+            buildOutputMap = {
+                var outputMap: [String: String] = [:]
+                for target in request.buildTargets {
+                    for buildOnlyThisPath in buildOnlyThesePaths {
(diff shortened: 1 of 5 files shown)


jakepetroules:
@swift-ci test

Sources/SWBCore/BuildRequestContext.swift
```diff
@@ -134,8 +134,8 @@ extension BuildRequestContext {
         }
     }
 
-    /// Compute output paths for a list of source files in a specific target.
-    public func computeOutputPaths(for inputPaths: [Path], workspace: Workspace, target: BuildRequest.BuildTargetInfo, command: BuildCommand, parameters: BuildParameters? = nil) -> [String] {
+    /// Compute output paths for a source file in a specific target. There may be multiple results if the build is a multi-arch build.
+    public func computeOutputPaths(for inputPath: Path, workspace: Workspace, target: BuildRequest.BuildTargetInfo, command: BuildCommand, parameters: BuildParameters? = nil) -> [String] {
```
jakepetroules, line 138: This is no longer a many-to-many because that leaves the caller unable to distinguish which outputs corresponded to which inputs.

Sources/SWBBuildSystem/BuildManager.swift
```diff
@@ -59,9 +59,22 @@ package actor BuildManager {
 
         let buildOutputMap: [String:String]?
         if let buildOnlyThesePaths {
-            precondition(request.buildTargets.count == 1, "the client should ensure we have one target only")
-            let buildOnlyTheseOutputs = buildRequestContext.computeOutputPaths(for: buildOnlyThesePaths, workspace: workspaceContext.workspace, target: request.buildTargets.first!, command: request.buildCommand, parameters: request.parameters)
-            buildOutputMap = Dictionary(uniqueKeysWithValues: zip(buildOnlyTheseOutputs, buildOnlyThesePaths.map{ $0.str }))
```
jakepetroules, line 64: This usage of zip could lead to silent data loss because there was no guarantee the arrays were of the same length.

mhrawdon (approved):

jakepetroules:
@swift-ci test macOS

jakepetroules:
@swift-ci test macOS

owenv (approved):

jakepetroules:
@swift-ci test macOS

PR #220, opened 2025-02-25
Jobs planned by SwiftDriver should not re-request driver-reported inputs

owenv:
SwiftDriver has exclusive control over the incrementality of frontend tasks used to compile a module. We should not redundantly re-request the inputs it reports for individual jobs, as those paths may be remapped by a VFS manually passed to the driver and handled internally.

I also did some refactoring as part of this PR to remove the taskInputs property on DynamicTask. Modeling inputs as part of the task key instead makes requesting them at execution time less error-prone, and eliminates some boilerplate from dynamic tasks which don't need to request any inputs.

Resolves https://github.com/swiftlang/swift-driver/issues/1705 / rdar://143985963


owenv:
@swift-ci test

artemcm (approved):

neonichu (approved):

chiragramani:
This is great, thank you, @owenv ! 🙏🏼

PR #221, opened 2025-02-25
Add versioning to task action cache key

owenv:
This can be used as a tool to force invalidation when the implementation changes

rdar://142442134

### DIFF
--- a/Sources/SWBTaskExecution/TaskActions/GenericCachingTaskAction.swift
+++ b/Sources/SWBTaskExecution/TaskActions/GenericCachingTaskAction.swift
@@ -21,6 +21,9 @@ public final class GenericCachingTaskAction: TaskAction {
         return "generic-caching"
     }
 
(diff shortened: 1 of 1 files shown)


owenv:
@swift-ci test

neonichu (approved):

Sources/SWBTaskExecution/TaskActions/GenericCachingTaskAction.swift
```diff
@@ -320,11 +326,12 @@ fileprivate struct CacheKey {
             throw StubError.error("Could not load cached value")
         }
         let deserializer = MsgPackDeserializer(casObject.data)
-        try deserializer.beginAggregate(2)
+        try deserializer.beginAggregate(4)
```
jakepetroules, line 329: Huh, what happened here? store went from 3 to 4 but load went from 2 to 4. Was this just completely broken before?

owenv: It was a bug

PR #225, opened 2025-02-25
Make some tests more resilient to Swift explicit modules status

owenv:
Either explicitly enable/disable the feature where appropriate, or adapt test expectations to pass in both modes

rdar://145166637


owenv:
@swift-ci test

jakepetroules (approved):

PR #226, opened 2025-02-26
Ensure we do not block reading the input pipe on a Swift concurrency thread

owenv:
I think this may be the cause of intermittent linux test hangs


owenv:
@swift-ci test Linux

owenv:
@swift-ci test Linux

owenv:
This doesn't work

PR #232, opened 2025-02-26
Prefer the copy of Swift Testing in the toolchain for importing and linking, if available

stmontgomery:
This adds a special case to the logic which configures additional build settings when `ENABLE_TESTING_SEARCH_PATHS` is enabled so that it inserts an additional library search path to locate Swift Testing from the toolchain if it's present.

Motivation: This is one step towards a larger goal of ensuring that if a user installs an open source Swift.org toolchain and selects it in Xcode, the copy of Swift Testing in that toolchain will be preferred over the analogous copy (`Testing.framework`) built in to Xcode. This matches existing precedent with other Swift standard libraries, which are used from the toolchain when available.

Fixes rdar://145689942

### DIFF
--- a/Sources/SWBCore/Settings/Settings.swift
+++ b/Sources/SWBCore/Settings/Settings.swift
@@ -4217,6 +4217,14 @@ private class SettingsBuilder {
             table.push(BuiltinMacros.LIBRARY_SEARCH_PATHS, BuiltinMacros.namespace.parseStringList(["$(inherited)", "$(TEST_LIBRARY_SEARCH_PATHS$(TEST_BUILD_STYLE))"]))
             table.push(BuiltinMacros.SWIFT_INCLUDE_PATHS, BuiltinMacros.namespace.parseStringList(["$(inherited)", "$(TEST_LIBRARY_SEARCH_PATHS$(TEST_BUILD_STYLE))"]))
 
+            // If the toolchain contains a copy of Swift Testing, prefer it.
+            let toolchainPath = Path(scope.evaluateAsString(BuiltinMacros.TOOLCHAIN_DIR))
+            let platformName = scope.evaluate(BuiltinMacros.PLATFORM_NAME)
+            let toolchainTestingLibraryDir = toolchainPath.join("usr/lib/swift/\(platformName)/testing")
+            if workspaceContext.fs.exists(toolchainTestingLibraryDir) {
+                table.push(BuiltinMacros.TEST_LIBRARY_SEARCH_PATHS, BuiltinMacros.namespace.parseStringList(["$(inherited)", toolchainTestingLibraryDir.str]))
+            }
+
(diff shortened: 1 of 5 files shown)


stmontgomery:
@swift-ci test

neonichu (approved):

Sources/SWBCore/Settings/Settings.swift
```diff
@@ -4217,6 +4217,14 @@ private class SettingsBuilder {
             table.push(BuiltinMacros.LIBRARY_SEARCH_PATHS, BuiltinMacros.namespace.parseStringList(["$(inherited)", "$(TEST_LIBRARY_SEARCH_PATHS$(TEST_BUILD_STYLE))"]))
             table.push(BuiltinMacros.SWIFT_INCLUDE_PATHS, BuiltinMacros.namespace.parseStringList(["$(inherited)", "$(TEST_LIBRARY_SEARCH_PATHS$(TEST_BUILD_STYLE))"]))
 
+            // If the toolchain contains a copy of Swift Testing, prefer it.
+            let toolchainPath = Path(scope.evaluateAsString(BuiltinMacros.TOOLCHAIN_DIR))
+            let platformName = scope.evaluate(BuiltinMacros.PLATFORM_NAME)
+            let toolchainTestingLibraryDir = toolchainPath.join("usr/lib/swift/\(platformName)/testing")
+            if workspaceContext.fs.exists(toolchainTestingLibraryDir) {
```
owenv, line 4224: Instead of hitting the filesystem for an existence check here, could we check the toolchain ID to see if it's a swift.org toolchain? Or do you think that would be less reliable?

stmontgomery: One consideration is that the library isn't currently available for every platform, and as of this writing it's only present for `macosx` platform (plus non-Apple platforms). This way seemed a bit more flexible since it would begin working for other Apple platforms automatically if/when they are added to the toolchain.

jakepetroules: That seems like a reasonable argument to me, but I wonder if we could cache the existence in ToolchainRegistry and thread it through to here so we don't need to hit the FS potentially many times during settings construction?

stmontgomery:
@swift-ci test Linux

Sources/SWBCore/Settings/Settings.swift
jakepetroules: Perhaps you could add a new `Set<String>` property on `Toolchain` which is the list of platforms which have the testing assets, and in the constructor you can populate that (you have access to an `FSProxy` and can call `listdir`, etc.). Then in Settings you could look up the active toolchain and query its supported testing platforms.

stmontgomery:
@swift-ci test

Sources/SWBCore/Settings/Settings.swift
stmontgomery: I've implemented this, please take another look!

owenv (approved):

jakepetroules:
👍

PR #233, opened 2025-02-26
Add workspace name and path to the optional SWIFTBUILD_TRACE file

owenv:
This can be used to identify the top level workspace/project/package referenced by a trace entry.

rdar://144816105


owenv:
@swift-ci test

jakepetroules (approved):

PR #234, opened 2025-02-26
Add a build setting for xcode's EnableDebugActivityLogs

owenv:
Add an EMIT_FRONTEND_COMMAND_LINES build setting which can be used to opt into frontend command line emission. This is useful to be able to turn on debug logs in environments where user defaults aren't easily changes. In order to ensure this doesn't cause us to excessively rebuild modules, setting the build setting on at any level applicable to a compile task will enable it globally.

rdar://143331197


owenv:
@swift-ci test

jakepetroules (approved):
Feels a little odd as a build setting, but I suppose it's not the first example of something like this.

PR #238, opened 2025-02-27
Fix gold linker detection for Fedora/RHEL

kcieplak:
* Add additional regular expressions to match version strings on Fedora, RHEL, and Amazon Linux platforms.


kcieplak:
@swift-ci  test

kcieplak:
@swift-ci test

jakepetroules (approved):

PR #242, opened 2025-02-28
 Enable Windows linker discovery

kcieplak:
* Make Platform Registry initialization async
    * Change the function signature of
      additionalPlatformExecutableSearchPaths to allow
      passing of a filesystem for discovery. Plugins
      will need to be updated to match new signatures
    * Add the visual studio install directory to
      platform search paths for executables.
    * Enable Windows linker discovery to get link.exe
      type and version
    * Add Windows platform plugin extension to get
      install directory
    * Add test for discovery of windows linker

### DIFF
--- a/Sources/SWBCore/SDKRegistry.swift
+++ b/Sources/SWBCore/SDKRegistry.swift
@@ -646,14 +646,6 @@ public final class SDKRegistry: SDKRegistryLookup, CustomStringConvertible, Send
             tripleEnvironment = ""
         }
 
-        let archs: PropertyListItem =
-            switch operatingSystem {
-            case .windows:
-                .plArray([.plString("x86_64"), .plString("aarch64")])
-            default:
-                .plArray([.plString(Architecture.hostStringValue ?? "unknown")])
-            }
-
         return try [
             "Type": .plString("SDK"),
             "Version": .plString(Version(ProcessInfo.processInfo.operatingSystemVersion).zeroTrimmed.description),
@@ -664,7 +656,7 @@ public final class SDKRegistry: SDKRegistryLookup, CustomStringConvertible, Send
             ].merging(defaultProperties, uniquingKeysWith: { _, new in new })),
             "SupportedTargets": .plDict([
                 operatingSystem.xcodePlatformName: .plDict([
-                    "Archs": archs,
+                    "Archs": .plArray([.plString(Architecture.hostStringValue ?? "unknown")]),
                     "LLVMTargetTripleEnvironment": .plString(tripleEnvironment),
                     "LLVMTargetTripleSys": .plString(operatingSystem.xcodePlatformName),
                     "LLVMTargetTripleVendor": .plString("unknown"),
--- a/Sources/SWBCore/SDKRegistry.swift
+++ b/Sources/SWBCore/SDKRegistry.swift
@@ -646,6 +646,14 @@ public final class SDKRegistry: SDKRegistryLookup, CustomStringConvertible, Send
             tripleEnvironment = ""
         }
 
+        let archs: PropertyListItem =
+            switch operatingSystem {
+            case .windows:
+                .plArray([.plString("x86_64"), .plString("aarch64")])
+            default:
+                .plArray([.plString(Architecture.hostStringValue ?? "unknown")])
+            }
+
         return try [
             "Type": .plString("SDK"),
             "Version": .plString(Version(ProcessInfo.processInfo.operatingSystemVersion).zeroTrimmed.description),
@@ -656,7 +664,7 @@ public final class SDKRegistry: SDKRegistryLookup, CustomStringConvertible, Send
             ].merging(defaultProperties, uniquingKeysWith: { _, new in new })),
             "SupportedTargets": .plDict([
                 operatingSystem.xcodePlatformName: .plDict([
-                    "Archs": .plArray([.plString(Architecture.hostStringValue ?? "unknown")]),
+                    "Archs": archs,
                     "LLVMTargetTripleEnvironment": .plString(tripleEnvironment),
                     "LLVMTargetTripleSys": .plString(operatingSystem.xcodePlatformName),
                     "LLVMTargetTripleVendor": .plString("unknown"),
--- a/Sources/SWBCore/Settings/BuiltinMacros.swift
+++ b/Sources/SWBCore/Settings/BuiltinMacros.swift
@@ -803,6 +803,8 @@ public final class BuiltinMacros {
     public static let LD_RUNPATH_SEARCH_PATHS = BuiltinMacros.declareStringListMacro("LD_RUNPATH_SEARCH_PATHS")
     public static let LD_SDK_IMPORTS_FILE = BuiltinMacros.declarePathMacro("LD_SDK_IMPORTS_FILE")
     public static let LD_WARN_UNUSED_DYLIBS = BuiltinMacros.declareBooleanMacro("LD_WARN_UNUSED_DYLIBS")
+    public static let LD_MULTIARCH = BuiltinMacros.declareBooleanMacro("LD_MULTIARCH")
+    public static let LD_MULTIARCH_PREFIX_MAP = BuiltinMacros.declareStringListMacro("LD_MULTIARCH_PREFIX_MAP")
     public static let LEX = BuiltinMacros.declarePathMacro("LEX")
     public static let LEXFLAGS = BuiltinMacros.declareStringListMacro("LEXFLAGS")
     public static let LIBRARIAN = BuiltinMacros.declareStringMacro("LIBRARIAN")
@@ -1858,6 +1860,8 @@ public final class BuiltinMacros {
         LD_RUNPATH_SEARCH_PATHS,
         LD_SDK_IMPORTS_FILE,
         LD_WARN_UNUSED_DYLIBS,
+        LD_MULTIARCH,
+        LD_MULTIARCH_PREFIX_MAP,
         LEGACY_DEVELOPER_DIR,
         LEX,
         LEXFLAGS,
--- a/Sources/SWBCore/Settings/BuiltinMacros.swift
+++ b/Sources/SWBCore/Settings/BuiltinMacros.swift
@@ -803,8 +803,8 @@ public final class BuiltinMacros {
     public static let LD_RUNPATH_SEARCH_PATHS = BuiltinMacros.declareStringListMacro("LD_RUNPATH_SEARCH_PATHS")
     public static let LD_SDK_IMPORTS_FILE = BuiltinMacros.declarePathMacro("LD_SDK_IMPORTS_FILE")
     public static let LD_WARN_UNUSED_DYLIBS = BuiltinMacros.declareBooleanMacro("LD_WARN_UNUSED_DYLIBS")
-    public static let LD_MULTIARCH = BuiltinMacros.declareBooleanMacro("LD_MULTIARCH")
-    public static let LD_MULTIARCH_PREFIX_MAP = BuiltinMacros.declareStringListMacro("LD_MULTIARCH_PREFIX_MAP")
+    public static let _LD_MULTIARCH = BuiltinMacros.declareBooleanMacro("_LD_MULTIARCH")
+    public static let _LD_MULTIARCH_PREFIX_MAP = BuiltinMacros.declareStringListMacro("_LD_MULTIARCH_PREFIX_MAP")
     public static let LEX = BuiltinMacros.declarePathMacro("LEX")
     public static let LEXFLAGS = BuiltinMacros.declareStringListMacro("LEXFLAGS")
     public static let LIBRARIAN = BuiltinMacros.declareStringMacro("LIBRARIAN")
@@ -1860,8 +1860,8 @@ public final class BuiltinMacros {
         LD_RUNPATH_SEARCH_PATHS,
         LD_SDK_IMPORTS_FILE,
         LD_WARN_UNUSED_DYLIBS,
-        LD_MULTIARCH,
-        LD_MULTIARCH_PREFIX_MAP,
+        _LD_MULTIARCH,
+        _LD_MULTIARCH_PREFIX_MAP,
         LEGACY_DEVELOPER_DIR,
         LEX,
         LEXFLAGS,
--- a/Sources/SWBCore/SpecImplementations/Tools/LinkerTools.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/LinkerTools.swift
@@ -1273,7 +1273,6 @@ public final class LdLinkerSpec : GenericLinkerSpec, SpecIdentifierType, @unchec
     }
 
     override public func discoveredCommandLineToolSpecInfo(_ producer: any CommandProducer, _ scope: MacroEvaluationScope, _ delegate: any CoreClientTargetDiagnosticProducingDelegate) async -> (any DiscoveredCommandLineToolSpecInfo)? {
-        let alternateLinker = scope.evaluate(BuiltinMacros.ALTERNATE_LINKER)
         // The ALTERNATE_LINKER is the 'name' of the linker not the executable name, clang will find the linker binary based on name passed via -fuse-ld, but we need to discover
         // its properties by executing the actual binary. There is a common filename when the linker is not "ld" across all platforms using "ld.<ALTERNAME_LINKER>(.exe)"
         // macOS (Xcode SDK)
@@ -1309,9 +1308,43 @@ public final class LdLinkerSpec : GenericLinkerSpec, SpecIdentifierType, @unchec
         //
         // Note: On Linux you cannot invoke the llvm linker by the direct name for determining the version,
         // you need to use ld.<ALTERNATE_LINKER>
-        var linkerPath = producer.hostOperatingSystem == .windows ? Path("link") : Path("ld")
-        if alternateLinker != "" && alternateLinker != "ld" {
+        let alternateLinker = scope.evaluate(BuiltinMacros.ALTERNATE_LINKER)
+        let isLinkerMultiarch = scope.evaluate(BuiltinMacros.LD_MULTIARCH)
+
+        var linkerPath = producer.hostOperatingSystem == .windows ? Path("ld.lld") : Path("ld")
+        if alternateLinker != "" && alternateLinker != "ld" && alternateLinker != "link" {
             linkerPath = Path(producer.hostOperatingSystem.imageFormat.executableName(basename: "ld.\(alternateLinker)"))
+        } else if alternateLinker != "" {
+            linkerPath =  Path(alternateLinker)
+        }
+        // If the linker does not support multiple architectures update the path to include a subfolder based on the prefix map
+        // to find the architecture specific executable.
+        if !isLinkerMultiarch {
+            let archMap = scope.evaluate(BuiltinMacros.LD_MULTIARCH_PREFIX_MAP)
+            let archMappings = archMap.reduce(into: [String: String]()) { mappings, map in
+                let split = map.components(separatedBy: ":")
+                if !split.isEmpty {
+                    return mappings[split[0]] = split[1]
+                }
+            }
+            if archMappings.isEmpty {
+                delegate.error("LD_MULTIARCH is 'false', but no prefix mappings are present in LD_MULTIARCH_PREFIX_MAP")
+                return nil
+            }
+            // Linkers that don't support multiple architectures cannot support universal binaries, so ARCHS will
+            // contain the target architecture and can only be a single value.
+            let arch = scope.evaluate(BuiltinMacros.ARCHS)
+            if arch.count > 1 {
+                delegate.error("LD_MULTIARCH is 'false', but multiple ARCHS have been given, this is invalid")
+                return nil
+            }
+            if let prefix = archMappings[arch[0]] {
+                // Add in the target architecture prefix directory to path for search.
+                linkerPath = Path(prefix).join(linkerPath)
+            } else {
+                delegate.error("Could not find prefix mapping for \(arch[0]) in LD_MULTIARCH_PREFIX_MAP")
+                return nil
+            }
         }
         guard let toolPath = producer.executableSearchPaths.findExecutable(operatingSystem: producer.hostOperatingSystem, basename: linkerPath.str) else {
             return nil
--- a/Sources/SWBCore/SpecImplementations/Tools/LinkerTools.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/LinkerTools.swift
@@ -1309,15 +1309,14 @@ public final class LdLinkerSpec : GenericLinkerSpec, SpecIdentifierType, @unchec
         //
         // Note: On Linux you cannot invoke the llvm linker by the direct name for determining the version,
         // you need to use ld.<ALTERNATE_LINKER>
-        var linkerPath = Path("ld")
+        var linkerPath = producer.hostOperatingSystem == .windows ? Path("link") : Path("ld")
         if alternateLinker != "" && alternateLinker != "ld" {
             linkerPath = Path(producer.hostOperatingSystem.imageFormat.executableName(basename: "ld.\(alternateLinker)"))
         }
-        // Create the cache key.  This is just the path to the linker we would invoke if we were invoking the linker directly.
-        guard let toolPath = producer.executableSearchPaths.lookup(linkerPath) else {
+        guard let toolPath = producer.executableSearchPaths.findExecutable(operatingSystem: producer.hostOperatingSystem, basename: linkerPath.str) else {
             return nil
         }
-
+        // Create the cache key.  This is just the path to the linker we would invoke if we were invoking the linker directly.
         return await discoveredLinkerToolsInfo(producer, delegate, at: toolPath)
     }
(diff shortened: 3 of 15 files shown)


kcieplak:
@swift-ci test

kcieplak:
@swift-ci test

kcieplak:
Making a draft as there seems to be issues where the registry property is accessed before initialized, but only in CI

kcieplak:
@swift-ci test

kcieplak:
@swift-ci test

Tests/SWBBuildSystemTests/LinkerTests.swift
```diff
@@ -247,7 +249,14 @@ fileprivate struct LinkerTests: CoreBasedTests {
                     results.checkTaskOutput(task) { taskOutput in
                         results.checkTaskOutput(task) { output in
                             // Expect that one of the installed linkers is used, we are not sure which one.
-                            #expect(installedLinkerPaths.map { $0.str }.contains(where: output.asString.contains))
+                            if runDestination == .windows && Architecture.hostStringValue == "aarch64" {
+                                withKnownIssue("'clang' picks the wrong binary for link.exe using x86 version") {
+                                    // On windows aarch64 'clang' picks the wrong host architecture for link.exe, choosing "MSVC\14.41.34120\bin\Hostx86\arm64\link.exe"
```
jakepetroules, line 254: Is this a clang bug or we're passing in the wrong environment? Can we get an issue filed for this if needed?

kcieplak: I looked at my PATH and the arm64 path is the only path I can see, yet if you invoke clang with a fake.o to link you can see it picks the x86 host 

`C:\Users\kcieplak\swift-build>clang -v foobar.o
clang version 17.0.6
Target: aarch64-unknown-windows-msvc
Thread model: posix
InstalledDir: C:\Users\kcieplak\AppData\Local\Programs\Swift\Toolchains\6.0.3+Asserts\usr\bin
 "C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\Tools\\MSVC\\14.41.34120\\bin\\Hostx86\\arm64\\link.exe"`

And my PATH
`Path=C:\Program Files\Parallels\Parallels Tools\Applications;C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\;C:\Windows\System32\OpenSSH\;C:\Program Files (x86)\Windows Kits\10\Windows Performance Toolkit\;C:\Program Files\Git\cmd;C:\Program Files\WindowsPowerShell\Scripts;C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.41.34120\bin\Hostarm64\arm64\`

I will open an issue for this

jakepetroules: What if you pass an explicit target triple to clang? Like `clang -target aarch64-unknown-windows-pc`

kcieplak: Still the incorrect host linker.
`C:\Users\kcieplak\swift-build>clang -v -target aarch64-unknown-windows-pc foobar.o
clang version 17.0.6
Target: aarch64-pc-windows-msvc
Thread model: posix
InstalledDir: C:\Users\kcieplak\AppData\Local\Programs\Swift\Toolchains\6.0.3+Asserts\usr\bin
 "C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\Tools\\MSVC\\14.41.34120\\bin\\Hostx86\\arm64\\link.exe" `

jakepetroules: I'm not sure why it's picking Hostx86 for you, but that's not a problem as long as the target arch (arm64) is correct, which it is here.

Can you update the PR to pass in the target triple?

kcieplak: Agreed this is not really a problem as long as you have 'Prism' emulation enabled, the x86 binary will execute and correctly link to the target arm64 architecture.  At worst it will be slightly less performant as emulation will need to run.

kcieplak:
@swift-ci test

kcieplak:
There are failures for both the macOS and Windows test pipelines.  These failures are occurring in other PRs as well (https://ci-external.swift.org/job/swift-build-pull-request-windows/177/changes) , so not isolated to my changes.  In the Windows case the toolchain executables seem to just be exiting with signal 2.  macOS seems to have issues with the compiler driver.

kcieplak:
@swift-ci test

kcieplak:
macOS failures are due to swift-driver
https://github.com/swiftlang/swift-driver/pull/1826

kcieplak:
@swift-ci test

kcieplak:
@swift-ci test

Sources/SWBWindowsPlatform/Plugin.swift
```diff
@@ -25,22 +26,42 @@ struct WindowsPlatformSpecsExtension: SpecificationsExtension {
     }
 }
 
+private func findLatestInstallDirectory(fs: any FSProxy) async throws -> Path? {
+    if try ProcessInfo.processInfo.hostOperatingSystem() == .windows {
+        let installations = try await VSInstallation.findInstallations(fs: fs)
+            .sorted(by: { $0.installationVersion > $1.installationVersion })
+        if let latest = installations.first {
+            let msvcDir = latest.installationPath.join("VC").join("Tools").join("MSVC")
+            let versions = try fs.listdir(msvcDir).map { try Version($0) }.sorted { $0 > $1 }
```
jakepetroules, line 35: Based on recent lessons learned... we should not use the latest version. Instead, locate this file:

```swift
latest.installationPath.join("VC").join("Auxiliary").join("Build").join("Microsoft.VCToolsVersion.default.txt")
```

...and then read its contents (stripping whitespace) to determine the value of `latestVersion`.

This is what vswhere and clang will do, so we'll want to match. _Usually_ this is the same as the latest version, but sometimes it's not, and the "latest" code will fail catastrophically if that's the case. :)

You should also check the `VCToolsVersion` environment variable and use that in preference to `Microsoft.VCToolsVersion.default.txt` if it exists.

jakepetroules (changes requested):

Sources/SWBWindowsPlatform/Plugin.swift
jakepetroules: I also realize this existing code (that I wrote) that you're just moving unchanged, so if you'd rather I fix this in a separate PR I'm also happy to do that, especially so it won't invalidate your CI run. Let me know your preference.

jakepetroules: Hmm, or we could actually run vcvarsall which is probably better than trying to construct paths manually. Actually, let me just take a look at this separately.

Sources/SWBWindowsPlatform/Plugin.swift
```diff
@@ -25,22 +26,42 @@ struct WindowsPlatformSpecsExtension: SpecificationsExtension {
...
+        return [vcToolsInstallDir: dir.str]
+    }
+}
+
+struct WindowsPlatformExtension: PlatformInfoExtension {
+    public func additionalPlatformExecutableSearchPaths(platformName: String, platformPath: Path, fs: any FSProxy) async -> [Path] {
+        guard let dir = try? await findLatestInstallDirectory(fs: fs) else {
+            return []
+        }
+        if Architecture.hostStringValue == "aarch64" {
```
jakepetroules, line 61: This is not correct because the search paths need to be dependent on the target architecture. Unfortunately, Microsoft does not have a universal driver like clang, there's one per-arch. This also must be why the test is failing...

This is going to be hard to fix because these search paths are provided in an architecture independent context.

kcieplak: Good point, I will see if I can come up with something to solve this.

kcieplak:
@swift-ci test

Sources/SWBWindowsPlatform/Plugin.swift
kcieplak: This has been addressed with the latest update to the PR.

Sources/SWBCore/Settings/BuiltinMacros.swift
```diff
@@ -803,6 +803,8 @@ public final class BuiltinMacros {
     public static let LD_RUNPATH_SEARCH_PATHS = BuiltinMacros.declareStringListMacro("LD_RUNPATH_SEARCH_PATHS")
     public static let LD_SDK_IMPORTS_FILE = BuiltinMacros.declarePathMacro("LD_SDK_IMPORTS_FILE")
     public static let LD_WARN_UNUSED_DYLIBS = BuiltinMacros.declareBooleanMacro("LD_WARN_UNUSED_DYLIBS")
+    public static let LD_MULTIARCH = BuiltinMacros.declareBooleanMacro("LD_MULTIARCH")
```
kcieplak, line 806: @jakepetroules Do you think I should make these **_LD_MULTIARCH** and **_LD_MULTIARCH_PREFIX_MAP** as the values should not really be adjusted by external clients.  The values are set from platform xcspec, based on the ALTERNATE_LINKER value.

kcieplak:
@swift-ci test

kcieplak:
@swift-ci test macOS

Tests/SWBCoreTests/CommandLineToolSpecDiscoveredInfoTests.swift
```diff
@@ -141,31 +142,134 @@ import Testing
...
+        try await withSpec(LdLinkerSpec.self, .deferred, platform: "windows", additionalTable: table) { (info: DiscoveredLdLinkerToolSpecInfo) in
+            #expect(!info.toolPath.isEmpty)
+            #expect(info.toolVersion != nil)
+            if let toolVersion = info.toolVersion {
+                #expect(toolVersion > Version(0, 0, 0))
+            }
+            #expect(info.linker == .linkExe)
+        }
+
+        // // link aarch64
```
kcieplak, line 263: Fails on swift-ci, passes locally.

Sources/SWBCore/Settings/BuiltinMacros.swift
jakepetroules: Yeah that's probably a good idea.

In the longer term, we've had a longstanding desire to have actually-read-only build settings.

Sources/SWBCore/SpecImplementations/Tools/LinkerTools.swift
```diff
@@ -1309,15 +1308,48 @@ public final class LdLinkerSpec : GenericLinkerSpec, SpecIdentifierType, @unchec
...
             linkerPath = Path(producer.hostOperatingSystem.imageFormat.executableName(basename: "ld.\(alternateLinker)"))
+        } else if alternateLinker != "" {
+            linkerPath =  Path(alternateLinker)
+        }
+        // If the linker does not support multiple architectures update the path to include a subfolder based on the prefix map
+        // to find the architecture specific executable.
+        if !isLinkerMultiarch {
+            let archMap = scope.evaluate(BuiltinMacros.LD_MULTIARCH_PREFIX_MAP)
+            let archMappings = archMap.reduce(into: [String: String]()) { mappings, map in
+                let split = map.components(separatedBy: ":")
```
jakepetroules, line 1325: Use `String.split(_:)` as this is guaranteed to return two results. `split` here could contain only one element and therefore crash.

Sources/SWBCore/SpecImplementations/Tools/LinkerTools.swift
```diff
@@ -1309,15 +1308,48 @@ public final class LdLinkerSpec : GenericLinkerSpec, SpecIdentifierType, @unchec
...
+                }
+            }
+            if archMappings.isEmpty {
+                delegate.error("LD_MULTIARCH is 'false', but no prefix mappings are present in LD_MULTIARCH_PREFIX_MAP")
+                return nil
+            }
+            // Linkers that don't support multiple architectures cannot support universal binaries, so ARCHS will
+            // contain the target architecture and can only be a single value.
+            let arch = scope.evaluate(BuiltinMacros.ARCHS)
+            if arch.count > 1 {
```
jakepetroules, line 1337: nit: `guard let arch = scope.evaluate(BuiltinMacros.ARCHS).only else {`

Sources/SWBTestSupport/CoreBasedTests.swift
```diff
@@ -272,53 +272,55 @@ extension CoreBasedTests {
...
+                if let executable = platform.executableSearchPaths.findExecutable(operatingSystem: core.hostOperatingSystem, basename: "ld.gold") {
                     return executable
                 }
             }
             return nil
         }
     }
+    package func linkPath(_ targetArchitecture: String) async throws -> Path? {
+        let (core, defaultToolchain) = try await self.coreAndToolchain()
+        let prefixMapping = [ "x86_64": "x64", "aarch64": "arm64", "arm64": "arm64" ]
```
jakepetroules, line 305: Add:

i386 -> x86
thumbv7 -> arm

I don't think you need the arm64 -> arm64 mapping, Windows uses the aarch64 spelling.

Sources/SWBWindowsPlatform/Plugin.swift
jakepetroules: Thanks. Can you do this as a switch statement though?

```swift
switch Architecture.hostStringValue {
case "aarch64":
    return [dir.join("bin/Hostarm64")]
case "x86_64":
    return [dir.join("bin/Hostx64")]
default:
    return []
}
```

Prefer to be explicit, maybe Windows adds riscv64 someday or something.

Tests/SWBCoreTests/CommandLineToolSpecDiscoveredInfoTests.swift
jakepetroules: swift-ci might not have the ARM compilers installed.

@shahmishal can we get those added if they aren't present? All versions of Windows can cross compile to every architecture; we want to be able to build arm64 binaries even on the current x86_64 CI.

The component string is `Microsoft.VisualStudio.Component.VC.Tools.ARM64`

kcieplak:
@swift-ci test

kcieplak:
@swift-ci test

jakepetroules:
FYI you need to rebase this as it conflicts with main

Sources/SWBTestSupport/CoreBasedTests.swift
```diff
@@ -272,53 +272,55 @@ extension CoreBasedTests {
...
+                if let executable = platform.executableSearchPaths.findExecutable(operatingSystem: core.hostOperatingSystem, basename: "ld.gold") {
                     return executable
                 }
             }
             return nil
         }
     }
+    package func linkPath(_ targetArchitecture: String) async throws -> Path? {
+        let (core, defaultToolchain) = try await self.coreAndToolchain()
+        let prefixMapping = [ "x86_64": "x64", "aarch64": "arm64", "i386": "x86", "thumbv7" : "arm64" ]
```
jakepetroules, line 305: thumbv7 maps to "arm" not "arm64"

kcieplak: Fixed

kcieplak:
@swift-ci test

kcieplak:
> FYI you need to rebase this as it conflicts with main

Fixing up now, had conflicts with two PR merges.

kcieplak:
@swift-ci test

kcieplak:
@swift-ci  test

kcieplak:
@swift-ci  test

kcieplak:
@swift-ci test

Tests/SWBCoreTests/CommandLineToolSpecDiscoveredInfoTests.swift
```diff
@@ -268,6 +268,7 @@ import SWBMacro
             #expect(info.linker == .linkExe)
         }
 
+        // FIXME: Fails in swift-ci missing arm64 toolchain on x86_64 host.
```
jakepetroules, line 271: I'd move this test to SWBWindowsPlatformTests, and then you can use #290 to conditionalize it.

kcieplak: Done

kcieplak:
@swift-ci test

kcieplak:
> @swift-ci test

Crash on Linux and MacOS, seen in other PRs not isolated to this change

kcieplak:
@swift-ci test

Sources/SWBCore/SDKRegistry.swift
```diff
@@ -646,6 +646,14 @@ public final class SDKRegistry: SDKRegistryLookup, CustomStringConvertible, Send
             tripleEnvironment = ""
         }
 
+        let archs: PropertyListItem =
```
jakepetroules, line 649: Diffs in this file should be removed now, after the refactoring in #279

Sources/SWBTestSupport/CoreBasedTests.swift
```diff
@@ -272,53 +272,55 @@ extension CoreBasedTests {
...
+                if let executable = platform.executableSearchPaths.findExecutable(operatingSystem: core.hostOperatingSystem, basename: "ld.gold") {
                     return executable
                 }
             }
             return nil
         }
     }
+    package func linkPath(_ targetArchitecture: String) async throws -> Path? {
+        let (core, defaultToolchain) = try await self.coreAndToolchain()
+        let prefixMapping = [ "x86_64": "x64", "aarch64": "arm64", "i386": "x86", "thumbv7" : "arm" ]
```
jakepetroules, line 305: Sorry for the churn but I ended up changing thumbv7 to armv7 in #304. And i386 should be i686.

Sources/SWBWindowsPlatform/Specs/WindowsLd.xcspec
```diff
@@ -52,6 +52,18 @@
...
+            {
+                Name = _LD_MULTIARCH;
+                Type = Boolean;
+                DefaultValue = NO;
+                Condition = "$(ALTERNATE_LINKER) == link";
+            },
+            {
+                Name = _LD_MULTIARCH_PREFIX_MAP;
+                Type = StringList;
+                DefaultValue = "x86_64:x64 aarch64:arm64 i386:x86 thumbv7:arm";
```
jakepetroules, line 64: Sorry for the churn but I changed thumbv7 to armv7 in #304. And i386 should be i686.

Sources/SWBWindowsPlatform/Plugin.swift
```diff
@@ -18,19 +18,37 @@ import Foundation
...
     manager.register(WindowsPlatformSpecsExtension(), type: SpecificationsExtensionPoint.self)
     manager.register(WindowsEnvironmentExtension(plugin: plugin), type: EnvironmentExtensionPoint.self)
-    manager.register(WindowsPlatformExtension(), type: PlatformInfoExtensionPoint.self)
+    manager.register(WindowsPlatformExtension(plugin: plugin), type: PlatformInfoExtensionPoint.self)
     manager.register(WindowsSDKRegistryExtension(), type: SDKRegistryExtensionPoint.self)
 }
 
 public final class WindowsPlugin: Sendable {
     private let vsInstallations = AsyncSingleValueCache<[VSInstallation], any Error>()
+    private let latestVsInstalltionDirectory = AsyncSingleValueCache<Path?, any Error>()
```
jakepetroules, line 27: "Installtion" here is mis-spelled

Sources/SWBWindowsPlatform/Plugin.swift
```diff
@@ -44,26 +62,34 @@ struct WindowsPlatformSpecsExtension: SpecificationsExtension {
...
+            guard let dir = try? await findLatestInstallDirectory(fs: context.fs) else {
+                return [:]
+            }
+            return [vcToolsInstallDir: dir.str]
+        } else {
+           return [:]
+        }
+    }
+
+    @_spi(Testing)public func findLatestInstallDirectory(fs: any FSProxy) async throws -> Path? {
```
jakepetroules, line 75: This function seems to be duplicated with `cachedLatestVSInstallDirectory` above. Can we unify them?

Tests/SWBBuildSystemTests/LinkerTests.swift
```diff
@@ -303,28 +307,55 @@ fileprivate struct LinkerTests: CoreBasedTests {
...
+                parameters = BuildParameters(configuration: "Debug", overrides: ["ARCHS": "x86_64", "ALTERNATE_LINKER": "link"])
                 try await tester.checkBuild(parameters: parameters, runDestination: .host) { results in
                     results.checkTask(.matchRuleType("Ld")) { task in
                         task.checkCommandLineContains(["-fuse-ld=link"])
                         results.checkTaskOutput(task) { output in
-                            // Expect that the default linker is called by clang
-                            #expect(output.asString.contains(linkLinkerPath.str))
+                            // Expect that the 'link' linker is called by clang
+                            if runDestination == .windows && Architecture.hostStringValue == "aarch64" {
+                                // On windows aarch64 'clang' picks the wrong host architecture for link.exe, choosing "MSVC\14.41.34120\bin\Hostx86\x64\link.exe"
```
jakepetroules, line 327: Is this still an issue?

Tests/SWBBuildSystemTests/LinkerTests.swift
```diff
@@ -303,28 +307,55 @@ fileprivate struct LinkerTests: CoreBasedTests {
...
             }
+            if let linkLinkerPathAarch64 {
+                parameters = BuildParameters(configuration: "Debug", overrides: ["ARCHS": "aarch64", "ALTERNATE_LINKER": "link"])
+                try await tester.checkBuild(parameters: parameters, runDestination: .host) { results in
+                    results.checkTask(.matchRuleType("Ld")) { task in
+                        task.checkCommandLineContains(["-fuse-ld=link"])
+                         results.checkTaskOutput(task) { output in
+                            // Expect that the 'link' linker is called by clang
+                            if runDestination == .windows && Architecture.hostStringValue == "aarch64" {
+                                // On windows aarch64 'clang' picks the wrong host architecture for link.exe, choosing "MSVC\14.41.34120\bin\Hostx86\x64\link.exe"
```
jakepetroules, line 347: Is this still an issue?

Tests/SWBWindowsPlatformTests/SWBWindowsPlatformTests.swift
```diff
@@ -9,4 +9,69 @@
...
+        }
+
+        // link.exe cannot be used for multipler architectures and requires a distinct link.exe for each target architecture
+        table.push(BuiltinMacros.ALTERNATE_LINKER, literal: "link")
+        table.push(BuiltinMacros._LD_MULTIARCH, literal: false)
+        table.push(BuiltinMacros._LD_MULTIARCH_PREFIX_MAP, literal: ["x86_64:x64", "aarch64:arm64", "i386:x86", "thumbv7:arm"])
+
+        // x86_64
+        if try await core.hasVisualStudioComponent(.visualCppBuildTools_x86_x64) {
+            table.push(BuiltinMacros.ARCHS, literal: ["x86_64"])
```
jakepetroules, line 53: Can we add an i686 test in here? This will help ensure we're properly exercising the arch mapping.

(note that the visualCppBuildTools_x86_x64 component covers both 32-bit and 64-bit Intel)

Tests/SWBWindowsPlatformTests/SWBWindowsPlatformTests.swift
```diff
@@ -9,4 +9,69 @@
...
+            try await withSpec(LdLinkerSpec.self, .deferred, platform: "windows", additionalTable: table) { (info: DiscoveredLdLinkerToolSpecInfo) in
+                #expect(!info.toolPath.isEmpty)
+                #expect(info.toolVersion != nil)
+                if let toolVersion = info.toolVersion {
+                    #expect(toolVersion > Version(0, 0, 0))
+                }
+                #expect(info.linker == .linkExe)
+            }
+        }
+    }
```
jakepetroules, line 76: Can we add armv7 (visualCppBuildTools_arm) and arm64ec (visualCppBuildTools_arm64ec) blocks here as well? Then we'll have coverage of all 5 ABIs.

arm64ec will map to arm64 in the linker arch map.

jakepetroules (changes requested):
Looking good, just a few more changes needed!

Sources/SWBWindowsPlatform/Plugin.swift
kcieplak: Yes, I actually was calling plugin.cachedLatestVSInstallDirectory, but had merge conflicts when rebasing on main, and took the wrong sections.  Updating...

Tests/SWBBuildSystemTests/LinkerTests.swift
kcieplak: Yes adding issue number to the comment.

Tests/SWBBuildSystemTests/LinkerTests.swift
kcieplak: Yes adding issue number to the comment.

kcieplak:
@swift-ci test

PR #243, opened 2025-03-03
build: attempt to bundle files for the modules

compnerd:
This adds a helper to bundle files into the resource directory for the module.

### DIFF
--- a/Sources/CMakeLists.txt
+++ b/Sources/CMakeLists.txt
@@ -8,6 +8,21 @@ See http://swift.org/LICENSE.txt for license information
 See http://swift.org/CONTRIBUTORS.txt for Swift project authors
 ]]
 
+include(CMakeParseArguments)
+function(SwiftBuild_Bundle)
+  set(Options)
+  set(OneValueArguments MODULE)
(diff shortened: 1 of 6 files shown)


compnerd:
@swift-ci please test

owenv (approved):

jakepetroules:
What about SWBApplePlatform, SWBUniversalPlatform, and SWBCore?

compnerd:
I wanted to get the basic files packaged up first. There was the question of whether the `Specs` directory needed to be preserved and I figured that we should do those in a follow up.

owenv:
fwiw preserving the Specs/ directory structure is not important, it's just a way of organizing the sources

owenv:
@compnerd would you mind if I picked these changes over to another PR? I've been testing them out with some local additions to begin bundling the rest of the specs and simplify the resource accessor codegen a bit.

compnerd:
@owenv nope, by all means! Sorry, I've not been able to push this forward, been trying to finish up the static linking support to get the early swift-driver enabled on Windows.

PR #244, opened 2025-03-03
Add a workspace for local co-development of Swift Build and swift-driver

owenv:
This is useful for local development of the two projects together on a macOS host. In the future we could consider including some of the other relevant repos as well


owenv:
@swift-ci test

jakepetroules (approved):

owenv:
@swift-ci test

PR #253, opened 2025-03-05
Use SwiftPM's builtin resource rules to copy spec files

owenv:
This removes the use of the SWBSpecificationsCompiler executable, which will simplify the macOS toolchain build that currently relies on `--build-system xcode`


jakepetroules (approved):

owenv:
ok, make build is fixed up now

@swift-ci test

owenv:
@swift-ci test

Sources/SWBSpecificationsCompiler/Compiler.swift
```diff
@@ -1,72 +0,0 @@
...
-            try? FileHandle.standardError.write(contentsOf: Data("error: \(error)\n".utf8))
-            exit(EXIT_FAILURE)
-        }
-    }
-
-    /// Performs some very basic stripping of `#if` directives from xcspec files in case they contain them.
-    ///
-    /// In this build step we are not actually going to read the values and conditionally include the inner spec portions.
-    /// Instead, we will just remove the `#if` lines themselves, which can be needed in certain build configurations that may contain them.
-    private static func unifdef(_ data: inout Data) {
```
matthewseaman, line 54: I'm guessing SwiftPM's builtin build rule for this does not do this.
Is that concerning for any configurations?

owenv: This is a temporary measure to simplify bootstrapping, so I expect we'll be able to bring it back if needed

PR #254, opened 2025-03-05
Fixed bug with expandedSearchPaths when dealing with windows paths

daveinglis:
- enabled recursiveSearchPathValidation test for all hosts
- enabled some Path settings tests for all hosts
- default some testing support methods to be .host instead of .macOS


daveinglis:
@swift-ci test

Sources/SWBWindowsPlatform/Windows.xcspec
```diff
@@ -42,6 +42,13 @@
         BasedOn = com.apple.product-type.tool;
     },
 
+    {
```
jakepetroules, line 45: I don't think we should add this.

Application bundles are a distinct Apple platforms concept.

jakepetroules (changes requested):

Tests/SWBCoreTests/SettingsTests.swift
```diff
@@ -3553,7 +3553,7 @@ import SWBMacro
     }
 
     /// Check the behavior of user defined settings, especially w.r.t. quoting rules.
-    @Test(.requireSDKs(.macOS))
```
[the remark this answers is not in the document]
jakepetroules: It works because of https://github.com/swiftlang/swift-build/pull/254#discussion_r1982315531, but I don't think we should add that.

daveinglis:
@swift-ci test

Sources/SWBWindowsPlatform/Windows.xcspec
daveinglis: fixed

Tests/SWBCoreTests/SettingsTests.swift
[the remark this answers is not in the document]
daveinglis: fixed

daveinglis:
@swift-ci test windows

Tests/SWBCoreTests/SettingsTests.swift
```diff
@@ -1957,8 +1957,8 @@ import SWBMacro
         }
 
         let scope = settings.globalScope
-        #expect(scope.evaluate(BuiltinMacros.SYMROOT) == Path("/tmp/Workspace/aProject/build"))
-        #expect(scope.evaluate(BuiltinMacros.DSTROOT) == Path("/tmp/Workspace/aProject/build/foo/bar/dst"))
+        #expect(scope.evaluate(BuiltinMacros.SYMROOT) == Path.root.join("/tmp/Workspace/aProject/build"))
```
jakepetroules, line 1960: Need to remove the leading slash from the call to join. Maybe it works by accident right now, but this is not semantically correct (on Windows you want a path relative to the left hand side, not relative to the current drive root). Same thing in BuildDescriptionTests.swift

jakepetroules (changes requested):

daveinglis:
@swift-ci test

jakepetroules (approved):

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test

PR #261, opened 2025-03-06
[swift] Always emit pch output path in OutputFileMap

cachemeifyoucan:
Always emit pch output path in OutputfileMap to support possible chained bridging header support. In the chained bridging header support mode, swift-driver can plan a generate PCH job even when there are no explicit bridging header is passed on command-line, and it needs the path in output file map to write the PCH file into desired location. Since having additional unused entries in the output file map is not an issue, just always generate output path for PCH in the output file map.

rdar://146409233


cachemeifyoucan:
@swift-ci please test

neonichu (approved):

PR #264, opened 2025-03-07
Resolves #263 - Add all valid architectures for windows cross compilation

### LINKED ISSUE #263: Windows: Cross architecture compilation not possible
Cross compilation does not work on Windows when it should be possible.

When on an arm64 host system, and setting ARCHS = "x86_64" a warning will be emitted.

`→ found unexpected warning: None of the architectures in ARCHS (x86_64) are valid. Consider setting ARCHS to $(ARCHS_STANDARD) or updating it to include at least one value from VALID_ARCHS (aarch64). (in target 'CommandLineTool' from project 'TestProject')`

VALID_ARCHS is only "aarch64"

kcieplak:
* The fallback SDK registration was only adding the host architecture as a valid architecture.  Add all valid architectures for Windows cross compilation.

### DIFF
--- a/Sources/SWBCore/SDKRegistry.swift
+++ b/Sources/SWBCore/SDKRegistry.swift
@@ -664,7 +664,7 @@ public final class SDKRegistry: SDKRegistryLookup, CustomStringConvertible, Send
 
         let archs: PropertyListItem = switch operatingSystem {
             case .windows:
-                .plArray([.plString("x86_64"), .plString("aarch64")])
+                .plArray([.plString("x86_64"), .plString("i686"), .plString("aarch64"), .plString("thumbv7")])
             default:
                 .plArray([.plString(Architecture.hostStringValue ?? "unknown")])
         }
--- a/Sources/SWBCore/SDKRegistry.swift
+++ b/Sources/SWBCore/SDKRegistry.swift
@@ -662,6 +662,12 @@ public final class SDKRegistry: SDKRegistryLookup, CustomStringConvertible, Send
             tripleEnvironment = ""
         }
 
+        let archs: PropertyListItem = switch operatingSystem {
+            case .windows:
+                .plArray([.plString("x86_64"), .plString("aarch64")])
+            default:
+                .plArray([.plString(Architecture.hostStringValue ?? "unknown")])
+        }
         return try [
             "Type": .plString("SDK"),
             "Version": .plString(Version(ProcessInfo.processInfo.operatingSystemVersion).zeroTrimmed.description),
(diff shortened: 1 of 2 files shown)


kcieplak:
@swift-ci test

Sources/SWBCore/SDKRegistry.swift
```diff
@@ -662,6 +662,12 @@ public final class SDKRegistry: SDKRegistryLookup, CustomStringConvertible, Send
             tripleEnvironment = ""
         }
 
+        let archs: PropertyListItem = switch operatingSystem {
+            case .windows:
+                .plArray([.plString("x86_64"), .plString("aarch64")])
```
jakepetroules, line 667: `i686` is also supported by the Swift toolchain; we should add that too.

I'd also add `armv7` (32-bit) (CC @compnerd for the specific spelling) since it's technically still supported by the Microsoft toolchains and you could build a C package for Windows IoT if you really wanted to. The Swift toolchain doesn't currently include it but I just checked with @compnerd and he said he'd be open to adding it.

jakepetroules (changes requested):

jakepetroules:
I think you'll also want to change Windows.xcspec to default `ARCHS_STANDARD` to `$(HOST_ARCH)` instead of `$(VALID_ARCHS)` since building for multiple archs is not really going to work correctly on non-Apple platforms at this time.

Sources/SWBCore/SDKRegistry.swift
kcieplak: clang also supports a few others...

`Registered Targets:
    aarch64    - AArch64 (little endian)
    aarch64_32 - AArch64 (little endian ILP32)
    aarch64_be - AArch64 (big endian)
    arm        - ARM
    arm64      - ARM64 (little endian)
    arm64_32   - ARM64 (little endian ILP32)
    armeb      - ARM (big endian)
    thumb      - Thumb
    thumbeb    - Thumb (big endian)
    wasm32     - WebAssembly 32-bit
    wasm64     - WebAssembly 64-bit
    x86        - 32-bit X86: Pentium-Pro and above
    x86-64     - 64-bit X86: EM64T and AMD64`

kcieplak:
> I think you'll also want to change Windows.xcspec to default `ARCHS_STANDARD` to `$(HOST_ARCH)` instead of `$(VALID_ARCHS)` since building for multiple archs is not really going to work correctly on non-Apple platforms at this time.

I did notice that if you set ARCHS = "x86_64 aarch64" only one was ever built for.  That explains that.

kcieplak:
@swift-ci test

PR #265, opened 2025-03-07
Don't emit a note if previews is disabled in the build system <rdar://145843881>

jonathanpenn:
This is already detected at higher levels. The machinery to emit the note is still here in case we need to come back to it or add to it, though.


neonichu (approved):

jakepetroules (approved):

jonathanpenn:
@swift-ci test

jonathanpenn:
@swift-ci test macos

PR #266, opened 2025-03-07
Add support for looking up spec resources in a toolchain install

owenv:
This will allow us to load resource bundles when installed as part of SwiftPM in a Swift toolchain

### DIFF
--- a/Sources/SWBCore/Extensions/SpecificationsExtension.swift
+++ b/Sources/SWBCore/Extensions/SpecificationsExtension.swift
@@ -50,7 +50,7 @@ public struct SpecificationsExtensionPoint: ExtensionPoint {
 
 public protocol SpecificationsExtension: Sendable {
     /// Returns the bundle containing the `.xcspec` files.
-    func specificationFiles() -> Bundle?
+    func specificationFiles(resourceSearchPaths: [Path]) -> Bundle?
     func specificationDomains() -> [String: [String]]
     func specificationTypes() -> [any SpecType.Type]
     func specificationClasses() -> [any SpecIdentifierType.Type]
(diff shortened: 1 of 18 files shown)


owenv:
@swift-ci test

neonichu (approved):

owenv:
@swift-ci test

Sources/SWBCore/Extensions/SpecificationsExtension.swift
```diff
@@ -69,4 +69,22 @@ extension SpecificationsExtension {
...
+
+    public func findResourceBundle(nameWhenInstalledInToolchain: String, defaultBundle: Bundle?) -> Bundle? {
+        #if SEARCH_TOOLCHAIN_FOR_RESOURCES
+        let resourceBundleExtension: String
+        #if canImport(Darwin)
+        resourceBundleExtension = "bundle"
+        #else
+        resourceBundleExtension = "resources"
+        #endif
+        if let executablePath = try? Path(CommandLine.arguments[0]).resolveSymlink(fs: localFS) {
```
jakepetroules, line 81: CommandLine.arguments in libraries is not great, can we instead plumb this down or base it on `core.developerDir`?

jakepetroules (changes requested):

Sources/SWBCore/Extensions/SpecificationsExtension.swift
owenv: I don't know of any good alternatives right now, core.developerDir really isn't consistent enough across platforms

jakepetroules: Actually, this should be toolchain relative. That should already be consistent across platforms, including on Windows and in Xcode. Can you pass in the toolchain registry to specificationFiles() and look it up relative to the default toolchain?

Would require swapping the order of initializeSpecRegistry() and initializeToolchainRegistry(); hopefully that's OK...

owenv: That doesn't work on linux where the fallback system toolchain exists

jakepetroules: Why not? On Linux the path of the fallback system toolchain is going to be the path 3 directories up from where the `swift` executable lives, so I think that isn't any different from other platforms. You should be able to construct `toolchain.path.join("share")` to get to the resources location...

owenv: I ended up exposing API for search paths here instead, so SwiftPM can point us to its own resource dir

owenv:
@swift-ci test

owenv:
@swift-ci test

jakepetroules (approved):
Thanks, this looks good!

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

PR #267, opened 2025-03-08
Disable Swift explicit modules for language modes < 5

owenv:
This avoids a situation which could result in mixing incompatible versioned API notes until this is handled correctly by the compiler

rdar://146404229


owenv:
@swift-ci test

neonichu (approved):

owenv:
@swift-ci test macOS

PR #276, opened 2025-03-09
Remove TestStandardTarget's default product type argument

jakepetroules:
This makes it more obvious that a test target is an app at first glance, and makes our test infrastructure slightly less Apple-centric by ceasing to encode an implicit assumption of a default app product type.


jakepetroules:
@swift-ci test

neonichu (approved):

PR #280, opened 2025-03-10
Filter out linker warnings about an empty SubFrameworks directory in Catalyst tests.

mhrawdon:
Warnings about this are an artifact of how Catalyst support is implemented rather than providing actionable information.

rdar://146492614


mhrawdon:
@swift-ci test

jakepetroules (approved):

PR #281, opened 2025-03-11
Explicitly annotate sendability of public 'Purpose' enum

owenv:
Fix a warning in contexts where we require explicit annotation of public types

rdar://144845295


owenv:
@swift-ci test

jakepetroules (approved):

owenv:
@swift-ci test linux

PR #283, opened 2025-03-11
Don't assert equality of modification timestamps after file copies

jakepetroules:
FSProxy's copy() function (which uses FileManager in the LocalFS implementation) does not appear to be able to guarantee that timestamps will be preserved for copied files or directories, on any filesystem or platform. Small variances are often observed here when the copy (for whatever reason) takes longer than the minimum timestamp resolution of the underlying filesystem with respect to the creation time of the original file. This leads to flaky tests on Windows and Linux. This is generally not observed on macOS presumably due to the use of copyfile() in Foundation's implementation which uses APFS cloning and therefore doesn't go down the fallback path which creates new directory and filesystem nodes at the destination.

So, drop the assertions as we can't actually make this guarantee, and it's not clear we really want to make this guarantee with the desire to shift the build system away from reliance on timestamps as much as possible in the fullness of time.

rdar://146697830


jakepetroules:
@swift-ci test

PR #289, opened 2025-03-12
Add build warning when usage description key has empty string for value.

baujla:
A common issue developers face when uploading to App Store Connect is the usage string they select has an empty value. Adding a build warning for this case will help developers resolve this issue in advance.


jakepetroules:
@swift-ci test

jakepetroules (approved):

neonichu (approved):

PR #291, opened 2025-03-12
Remove variant name from explicit precompile rule info to allow deduping

owenv:
Not all variants depend on incompatible modules. Remove the variant name from the rule info so that identical tasks are deduplicated properly and cannot race.


owenv:
@swift-ci test

jakepetroules (approved):

neonichu (approved):

cachemeifyoucan (commented):
Change LGTM.

The test seems a bit fragile since where that will be one or two task scheduled is at the mercy of the compiler implementation to produce the same module task or not. Fine with me for now but I am not sure if we decide to stick with the current module flags for swift asan.

owenv:
@cachemeifyoucan in this test instead of using ASAN I'm building a variant with identical compiler args, which should ensure that even if the compiler changes the test is still sound

cachemeifyoucan (approved):
I see. It is testing `other` variants. LGTM.

aciidgh (approved):

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test Linux

owenv:
@swift-ci test macOS

PR #293, opened 2025-03-12
Ensure all overloads of SWBBuildService.createSession pass along the …

### LINKED ISSUE #292: SWBBuildService's deprecated createSession overloads do not pass along developerPath properly
After 2e6eb49, SWBBuildService's deprecated createSession overloads do not pass along developerPath properly: the one which is deprecated does pass it along, but then it gets dropped by the one marked `ABI compatibility`. This results in clients which use the deprecated overload(s) failing to find the correct Developer directory.

cwakamo:
…developerPath parameter.

Without this, clients which use a deprecated createSession overload fail to find the Developer directory during initialization.

This fixes #292.


cwakamo:
@swift-ci please test

mirza-garibovic (approved):

PR #297, opened 2025-03-12
Fix accidentally quadratic lookup operation in GlobalExplicitDependencyTracker

owenv:
Previously, filter was being inlined here, but now that it isn't this implementation is prohibitively slow

rdar://146634377


owenv:
@swift-ci test

jakepetroules (approved):

owenv:
@swift-ci test

aciidgh (approved):

PR #299, opened 2025-03-13
Start adopting the new Environment type throughout the codebase

### LINKED ISSUE #296: Error building on Windows with git-bash
Using the Windows Terminal with git-bash as my shell, I get the following error when running `swift build` against the library package template, though that doesn't seem related.

```
$ ../swiftpm/.build/debug/swift-build --build-system swiftbuild --build-tests

 0% [------------------------------------------------------------------------------------------------------]
error: unknown Can't determine path to cmd.exe because the ComSpec environment variable is not set []
```

Interesting enough COMSPEC is defined (all upper case).

jakepetroules:
Mostly starting at the "edges", in tests and in some central APIs. Does nothing for Unix-like platforms, but will fix some case sensitivity issues on Windows when accessing specific environment variables.

Closes #296


jakepetroules:
@swift-ci test

neonichu (approved):

PR #301, opened 2025-03-13
[SwiftCaching] Improve swift caching batch build performance

cachemeifyoucan:
Improve swift caching build performance, especially batch build performance by:
* Use more accurate dependency tracking by requesting cache key querying jobs in parallel when requesting all the explicit module dependencies. This allows build system to plan cache key querying jobs in the gaps of explicit module build, thus the compilation job can start sooner.
* Instead of issuing multiple key queries per batched job, issue one key querying job for multiple keys instead. This can dramatically reduce the number of jobs planned thus improves the efficiency. Also in the cache miss cases, if the first key hits the cache miss, there is no need to continue querying other keys since the job needs to be run anyway.

rdar://146429825

### DIFF
--- a/Sources/SWBTaskExecution/TaskActions/SwiftCachingMaterializeKeyTaskAction.swift
+++ b/Sources/SWBTaskExecution/TaskActions/SwiftCachingMaterializeKeyTaskAction.swift
@@ -41,7 +41,7 @@ public final class SwiftCachingMaterializeKeyTaskAction: TaskAction {
         /// The action is in its initial state, and has not yet performed any work.
         case initial
         /// Waiting for the caching key query to finish.
-        case waitingForKeyQuery(jobTaskIDBase: UInt, cas: SwiftCASDatabases, key: String)
+        case waitingForKeyQuery(jobTaskIDBase: UInt, cas: SwiftCASDatabases, keys: [String])
         /// Waiting for the outputs to finish downloading.
         case waitingForOutputDownloads
         /// Not waiting for any other dependency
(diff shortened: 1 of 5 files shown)


cachemeifyoucan:
@swift-ci please test

neonichu (approved):

jakepetroules (approved):

Sources/SWBTaskExecution/TaskActions/SwiftCachingMaterializeKeyTaskAction.swift
```diff
@@ -69,22 +69,16 @@ public final class SwiftCachingMaterializeKeyTaskAction: TaskAction {
...
-                    state = .done
-                }
-            } else {
-                let key = SwiftCachingKeyQueryTaskKey(casOptions: taskKey.casOptions, cacheKey: taskKey.cacheKey, compilerLocation: taskKey.compilerLocation)
+            let compOutputs = try taskKey.cacheKeys.compactMap {
+                try cas.queryLocalCacheKey($0)
+            }
+
+            if compOutputs.count != taskKey.cacheKeys.count {
+                let key = SwiftCachingKeyQueryTaskKey(casOptions: taskKey.casOptions, cacheKeys: taskKey.cacheKeys, compilerLocation: taskKey.compilerLocation)
```
akyrtzi, line 77: Should we only do a remote key query for the keys that were not found locally above?

cachemeifyoucan: Good catch. I can keep track of the missing keys and only request querying those.

cachemeifyoucan:
@swift-ci please test

akyrtzi (approved):

PR #304, opened 2025-03-14
Update the list of supported Windows archs

jakepetroules:
Change thumbv7 to armv7 -- even though it gets canonicalized as such internally, armv7 is the spelling given throughout the LLVM codebase. Also add the triple for the ARM64EC ABI (https://learn.microsoft.com/en-us/windows/arm/arm64ec)


jakepetroules:
@swift-ci test

neonichu (approved):

PR #315, opened 2025-03-19
Fix unsafeToInterrupt on Amazon Linux 2

jakepetroules:
This distro is going EoL in July, but fix the test anyways for better future-proofing.

What happens is that the task fails to execute because there is no working directory support on Amazon Linux 2, so the build fails. However, in that case the background task which waits for the sentinel file to be written will run forever. Ensure that we cancel the background task in cases where the build has completed.


neonichu (approved):

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test macos

PR #320, opened 2025-03-20
Make default fallback toolchain lookup more robust

jakepetroules:
Make the logic more resilient so that it works with Swift toolchains installed by Swiftly, which use symlinks.


jakepetroules:
@swift-ci test

neonichu (approved):

jakepetroules:
@swift-ci test

jakepetroules:
Ended up rolling this into #349

PR #326, opened 2025-03-20
Ensure headers contributing to system modules outside the sysroot invalidate swift planning

owenv:
Depends on https://github.com/swiftlang/swift-driver/pull/1848

Swift applies the equivalent of Clang's -MMD to headers included in its make style dependencies. This filters system headers even if they're outside the sysroot, which is unsound. Like we do with clang, register dependencies using the dependency scanner instead, and filter based on the SDK path. This is also a step towards fully eliminating use of make-style dependencies for Swift, which unblocks some perf optimizations.


jakepetroules (approved):

neonichu (approved):

owenv:
@swift-ci test

PR #338, opened 2025-03-22
Expand Linux checks to include Android or remove Linux checks

jakepetroules:
Anywhere we were checking for a Linux host, include Android as well if appropriate. Also adjust a couple other places to be non-Linux-specific where appropriate

### DIFF
--- a/Tests/SWBLLBuildTests/LLBuildTests.swift
+++ b/Tests/SWBLLBuildTests/LLBuildTests.swift
@@ -589,10 +589,6 @@ fileprivate class LoggingDelegate: BuildSystemDelegate {
     }
 
     @Test(.requireLLBuild(apiVersion: 15)) func commandSkipping() throws {
-        #if os(Linux) && !DEBUG
-        return // fails on Linux in release mode
-        #endif
-
         let fs = PseudoFS()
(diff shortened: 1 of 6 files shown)


jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test linux

jakepetroules:
@swift-ci test macos

jakepetroules:
@swift-ci test windows

Tests/SWBLLBuildTests/LLBuildTests.swift
```diff
@@ -616,7 +612,8 @@ fileprivate class LoggingDelegate: BuildSystemDelegate {
         // Configure the delegate
         class Delegate: LoggingDelegate {
             override func shouldCommandStart(_ command: Command) -> Bool {
-                assert(super.shouldCommandStart(command))
```
jakepetroules, line 619: The entire `assert`, including the call to the superclass implementation, would get optimized out, so the test would always fail in release mode - moved it to a separate statement to prevent this, and switched to `#expect` as well.

owenv (approved):

PR #339, opened 2025-03-22
Remove confusing special case in deployment target setting name lookup

jakepetroules:
Previously, the CommandProducer protocol provided a default implementation of `lookupPlatformInfo(platform:)` which simply returned the current SDK variant without considering the platform that was actually passed in.

This led to some cases around zippering (where an alternate platform's deployment target is requested) to give the wrong answer because the command producer was still operating in the context of the primary platform, leading to further workarounds in other places which special cased macOS and Mac Catalyst to return the right answer again.

To solve this, remove CommandProducer's default implementation and instead provide implementations in the concrete TaskProducerContext and MockCommandProducer (for testing) types which have access to a Core, and can therefore simply call through to the real canonical implementation.


jakepetroules:
@swift-ci test

owenv (approved):

PR #340, opened 2025-03-22
Remove the last usage of DispatchSemaphore in favor of native concurrency primitives

jakepetroules:
This avoids spinning up one thread to synchronize access to cached build systems.


jakepetroules:
@swift-ci test

neonichu (approved):

PR #343, opened 2025-03-24
Add assertion failure on out of bounds access to explicit dependency tracker

owenv:
A recent change turned a missing key here into an out of bounds access into an array. Downgrade it to an assertionFailure which is ignored in release builds in order to be more resilient to this failure and provide more information when investigating in debug builds.

rdar://147361460


owenv:
@swift-ci test

Issue #17, opened 2025-01-30
Move App Intents task producer and spec subclasses to SWBApplePlatform


owenv:
Currently, these live in SWBCore, we should move them to the SWBApplePlatform plugin alongside other support specific to Apple platforms. This may require adding a new extension point to inject the TaskProducer at the right point in product planning.

swiftlysingh:
Hey, is this issue being worked on, if not I would love to pick this up. 

Based on initial code go through, my plan is to:
1. Move AppIntentsMetadataCompilerSpec.swift from SWBCore to SWBApplePlatform
2. Move AppIntentsSSUTrainingCompiler.swift from SWBCore to SWBApplePlatform
3. Move AppIntentsMetadataTaskProducer.swift to SWBApplePlatform
4. Register these in the Apple platform's Plugin.swift

Let me know if this works, or you have some specific plans for it's implementation? @owenv

swiftlysingh:
Hi @owenv,  
Is this an issue that's already being worked on? If not, I'd be happy to take it on. Here's my initial plan based on a cursory review of the code:

- Move `AppIntentsMetadataCompilerSpec.swift` from SWBCore to SWBApplePlatform.
- Move `AppIntentsSSUTrainingCompiler.swift` from SWBCore to SWBApplePlatform.
- Move `AppIntentsMetadataTaskProducer.swift` to SWBApplePlatform.
- Register these in the Apple platform’s `Plugin.swift`.

Does this approach align with your expectations, or were you considering a different implementation strategy?  
Thanks!

owenv:
@swiftlysingh No, as far as I know nobody else is looking at this at the moment, so feel free to take a look! I think the plan you have sketched out sounds reasonable.

AppIntentsMetadataTaskProducer is probably the type which will take the most effort to move. I believe SWBApplePlatform could provide it via the recently added unorderedPostSetupTaskProducers extension point, though, similar to how StubBinaryTaskProducer is handled.

swiftlysingh:
Thanks, @owenv! I'll take a look and keep the `unorderedPostSetupTaskProducers` extension point in mind when moving `AppIntentsMetadataTaskProducer`.

PR #347, opened 2025-03-25
Move App Intents Task Producer to SWBApplePlatform

swiftlysingh:
This PR relocates the App Intents task producer and associated spec subclasses to the SWBApplePlatform directory, following the discussion in issue #17. 

**Important**: I have not yet updated the tests to account for this move. The build is successful. My focus is on getting feedback on the code changes before addressing test updates. I'm open to any input or suggestions on the refactor itself

### DIFF
--- a/Sources/SWBApplePlatform/AppIntentsMetadataTaskProducer.swift
+++ b/Sources/SWBApplePlatform/AppIntentsMetadataTaskProducer.swift
@@ -51,15 +51,14 @@ final class AppIntentsMetadataTaskProducer: PhasedTaskProducer, TaskProducer {
         guard !context.settings.globalScope.evaluate(BuiltinMacros.LM_SKIP_METADATA_EXTRACTION) else {
             return []
         }
+        guard let configuredTarget = self.targetContext.configuredTarget, let buildPhaseTarget = configuredTarget.target as? BuildPhaseTarget else {
+            return []
+        }
 
         context.addDeferredProducer {
             let scope = self.context.settings.globalScope
             var deferredTasks: [any PlannedTask] = []
             let buildFilesProcessingContext = BuildFilesProcessingContext(self.context.settings.globalScope)
-
-            guard let configuredTarget = self.targetContext.configuredTarget, let buildPhaseTarget = configuredTarget.target as? BuildPhaseTarget else {
-                fatalError("Target is not a BuildPhaseTarget")
-            }
             let swiftSources: [FileToBuild] = self.filterBuildFiles(buildPhaseTarget.sourcesBuildPhase?.buildFiles, identifiers: ["sourcecode.swift"], buildFilesProcessingContext: buildFilesProcessingContext)
             let perArchConstMetadataFiles = self.context.generatedSwiftConstMetadataFiles()
 
--- a/Sources/SWBApplePlatform/AppIntentsMetadataTaskProducer.swift
+++ b/Sources/SWBApplePlatform/AppIntentsMetadataTaskProducer.swift
@@ -12,14 +12,11 @@
 
 import SWBCore
 import SWBUtil
+import SWBTaskConstruction
 
 final class AppIntentsMetadataTaskProducer: PhasedTaskProducer, TaskProducer {
-    var sourcesBuildPhase: SourcesBuildPhase?
-    var resourcesBuildPhase: ResourcesBuildPhase?
 
-    init(_ context: TargetTaskProducerContext, sourcesBuildPhase: SourcesBuildPhase?, resourcesBuildPhase: ResourcesBuildPhase?, phaseStartNodes: [any PlannedNode], phaseEndNode: any PlannedNode) {
-        self.sourcesBuildPhase = sourcesBuildPhase
-        self.resourcesBuildPhase = resourcesBuildPhase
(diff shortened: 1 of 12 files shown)


Sources/SWBApplePlatform/AppIntentsMetadataTaskProducer.swift
```diff
@@ -58,8 +55,12 @@ final class AppIntentsMetadataTaskProducer: PhasedTaskProducer, TaskProducer {
         context.addDeferredProducer {
             let scope = self.context.settings.globalScope
             var deferredTasks: [any PlannedTask] = []
-            let buildFilesProcessingContext = BuildFilesProcessingContext(scope)
-            let swiftSources: [FileToBuild] = self.filterBuildFiles(self.sourcesBuildPhase?.buildFiles, identifiers: ["sourcecode.swift"], buildFilesProcessingContext: buildFilesProcessingContext)
+            let buildFilesProcessingContext = BuildFilesProcessingContext(self.context.settings.globalScope)
+
+            guard let configuredTarget = self.targetContext.configuredTarget, let buildPhaseTarget = configuredTarget.target as? BuildPhaseTarget else {
```
jakepetroules, line 60: nit: suggest you could just move this above the addDeferredProducer block and return an empty array instead of fatal-erroring

Sources/SWBApplePlatform/Plugin.swift
```diff
@@ -40,7 +40,8 @@ struct TaskProducersExtension: TaskProducerExtension {
     }
 
     var unorderedPostSetupTaskProducers: [any TaskProducerFactory] {
-        [StubBinaryTaskProducerFactory()]
+        [StubBinaryTaskProducerFactory(),
+         AppIntentsMetadataTaskProducerFactory()]
```
jakepetroules, line 44: nit: formatting would be more readable as:

```swift
[
    StubBinaryTaskProducerFactory(),
    AppIntentsMetadataTaskProducerFactory()
]
```

especially as things are added over time.

jakepetroules:
Looks like a great start, thanks!

swiftlysingh:
It appears that the following tests are failing after this refactor:

*   `AppIntentsMetadataTaskConstructionTests`
    *   `appIntentsMetadataWithoutYamlGeneration()`
    *   `appIntentsMetadataWithYamlGeneration()`
    *   `appIntentsMetadataWithAppNameOverrideAndDisable()`
    *   `multipleBuildVariants()`
*   `InstallLocTaskConstructionTests`
    *   `installLocSSUAppIntents()`

I'm not entirely sure how to best address these failures given the move to `SWBApplePlatform`. Any guidance or suggestions on how to update these tests would be greatly appreciated!

Sources/SWBTaskConstruction/ProductPlanning/ProductPlanner.swift
```diff
@@ -472,10 +472,6 @@ extension StandardTarget
             }
         }
 
-        createNewPhaseNode("AppIntentsMetadataTaskProducer")
-        taskProducers.append(AppIntentsMetadataTaskProducer(taskProducerContext, sourcesBuildPhase: self.sourcesBuildPhase, resourcesBuildPhase: self.resourcesBuildPhase, phaseStartNodes: [setupEndPhaseNode, buildPhasesEndNode], phaseEndNode: endPhaseNode))
```
owenv, line 476: I think most of the task construction failures you're seeing are because once this segment is replaced by a call into an extension as an unordered post-setup task producer, it loses the buildPhasesEndNode input which is added here. I think we could add a new method like `postBuildPhaseTaskProducers` to replicate the behavior before the task producer was moved

swiftlysingh:
Now we are passing in buildPhasesEndNode when we are constructing unorderedPostSetupTaskProducers. With this change, all task construction tests are passing.

Sources/SWBTaskConstruction/ProductPlanning/ProductPlanner.swift
```diff
@@ -467,15 +467,11 @@ extension StandardTarget
         for taskProducerExtension in taskProducerExtensions {
             for factory in taskProducerExtension.unorderedPostSetupTaskProducers {
                 createNewPhaseNode(factory.name)
-                taskProducers.append(factory.createTaskProducer(taskProducerContext, startPhaseNodes: [setupEndPhaseNode], endPhaseNode: endPhaseNode))
+                taskProducers.append(factory.createTaskProducer(taskProducerContext, startPhaseNodes: [setupEndPhaseNode, buildPhasesEndNode], endPhaseNode: endPhaseNode))
```
owenv, line 470: Some post-setup task producers (for example, StubBinaryTaskProducerFactory) will be injected via extension points here but shouldn't be ordered relative to buildPhasesEndNode. I think we might need to introduce an unorderedPostBuildPhasesTaskProducers counterpart to taskProducerExtension.unorderedPostSetupTaskProducers so we can differentiate between the two use cases

owenv (changes requested):
Looks great, thanks! - I just have one remaining comment before this is ready to land

swiftlysingh:
I have added `unorderedPostSetupTaskProducers` in `TaskProducerExtension` and add `AppIntentsMetadataTaskProducerFactory`

owenv:
@swift-ci test

owenv (approved):
lgtm, thanks!

neonichu (approved):

PR #348, opened 2025-03-25
Exclude additional directories from VALIDATE_DEPENDENCIES analysis

owenv:
Explicit modules output paths are produced by dynamic tasks, so we shouldn't compare them against the rootPathsPerTarget generated by analysis of the static build graph.

rdar://147840711


owenv:
@swift-ci test

jakepetroules (approved):

neonichu (approved):

PR #359, opened 2025-03-27
Add support for the libswiftCompatibilitySpan.dylib backward deployment library

slavapestov:


slavapestov:
This is a work-in-progress; I still want to add tests. Also the whole thing feels like a bunch of copy and paste. The problem is that "Swift in the OS" and "Swift concurrency" are two concepts plumbed through all these layers, and now I'm adding a third. Suggestions for improving this are welcome.

Sources/SWBCore/PlatformRegistry.swift
```diff
@@ -73,9 +73,12 @@ public final class Platform: Sendable {
     /// Minimum OS version for Swift-in-the-OS support. If this is `nil`, the platform does not support Swift-in-the-OS at all.
     fileprivate(set) var minimumOSForSwiftInTheOS: Version? = nil
 
-    /// Minimum OS version for built-in Swift concurrency support. If this is `nil`, the platform does not support Swift concurrency at all.
+    /// Minimum OS version for Swift concurrency (Swift 5.5). If this is `nil`, the platform does not support Swift concurrency at all.
     fileprivate(set) var minimumOSForSwiftConcurrency: Version? = nil
 
+    /// Minimum OS version for Span in the standard library (Swift 6.2). If this is `nil`, the platform does not support Swift concurrency at all.
```
jakepetroules, line 79: "the platform does not support Swift concurrency at all."

Think this needs to be updated to reflect Span, rather than Swift Concurrency

Sources/SWBCore/PlatformRegistry.swift
```diff
@@ -287,6 +295,29 @@ extension Platform {
 
         return version >= minimumSwiftConcurrencyVersion
     }
+
+    /// Determines if the platform natively supports Swift 6.2's Span type. If `false`, then the Swift Span back-compat concurrency libs needs to be copied into the app/framework's bundle.
+    public func supportsSwiftSpanNatively(_ scope: MacroEvaluationScope, forceNextMajorVersion: Bool = false, considerTargetDeviceOSVersion: Bool = true) -> Bool? {
```
jakepetroules, line 300: Why add default values for the arguments? This is only ever called in one place.

allevato:
I found my way here after noticing that `Span` was made to be back-deployable in Swift 6.2. I created a vanilla Xcode project (Xcode 26.0 beta 1) and made a macOS app bundle with an older deployment target and used a couple `Span` APIs, and I noticed that the compatibility library from the toolchain's `usr/lib/swift-6.2` folder wasn't being copied into my bundle. Does this need to land in order for back-deployment to work properly?

jakepetroules:
Closing in favor of #642

PR #361, opened 2025-03-28
Adopt libclang working directory optimization API

owenv:
Adopt new libclang API which the scanner uses to notify the build system when it is ok to compile a module using an arbitrary current working directory to facilitate increased module sharing. The relevant API was introduced in https://github.com/swiftlang/llvm-project/pull/10146.

rdar://145991369


owenv:
@swift-ci test

neonichu (approved):

jakepetroules (approved):

owenv:
@swift-ci test windows

Issue #23, opened 2025-01-31
Transition Path to FilePath on macOS and Linux


jakepetroules:
Currently Path has a useLegacyImplementation toggle which uses SwiftSystem.FilePath on Windows and our custom implementation elsewhere.

We should transition to FilePath on all platforms.

PR #363, opened 2025-03-29
Transition Path to FilePath

swiftlysingh:
This pull request refactors the Path struct within `Sources/SWBUtil/Path.swift` to exclusively use the FilePath type for internal path representation and manipulation. 

The legacy implementation toogle has been removed as part of this change. Feedback is appreciated. 

Relates to issue #23.

### DIFF
--- a/Sources/SWBUtil/Path.swift
+++ b/Sources/SWBUtil/Path.swift
@@ -36,34 +36,10 @@ private extension String.UTF8View {
 ///
 /// This struct adds support for common path manipulation operations.
 public struct Path: Serializable, Sendable {
-    private var useLegacyImplementation: Bool {
-        #if os(Windows)
-        return false
-        #else
-        return true
-        #endif
-    }
+    private var _impl: FilePath
 
-    private var _impl: FilePath {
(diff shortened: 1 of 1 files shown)


owenv:
> The legacy implementation toggle has been removed as part of this change. Note that some tests are currently failing due to behavioural differences after the FilePath migration. Should the FilePath implementation be adjusted to align with the expected test behaviours?

I think we'll need to look at the test failures on a case by case basis to make sure we're not unintentionally breaking compatibility with existing projects, but overall the implementation changes here lgtm. I'll kickoff CI so we can get an idea of where we might need to take a closer look for possible behavior changes

owenv:
@swift-ci please test

owenv:
based on the test output here it looks like there might be a lingering issue with path normalization:

```
✘ Test operators() recorded an issue at MacroEvaluationTests.swift:339:9: Expectation failed: (scope.evaluate(namespace.parseString("$(NON_STANDARD_REL_PATH:standardizepath)")) → "bar/baz") == "foo/../bar/baz"
```

swiftlysingh:
@swift-ci please test

PR #364, opened 2025-03-29
Emit warnings when conflicts detected in Info.plist key values

### LINKED ISSUE #5: Emit warnings when Info.plist key values from different user-supplied sources conflict
Currently when a build setting corresponding to an Info.plist key is set, and the same Info.plist key is set in the INFOPLIST_FILE, the build system will silently overwrite one with the other.

Since the behavior of which wins might be surprising or non-obvious to users, we should emit a warning when the two sides are in conflict, so users can ensure the intended effect of the merging process.

For scalar values, this is straightforward, and for arrays and dictionaries, we might choose an append/merge behavior depending on the key, and also emit warnings for conflicting keys hierarchically throughout nested dictionaries.

theSalted:
When a building setting is set from an info.plist key, if the same key is set from another source. The previous one will be silently overwritten. This pull request added `detectConflict()` to detects if two PropertyListItem values are in conflict and emits an appropriate warning message if they are.

Resolves #5


theSalted:
@swift-ci test

owenv (commented):
This looks great, thanks. Before it lands, do you mind adding a test to InfoPlistProcessorTaskTests.swift which validates the new warnings? `infoPlistProcessorTask()` and `diagnostics()` are examples of similar tests in that file.

owenv:
@swift-ci test

owenv:
@theSalted I think the bot used to update your fork overwrote the changes in this PR and It was auto-closed since it now matches `main`

theSalted:
> @theSalted I think the bot used to update your fork overwrote the changes in this PR and It was auto-closed since it now matches `main`

Sorry about that! I am gonna disable it for this repo. I will start a new PR once I am done with the test-case.

Issue #114, opened 2025-02-07
Add OpenBSD support


jakepetroules:
Building for OpenBSD hosts as well as building for a OpenBSD target on a OpenBSD host.

jakepetroules:
@3405691582, would you be interested in taking this? I expect the changes will be similar to #12.

3405691582:
@jakepetroules it seems like there's not an easy way to get the toolchain detected properly without installation to the base system (which I don't want to do and shouldn't have to do just yet). That might be a simple feature or change to `Sources/SWBGenericUnixPlatform/Plugin.swift`. But if I fudge this by editing  to look for `ar` instead of `swift`, I can make some progress, but then I get a strange message about `error: unknown working-directory unsupported on this platform`. I can't seem to find the `unsupported on this platform` string anywhere. Do you have any guidance?

jakepetroules:
@3405691582 I think what we need to do is implement a fork/exec fallback in llbuild when posix_spawn_file_actions_addchdir is unavailable. We need it for Amazon Linux 2 as well, which uses an old Glibc.

That said, OpenBSD is one of the only platforms where posix_spawn_file_actions_addchdir is not available, and that is a new function that's specified as part of the POSIX 2024 standard. Do you have influence in the OpenBSD community and is that something you could help push on getting added?

3405691582:
I'll investigate that option.

> Do you have influence in the OpenBSD community and is that something you could help push on getting added?

I don't. I could also look into this, but it would be something low on the list of relative priorities.

jakepetroules:
It would be nice for OpenBSD to come into POSIX compliance and add posix_spawn_file_actions_addchdir, but FWIW Swift Subprocess is switching to use fork/exec everywhere but Darwin (and we'll adopt it soon), and I added a fallback for the llbuild path in swiftlang/swift-llbuild#981, so we aren't blocked on this issue anymore.

3405691582:
I've noticed something crashing with `SIGILL` (`ILL_PRVOPC`) when using `--build-system swiftbuild`. I'll investigate further when I get some time.

3405691582:
`Sources/PackageModel/Toolchain.swift` also needs updating in `swiftpm` since it is throwing `UnknownToolchainLayout` when using `--build-system swiftbuild` on different repos.

3405691582:
Picking this up again (cc @owenv). Some brief notes.

Building swift-build is fine. Using it has some problems:

[...]

PR #365, opened 2025-03-29
OpenBSD support.

3405691582:
* Ensure the swift-crypto dependency is conditioned for the platform.

  The relevant swift-crypto changes are not landed upstream yet, but when they are, this picks up the dependency.

* Cast timestamps to time_t.

  This is done unconditionally here, because the standard specifies that `tv_sec` in `timespec` has type `time_t`, and doing so avoids a type mismatch error on the platform.

* Since pthread types are pointers here, make the usual changes to capture the fact they are optional types on the Swift side.

So far, this has been the only set of changes that have seemed necessary.

Fixes swiftlang/swift-build#114.

### DIFF
--- a/Sources/SWBUtil/Lock.swift
+++ b/Sources/SWBUtil/Lock.swift
@@ -26,6 +26,9 @@ public final class Lock: @unchecked Sendable {
     #if os(Windows)
     @usableFromInline
     let mutex: UnsafeMutablePointer<SRWLOCK> = UnsafeMutablePointer.allocate(capacity: 1)
+    #elseif os(OpenBSD)
+    @usableFromInline
+    let mutex: UnsafeMutablePointer<pthread_mutex_t?> = UnsafeMutablePointer.allocate(capacity: 1)
     #else
     @usableFromInline
     let mutex: UnsafeMutablePointer<pthread_mutex_t> = UnsafeMutablePointer.allocate(capacity: 1)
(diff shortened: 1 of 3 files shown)


3405691582:
@swift-ci please test.

owenv (approved):
lgtm, thanks!

Sources/SWBUtil/Lock.swift
```diff
@@ -26,6 +26,9 @@ public final class Lock: @unchecked Sendable {
     #if os(Windows)
     @usableFromInline
     let mutex: UnsafeMutablePointer<SRWLOCK> = UnsafeMutablePointer.allocate(capacity: 1)
+    #elseif os(OpenBSD)
+    @usableFromInline
+    let mutex: UnsafeMutablePointer<pthread_mutex_t?> = UnsafeMutablePointer.allocate(capacity: 1)
```
jakepetroules, line 31: Do you know why there's this difference between BSDs and Darwin/Linux? I needed it for FreeBSD as well.

Can we safely just use the Optional variant on all platforms?

3405691582: This happens because OpenBSD's value of `pthread_t` and subsidiary types like `pthread_mutex_t` is a pointer, whereas on Linux for example, it's typedef'd to an int (and `pthread_mutex_t` to a union). Of course, there's no nullability annotations on the 'BSDs, so Swift is interpreting the type as an optional like it would any other nullable pointer. 

After thinking about it briefly I think you have an incongruity either way, so I don't think you can use Optional always: you either have to declare, for example, `mutex` as Optional here and then conditionally check the optional or not depending on platform at the callsite to `pthread_mutex_init`, or you do what we do here, and conditionally declare `mutex` as Optional or not.

PR #367, opened 2025-03-31
Fix a race which could allow Swift tasks to be reported as failed instead of cancelled

owenv:
Ensure that if the process for a Job is never reported as started, we report it as cancelled rather than failed. This is consistent with C compiles and other task types.

rdar://138056158


owenv:
@swift-ci test

neonichu (approved):

owenv:
@swift-ci test macOS

owenv:
@swift-ci test macOS

jakepetroules (approved):

owenv:
@swift-ci test

owenv:
This needs a little more work

owenv:
@swift-ci test

PR #374, opened 2025-04-02
Move fallback developer directory determination into the plugin system

jakepetroules:
This moves some Windows and macOS specific code out of core.


jakepetroules:
@swift-ci test

Sources/SWBCore/Extensions/DeveloperDirectoryExtension.swift
```diff
@@ -0,0 +1,25 @@
+//===----------------------------------------------------------------------===//
```
owenv, line 1: This needs to be added to the make build

Sources/SWBCore/KnownFolders.swift
```diff
@@ -1,60 +0,0 @@
-//===----------------------------------------------------------------------===//
```
owenv, line 1: Has to be removed from the cmake build

owenv (changes requested):

Sources/SWBCore/Extensions/DeveloperDirectoryExtension.swift
jakepetroules: Thanks for catching this!

Sources/SWBCore/KnownFolders.swift
jakepetroules: Thanks for catching this!

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

neonichu (approved):

PR #378, opened 2025-04-03
Ensure we do not expect additional linker args to be produced in targets with Swift-produced object files

owenv:
If Swift produced an object in the linked libraries of a target, we allow it to generate additional linker options, even if it did not run to compile any source files. However, we should not attempt to inject top-level AST paths for debugging since no actual compile took place, and the paths won't exist at execution time. Use the object file type when injecting Swift tool usage in these targets so that we don't take the unintended code path.

rdar://144729550


owenv:
@swift-ci test

jakepetroules (approved):

Sources/SWBTaskConstruction/TaskProducers/BuildPhaseTaskProducers/SourcesTaskProducer.swift
```diff
@@ -896,7 +896,7 @@ final class SourcesTaskProducer: FilesBasedBuildPhaseTaskProducerBase, FilesBase
 
                 // If there is at least one object file that was built using Swift, ensure the Swift tool is present in the used tools to allow linker spec to add swift specific linker arguments.
```
jakepetroules, line 897: Is this comment still accurate?

owenv: yes, we're still adding most args, just not all of them

PR #379, opened 2025-04-06
Pass '-needed…' linker flags for XCTest libraries to suppress warnings in test targets which exclusively use Swift Testing and enable LD_WARN_UNUSED_DYLIBS

stmontgomery:
Xcode test targets are always linked with XCTest.framework and libXCTestSwiftSupport.dylib, since these flags are specified in their product type .xcspec:

```
-framework XCTest -lXCTestSwiftSupport
```

When a test target has enabled `LD_WARN_UNUSED_DYLIBS` (or manually passes `-warn_unused_dylibs` to the linker), and it exclusively uses Swift Testing (or more precisely, when it doesn’t directly link any symbols from any XCTest library listed above), currently two warnings will be shown:

```
⚠️ ld: warning: linking with (@rpath/XCTest.framework/Versions/A/XCTest) but not using any symbols from it
⚠️ ld: warning: linking with (@rpath/libXCTestSwiftSupport.dylib) but not using any symbols from it
```

These warnings should be suppressed by default. Test targets get these linker flags from the product type, and these libraries do need to be linked and loaded even if a test target doesn’t use XCTest.

This Radar tracks modifying the linker flags passed so they add `-needed_framework XCTest` and `-needed-lXCTestSwiftSupport` to suppress these warnings.

Fixes rdar://148694760


stmontgomery:
@swift-ci test

briancroom (approved):

owenv (approved):

PR #383, opened 2025-04-07
Move asset catalog code generation rule ordering into SWBApplePlatform

owenv:
Generalize the sources phase ordering support to the reference to asset catalogs can move into the SWBApplePlatform plugin


owenv:
@swift-ci test

neonichu (approved):

jakepetroules (approved):

PR #384, opened 2025-04-07
Use -Xlinker when constructing response files for Swift AST paths

owenv:
This way we don't need to vary the args based on which compiler is the linker driver

rdar://148630017


owenv:
@swift-ci test

mhrawdon:
Yes, `-Wl,` is especially bad if the path being passed itself contains a comma (the tool often fails to parse the argument). We switched to using `-Xlinker` for this reason some time ago, and I recommend to everyone that they do the same in their projects. But it's an inertia which is hard to overcome.

mhrawdon (approved):

neonichu (approved):

jakepetroules (approved):

PR #385, opened 2025-04-08
Set PATH to ensure the linker driver is able to find a linker in a different toolchain

owenv:
This is needed on macOS, where clang/swift are in the swift.org toolchain and the linker is in the Xcode default toolchain


owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test linux

jakepetroules (approved):

PR #386, opened 2025-04-08
Remove support for compiling with bitcode.

mhrawdon:
This removes support for compiling files with bitcode from SWBuild, as this functionality is no longer supported. This removes a lot of functionally dead code.

Support for stripping bitcode from copied files remains, as there are extant libraries and XCFrameworks which contain bitcode, which must be stripped. Bitcode was only supported for a subset of platforms, to the assignment of `STRIP_BITCODE_FROM_COPIED_FILES` in `Settings` based on platform remains, so time isn't spent stripping nonexistent bitcode from other platforms.

Some interfaces which Xcode uses continue to refer to bitcode for compatibility, but effectively do nothing.

Note that the references to bitcode in the OpenCL support are a concept internal to that tool, and independent of the support being removed here.

rdar://148287993


owenv (approved):

Sources/SWBApplePlatform/Specs/watchOS Shared.xcspec
```diff
@@ -130,35 +130,9 @@
             "sourcecode.cpp.objcpp",
             {
                 "FileType" = "sourcecode.asm";
-                "Condition" = "! $(ENABLE_BITCODE)";
```
jakepetroules, line 133: I think this will inadvertently block compiling assembly code on watchOS. Instead I think you should delete this entire compiler spec and let it inherit from the default spec.

jakepetroules (changes requested):

Sources/SWBApplePlatform/Specs/watchOS Shared.xcspec
mhrawdon: Huh, I'm not sure whether you're right that it will block compiling assembly on watchOS, but it looks like SWBuild never supported the `Condition` or `ErrorMessage` keys in this dictionary (the legacy build system may have). So all of that here seems to be functionally a no-op, and removing it is hopefully the right thing to do.

mhrawdon:
@swift-ci test

Sources/SWBApplePlatform/Specs/watchOS Shared.xcspec
mhrawdon: Done!

jakepetroules (approved):

PR #387, opened 2025-04-08
Refactor nextValueForMacro so it does not rely on tail call optimization

owenv:
Avoid overflowing the stack for large settings expressions when TCO fails to kick in


owenv:
@swift-ci test

jakepetroules (approved):

mhrawdon (approved):

owenv:
@swift-ci test macOS

PR #388, opened 2025-04-09
Simplify cross-platform hashing implementations

owenv:
Eliminate the package dependency on swift-crypto to simplify the bootstrapped build. Instead, vendor the SHA256 implementation from swift-tools-support-core and use it as a fallback if neither WinSDK nor CryptoKit is available.

### DIFF
--- a/Sources/SWBUtil/HashContext.swift
+++ b/Sources/SWBUtil/HashContext.swift
@@ -14,8 +14,6 @@
 import WinSDK
 #elseif canImport(CryptoKit)
 private import CryptoKit
-#else
-private import Crypto
 #endif
 
 public import Foundation
@@ -104,7 +102,7 @@ fileprivate final class BCryptHashContext: HashContext {
 
 @available(*, unavailable)
 extension BCryptHashContext: Sendable { }
-#else
+#elseif canImport(CryptoKit)
 fileprivate final class SwiftCryptoHashContext<HF: HashFunction>: HashContext {
     @usableFromInline
     internal var hash = HF()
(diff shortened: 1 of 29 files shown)


owenv:
@swift-ci test

owenv:
@swift-ci test macOS

owenv:
@swift-ci test

owenv:
@swift-ci test linux

owenv:
@swift-ci test

owenv:
@swift-ci test

Sources/SWBUtil/HashContext.swift
```diff
@@ -173,25 +364,29 @@ public class DelegatedHashContext: HashContext {
...
-public final class MD5Context: DelegatedHashContext {
+public final class InsecureHashContext: DelegatedHashContext {
     public init() {
         #if os(Windows)
         super.init(impl: BCryptHashContext(algorithm: "MD5", digestLength: 16))
-        #else
+        #elseif canImport(CryptoKit)
         super.init(impl: SwiftCryptoHashContext<Insecure.MD5>())
+        #else
+        super.init(impl: VendoredSHA256HashContext())
```
jakepetroules, line 374: This seems a little shaky to use MD5 on Windows and Apple but SHA256 on other platforms. Do you plan to unify this further? I'm even surprised the tests pass like this.

jakepetroules (approved):

owenv:
@swift-ci test

Sources/SWBUtil/HashContext.swift
owenv: I think we should probably standardize on SHA256 everywhere in the future. Very few tests actually depend on the specific hash values, it's just important that they remain stable across builds performed by the same build system executable

PR #390, opened 2025-04-09
Add initial support for generating non-Darwin test entrypoints

owenv:
Add basic support for generating an entry point when we build tests as executables on non-Darwin. Right now, this does not yet support XCTest test discovery (just Swift Testing), and is missing support for experimental test output. I decided to break those out into future changes.

### DIFF
--- a/Sources/SWBUniversalPlatform/Plugin.swift
+++ b/Sources/SWBUniversalPlatform/Plugin.swift
@@ -13,9 +13,13 @@
 public import SWBUtil
 import SWBCore
 import Foundation
+import SWBTaskConstruction
+import SWBTaskExecution
 
 @PluginExtensionSystemActor public func initializePlugin(_ manager: PluginManager) {
     manager.register(UniversalPlatformSpecsExtension(), type: SpecificationsExtensionPoint.self)
+    manager.register(UniversalPlatformTaskProducerExtension(), type: TaskProducerExtensionPoint.self)
+    manager.register(UniversalPlatformTaskActionExtension(), type: TaskActionExtensionPoint.self)
 }
 
 struct UniversalPlatformSpecsExtension: SpecificationsExtension {
@@ -26,6 +30,7 @@ struct UniversalPlatformSpecsExtension: SpecificationsExtension {
             CppToolSpec.self,
             LexCompilerSpec.self,
             YaccCompilerSpec.self,
+            TestEntryPointGenerationToolSpec.self,
         ]
     }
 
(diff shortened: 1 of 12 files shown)


owenv:
@swift-ci test

Sources/SWBUniversalPlatform/Plugin.swift
```diff
@@ -13,9 +13,13 @@
 public import SWBUtil
 import SWBCore
 import Foundation
+import SWBTaskConstruction
+import SWBTaskExecution
 
 @PluginExtensionSystemActor public func initializePlugin(_ manager: PluginManager) {
     manager.register(UniversalPlatformSpecsExtension(), type: SpecificationsExtensionPoint.self)
+    manager.register(UniveralPlatformTaskProducerExtension(), type: TaskProducerExtensionPoint.self)
```
jakepetroules, line 21: typo: "Univeral"

Sources/SWBUniversalPlatform/TestEntryPointGenerationTaskAction.swift
```diff
@@ -0,0 +1,83 @@
...
+    override func performTaskAction(_ task: any ExecutableTask, dynamicExecutionDelegate: any DynamicTaskExecutionDelegate, executionDelegate: any TaskExecutionDelegate, clientDelegate: any TaskExecutionClientDelegate, outputDelegate: any TaskOutputDelegate) async -> CommandResult {
+        do {
+            let options = try Options.parse(Array(task.commandLineAsStrings.dropFirst()))
+            try executionDelegate.fs.write(options.output, contents: #"""
+            #if canImport(Testing)
+            import Testing
+            #endif
+
+            @main
+            @available(macOS 10.15, iOS 11, watchOS 4, tvOS 11, *)
```
jakepetroules, line 32: What about visionOS?

Sources/SWBUniversalPlatform/TestEntryPointGenerationTaskAction.swift
```diff
@@ -0,0 +1,83 @@
...
+                            return libraryName.lowercased()
+                        }
+                    }
+
+                    // Fallback if not specified: run XCTest (legacy behavior)
+                    return "xctest"
+                }
+            
+                #if os(Linux)
+                @_silgen_name("$ss13_runAsyncMainyyyyYaKcF")
```
jakepetroules, line 48: Out of curiosity why's this needed only on Linux, but not Windows, FreeBSD, Android, etc?

owenv: This is a hack because of crashes on Amazon linux 2

owenv:
@swift-ci test

Sources/SWBUniversalPlatform/TestEntryPointGenerationTaskAction.swift
jakepetroules: Ah, OK

jakepetroules (approved):

dschaefer2:
I was just in the test entry point and discovery code on the native side. Glad it integrated easily. Thanks!

PR #391, opened 2025-04-09
Always record backtraces when checking for null builds

jakepetroules:
This should hopefully give us more insight into flaky tests.

### DIFF
--- a/Sources/SWBTestSupport/BuildOperationTester.swift
+++ b/Sources/SWBTestSupport/BuildOperationTester.swift
@@ -19,7 +19,7 @@ package import SWBBuildSystem
 @_spi(Testing) package import SWBCore
 package import SWBTaskConstruction
 package import SWBTaskExecution
-package import SWBUtil
+@_spi(Testing) package import SWBUtil
 
 private import class SWBLLBuild.BuildDB
 private import class SWBLLBuild.BuildKey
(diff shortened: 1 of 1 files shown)


jakepetroules:
@swift-ci test

Sources/SWBTestSupport/BuildOperationTester.swift
```diff
@@ -713,6 +713,21 @@ package final class BuildOperationTester {
 
         }
 
+        package func checkNoTaskWithBacktraces(_ conditions: TaskCondition..., sourceLocation: SourceLocation = #_sourceLocation) {
+            for matchedTask in findMatchingTasks([]) {
+                Issue.record("found unexpected task matching conditions '\(conditions)', found: \(matchedTask)", sourceLocation: sourceLocation)
+                let backtraces = events.compactMap ({ (event) -> String? in
```
owenv, line 719: This will show the first backtrace frame instead of following the full chain, but that might be enough in a lot of cases

jakepetroules: Ah, good point. I went ahead and just did it properly; we already had some backtrace enumeration code in other helpers already.

jakepetroules:
@swift-ci test

owenv (approved):

PR #397, opened 2025-04-10
Fix issue causing QNX plugin to create "null.o" in the current directory

jakepetroules:
Send that to /dev/null instead


jakepetroules:
@swift-ci test

owenv (approved):

PR #398, opened 2025-04-10
Adapt QNX support to the introduction of LINKER_DRIVER

jakepetroules:


jakepetroules:
I'm not thrilled with how QNX leaks out of its plugin. I think LINKER_DRIVER can't be an enum and this has to be done in a somewhat more flexible way instead. Any particular ideas?

Maybe LINKER_DRIVER can be a plain string, and we set it to clang by default in Settings.swift. For LD_OPTIMIZATION_LEVEL I think we need something like LINKER_DRIVER_IS_GCC_COMPATIBLE and maybe an extension point that returns a list of gcc-compatible linker drivers?

dschaefer2:
> I'm not thrilled with how QNX leaks out of its plugin.

I'm not thrilled swift-build needs to know about QNX. I look forward to adding it to the Swift SDK story and how swift-build can deal with new platforms it's never heard about :).

jakepetroules:
We'll see. Swift SDKs can only go so far, even the compiler has to have knowledge of specific platforms to a certain point.

jakepetroules:
@swift-ci test

PR #399, opened 2025-04-10
Disable explicit module builds for the index arena

bnbarham:
Explicit module arguments are removed for the arguments returned for indexing. Also disable them for preparation in general so we're not mixing between the two.

Resolves rdar://147354913.


bnbarham:
@swift-ci please test

owenv (approved):

Issue #395, opened 2025-04-10
Unrecognized '-rdynamic' flag when using swiftc linker driver


cmcgee1024:
While using the swiftc linker driver on Linux with the following setting override from SwiftPM:

```
LINKER_DRIVER=swiftc
```

I get the following error that the '-rdynamic' is an unknown argument:

```
/usr/bin/swiftc -target aarch64-unknown-linux-gnu -sysroot / -L/tmp/Miscellaneous_TestMultipleFailureXCTest.YZGbOI/Miscellaneous_TestMultipleFailureXCTest/.build/aarch64-unknown-linux-gnu/Intermediates.noindex/EagerLinkingTBDs/Debug-linux -L/tmp/Miscellaneous_TestMultipleFailureXCTest.YZGbOI/Miscellaneous_TestMultipleFailureXCTest/.build/aarch64-unknown-linux-gnu/Products/Debug-linux -L/usr/lib @/tmp/Miscellaneous_TestMultipleFailureXCTest.YZGbOI/Miscellaneous_TestMultipleFailureXCTest/.build/aarch64-unknown-linux-gnu/Intermediates.noindex/TestMultipleFailureXCTest.build/Debug-linux/TestMultipleFailureXCTestTests.build/Objects-normal/aarch64/TestMultipleFailureXCTestTests.LinkFileList -Xlinker -rpath -Xlinker @loader_path/Frameworks -Xlinker -rpath -Xlinker @loader_path/../Frameworks -rdynamic -lswiftCore -lswiftSwiftOnoneSupport -L/usr/lib/swift/linux -L/usr/lib/swift @/tmp/Miscellaneous_TestMultipleFailureXCTest.YZGbOI/Miscellaneous_TestMultipleFailureXCTest/.build/aarch64-unknown-linux-gnu/Intermediates.noindex/TestMultipleFailureXCTest.build/Debug-linux/TestMultipleFailureXCTestTests.build/Objects-normal/aarch64/TestMultipleFailureXCTestTests.autolink -o /tmp/Miscellaneous_TestMultipleFailureXCTest.YZGbOI/Miscellaneous_TestMultipleFailureXCTest/.build/aarch64-unknown-linux-gnu/Products/Debug-linux/TestMultipleFailureXCTestTests.xctest
info: error: unknown argument: '-rdynamic'
```

cmcgee1024:
It could be related to the setting `ENABLE_TESTABILITY=YES` that's set in the PIF coming from SwiftPM.

PR #400, opened 2025-04-10
Fix LD_EXPORT_GLOBAL_SYMBOLS when using swiftc as the linker driver

owenv:
https://github.com/swiftlang/swift-build/issues/395

Closes: #395

### DIFF
--- a/Sources/SWBGenericUnixPlatform/Specs/UnixLd.xcspec
+++ b/Sources/SWBGenericUnixPlatform/Specs/UnixLd.xcspec
@@ -146,6 +146,18 @@
                 Condition = "NO";
             },
             {
+                Name = "SWIFTC_LD_EXPORT_GLOBAL_SYMBOLS";
+                Type = Boolean;
+                DefaultValue = "$(LD_EXPORT_GLOBAL_SYMBOLS)";
+                Condition = "$(LINKER_DRIVER) == swiftc";
+                CommandLineArgs = {
+                    YES = (
+                        "-Xlinker", "--export-dynamic"
+                    );
+                    NO = ();
+                };
+            },
+            {
                 Name = "__INPUT_FILE_LIST_PATH__";
                 Type = Path;
                 DefaultValue = "$(LINK_FILE_LIST_$(variant)_$(arch))";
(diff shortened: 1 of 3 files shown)


owenv:
@swift-ci test

jakepetroules:
Gentle reminder that if this is meant to close #395, the PR description should say `Closes: #395` so GitHub hooks those up.

jakepetroules (approved):

Sources/SWBGenericUnixPlatform/Specs/UnixLd.xcspec
```diff
@@ -145,6 +145,18 @@
...
                 Condition = "NO";
             },
+            {
+                Name = "SWIFTC_LD_EXPORT_GLOBAL_SYMBOLS";
+                Type = Boolean;
+                DefaultValue = "$(LD_EXPORT_GLOBAL_SYMBOLS)";
+                Condition = "$(LINKER_DRIVER) == swiftc";
+                CommandLineArgs = {
+                    YES = (
+                        "-Xlinker", "--export-dynamic"
```
compnerd, line 155: Why not use `-Xclang-linker -rdynamic` instead?

cmcgee1024 (approved):

Sources/SWBGenericUnixPlatform/Specs/UnixLd.xcspec
dschaefer2: -rdynamic isn't available at all on Linux.

dschaefer2:
The question I had, is the behaviour -rdynamic provided not required on Linux?

Sources/SWBGenericUnixPlatform/Specs/UnixLd.xcspec
owenv: The clang driver accepts -rdynamic everywhere. I generally agree we should be using -Xclang-linker in more places and delegate platform differences to the driver. Will put up a PR in a bit

compnerd: @dschaefer2 `-rdynamic` is a platform agnostic spelling, can you help me understand why you believe it is unavailable on Linux?

dschaefer2: I get the following on Linux

warning: unknown Argument unused during compilation: '-rdynamic' [-Wunused-command-line-argument] []

compnerd: Yes, that is expected! _during compilation_ is the operative portion of the message. This flag has no impact on compilation, it is a link time flag only. This is why `-Xclang-linker` is used rather than `-Xcc`. This flag is to be passed to the linker driver (`clang`).

PR #401, opened 2025-04-10
Avoid nondeterministically clearing the build description manager's build description cache in tests

jakepetroules:
If we've told the build description manager to never evict build descriptions from its cache, ensure that the cache isn't attached to the heavy cache global state, so that it isn't cleared by other tests running concurrently.


jakepetroules:
@swift-ci test

owenv (approved):

PR #402, opened 2025-04-10
Remove workaround for executable files on Windows

jakepetroules:
Use our FSProxy abstraction instead of Foundation, which doesn't suffer from https://github.com/swiftlang/swift-foundation/issues/860


jakepetroules:
@swift-ci test

neonichu (approved):

PR #403, opened 2025-04-11
Simplify some settings definitions using -Xclang-linker

owenv:
Rely on clang to handle platform differences

### DIFF
--- a/Sources/SWBUniversalPlatform/Specs/Ld.xcspec
+++ b/Sources/SWBUniversalPlatform/Specs/Ld.xcspec
@@ -400,9 +400,9 @@
                 CommandLineArgs = {
                     "" = ();
                     "<<otherwise>>" = (
-                        "-Xlinker",
-                        "-dylib_install_name",
-                        "-Xlinker",
+                        "-Xclang-linker",
+                        "-install_name",
+                        "-Xclang-linker",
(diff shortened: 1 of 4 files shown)


owenv:
cc @compnerd

owenv:
@swift-ci test

jakepetroules (approved):

Sources/SWBUniversalPlatform/Specs/Ld.xcspec
```diff
@@ -400,9 +400,9 @@
                 CommandLineArgs = {
                     "" = ();
                     "<<otherwise>>" = (
-                        "-Xlinker",
-                        "-dylib_install_name",
-                        "-Xlinker",
+                        "-Xclang-linker",
+                        "-install_name",
```
jakepetroules, line 404: -install_name translates to -dylib_install_name on Mach-O platforms and -soname on ELF platforms?

compnerd: Sadly, `-install_name` is a driver flag that is Darwin specific, and will be translated to `-Xlinker -dylib_install_name`. It has no impact on ELFish targets.

neonichu (approved):

PR #408, opened 2025-04-14
Add FreeBSD platform to ProjectModel API

pmattos:
Add the `freebsd` platform to the new PIF encoding API (i.e., `ProjectModel`).

We need this when sending over the PIF JSON from SwiftPM (with the new `--build-system swiftbuild` option) given SwiftPM `PackageModel.Platform` already supports FreeBSD.


jakepetroules:
@swift-ci test macOS

PR #412, opened 2025-04-15
Fix linker swiftmodule registration in incremental builds

owenv:
Ensure we always enumerate the entire module graph looking for swift modules to register, as opposed to the subset of modules we built.

rdar://149248232

### DIFF
--- a/Sources/SWBCore/LibSwiftDriver/LibSwiftDriver.swift
+++ b/Sources/SWBCore/LibSwiftDriver/LibSwiftDriver.swift
@@ -177,6 +177,36 @@ public final class SwiftModuleDependencyGraph: SwiftGlobalExplicitDependencyGrap
         }
     }
 
+    public func querySwiftmodulesNeedingRegistrationForDebugging(for key: String) throws -> [String] {
+        let graph = try registryQueue.blocking_sync {
+            guard let driver = registry[key] else {
+                throw StubError.error("Unable to find jobs for key \(key). Be sure to plan the build ahead of fetching results.")
+            }
+            return driver.intermoduleDependencyGraph
+        }
+        guard let graph else { return [] }
+        var swiftmodulePaths: [String] = []
+        swiftmodulePaths.reserveCapacity(graph.modules.values.count)
+        for (_, moduleInfo) in graph.modules.sorted(byKey: { $0.moduleName < $1.moduleName }) {
+            guard moduleInfo != graph.mainModule else {
+                continue
+            }
+            switch moduleInfo.details {
+            case .swift:
+                if let modulePath = VirtualPath.lookup(moduleInfo.modulePath.path).absolutePath {
+                    swiftmodulePaths.append(modulePath.pathString)
+                }
+            case .swiftPrebuiltExternal(let details):
+                if let modulePath = VirtualPath.lookup(details.compiledModulePath.path).absolutePath {
+                    swiftmodulePaths.append(modulePath.pathString)
+                }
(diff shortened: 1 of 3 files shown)


Sources/SWBCore/LibSwiftDriver/LibSwiftDriver.swift
```diff
@@ -177,6 +177,38 @@ public final class SwiftModuleDependencyGraph: SwiftGlobalExplicitDependencyGrap
...
 
+    public func querySwiftmodulesNeedingRegistrationForDebugging(for key: String) throws -> [String] {
+        let graph = try registryQueue.blocking_sync {
+            guard let driver = registry[key] else {
+                throw StubError.error("Unable to find jobs for key \(key). Be sure to plan the build ahead of fetching results.")
+            }
+            return driver.intermoduleDependencyGraph
+        }
+        guard let graph else { return [] }
+        let directDependencies = graph.mainModule.directDependencies ?? []
```
artemcm, line 188: We are trying to move to a new `allDependencies` computed property:
https://github.com/swiftlang/swift-driver/pull/1869
So that `directDependencies` actually captures only the directly-imported modules. 

On top of that, the expectation is that `graph.modules` *is* the transitive closure of the `mainModule` dependencies, so if our goal is simply to iterate over all of them then I don't think we need to compute it separately here - it should be sufficient to do `for dependency in graph.modules {` instead.

owenv: makes sense, I updated it to just use `graph.modules` instead

owenv:
@swift-ci test

Sources/SWBCore/LibSwiftDriver/LibSwiftDriver.swift
```diff
@@ -177,23 +177,44 @@ public final class SwiftModuleDependencyGraph: SwiftGlobalExplicitDependencyGrap
...
                 throw StubError.error("Unable to find jobs for key \(key). Be sure to plan the build ahead of fetching results.")
             }
             return driver.intermoduleDependencyGraph
         }
         guard let graph else { return [] }
-        let directDependencies = graph.mainModule.directDependencies ?? []
-        let transitiveDependencies = Set(directDependencies + SWBUtil.transitiveClosure(directDependencies, successors: { moduleID in graph.modules[moduleID]?.directDependencies ?? [] }).0)
+        var swiftmodulePaths: [String] = []
+        swiftmodulePaths.reserveCapacity(graph.modules.values.count)
+        for (_, moduleInfo) in graph.modules.sorted(byKey: { $0.moduleName < $1.moduleName }) {
```
artemcm, line 190: Hmm, I think the only exception is that we may still need to explicitly skip over the main module itself here. Its `modulePath` in the graph is just `<Module_Name>.swiftmodule` and that won't be a thing that the linker will be able to resolve.

artemcm: Would also mean that right above this we instead need to:
```
swiftmodulePaths.reserveCapacity(graph.modules.values.count - 1)
```

owenv: fixed this up in the latest commit - I left the reserveCapacity calls as-is since that's an approximation anyway

owenv:
@swift-ci test

artemcm (approved):

mirza-garibovic (approved):

PR #413, opened 2025-04-16
Annotate "$targetA is missing a dependency on $targetB" diagnostics with target location (rdar://149397709)

mirza-garibovic:
Having target annotations here provides metadata for e.g. xcresult bundles and then makes it possible for automation to group issues by project and target.


mirza-garibovic:
@swift-ci test

owenv (approved):

Issue #21, opened 2025-01-31
Nondeterministic test failure with AsyncStream + Dispatch interop


jakepetroules:
Occasionally tests such as `linkerSignedBinaryDetection` will fail with an error message like:

    Caught error: dataStream(reading: DispatchFD(rawValue: 9))#1(): Is a directory (21)

This is due to the file descriptor that is read from in Dispatch+Async.swift becoming invalid, but it's not clear why this happens.

PR #421, opened 2025-04-18
CLIConnection tests: close output stream on shutdown

jakepetroules:
Noticed this during a bit of investigation into #21, and don't necessarily expect this to fix it, but it seems correct in order to prevent the fd possibly being closed while the stream is still being read.


jakepetroules:
@swift-ci test

neonichu (approved):

PR #423, opened 2025-04-19
Ensure long paths are handled correctly in calls to Win32 APIs

jakepetroules:
Use the same technique being proposed in SwiftSystem and partly-implemented in Foundation.

As a follow-on change, we also need to stop using POSIX filesystem APIs, which are not Unicode or long path aware.

### DIFF
--- a/Sources/SWBUtil/Library.swift
+++ b/Sources/SWBUtil/Library.swift
@@ -33,7 +33,7 @@ public enum Library: Sendable {
     @_alwaysEmitIntoClient
     public static func open(_ path: Path) throws -> LibraryHandle {
         #if os(Windows)
-        guard let handle = path.withPlatformString(LoadLibraryW) else {
+        guard let handle = try path.withPlatformString({ p in try p.withCanonicalPathRepresentation({ LoadLibraryW($0) }) }) else {
             throw LibraryOpenError(message: Win32Error(GetLastError()).description)
         }
         return LibraryHandle(rawValue: handle)
--- a/Sources/SWBUtil/CMakeLists.txt
+++ b/Sources/SWBUtil/CMakeLists.txt
@@ -68,6 +68,7 @@ add_library(SWBUtil
   OutputByteStream.swift
   Pair.swift
   Path.swift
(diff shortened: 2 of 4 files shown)


jakepetroules:
@swift-ci test

jakepetroules:
Same approach as in apple/swift-system#226

compnerd (commented):
I think that the `FSProxy` is better adjusted to simply use Foundation's `FileManager` for these operations rather than re-implementing them.

jakepetroules:
Sure, I can do that for the FSProxy bits. Still need our own copy of "withNTPathRepresentation" for LoadLibraryW and any other Win32 APIs we might need to directly call in future which don't have Foundation analogues.

I also fixed up the path handling in Foundation itself, take a look: swiftlang/swift-foundation#1257

jakepetroules:
Moved the FSProxy changes into #427, as far as existing code goes, this is now just the one-line change to LoadLibraryW.

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

Sources/SWBUtil/Library.swift
```diff
@@ -33,7 +33,7 @@ public enum Library: Sendable {
     @_alwaysEmitIntoClient
     public static func open(_ path: Path) throws -> LibraryHandle {
         #if os(Windows)
-        guard let handle = path.withPlatformString(LoadLibraryW) else {
+        guard let handle = try path.withPlatformString({ p in try p.withCanonicalPathRepresentation({ LoadLibraryW($0) }) }) else {
```
compnerd, line 36: I question if `withPlatformString` really helps readability here or not.

compnerd (approved):

Sources/SWBUtil/Library.swift
jakepetroules: It's not just for readability, my extension is on UnsafePointer rather than on String like Foundation's (felt it's more flexible that way).

PR #424, opened 2025-04-21
[CAS] For compilation tasks, add an unintrusive cache hit/miss indicator

akyrtzi:
This goes in the task's output, so in the Xcode build log the user would have to expand the task output to see such detail.

Looks like this for clang:
cache hit indication 2

and for Swift:
cache hit indication 3 swift


akyrtzi:
@swift-ci Please test

owenv (approved):

PR #426, opened 2025-04-21
Expand use of Win32Error

jakepetroules:
Have its string representation print the actual error message for improved debuggability.


jakepetroules:
@swift-ci test

owenv (approved):

PR #430, opened 2025-04-23
Pass -ftyped-cxx-new-delete as a linker arg.

usama54321:
The motivation here is to make back deployment easier by conditionally linking in a shim layer in the clang driver.

rdar://149709799


usama54321:
@swift-ci test

neonichu (approved):

jakepetroules (approved):

PR #431, opened 2025-04-23
Check for the index workspace description rather than index arena

bnbarham:
There's various paths that should be specific to the index workspace description rather than if the index arena is enabled. Specifically the recent change to replace parameters for host tools (https://github.com/swiftlang/swift-build/pull/345) broke the *package* description as it uses the passed in run destination rather than configuring for all platforms. This replacement would then create two configured targets for macOS for the same target (one for the given run destination and one for the macOS platform that the index uses).

This also changes the *target* description to only build for the run destination, which makes sense given its entire purpose is to be a replacement for the workspace description if it hasn't built yet (ie. it should be as fast as possible).

Resolves rdar://149896775.


bnbarham:
@swift-ci please test

neonichu (approved):

Tests/SWBBuildSystemTests/HostBuildToolBuildOperationTests.swift
```diff
@@ -439,6 +439,184 @@ fileprivate struct HostBuildToolBuildOperationTests: CoreBasedTests {
...
+                stream <<<
+                """
+                public class MyClass {}
+                """
+            }
+
+            try await tester.checkIndexBuild(prepareTargets: hostToolsPackage.targets.map(\.guid), workspaceOperation: false, runDestination: .anyMac, persistent: true) { results in
+//                results.checkWarning(.contains("annotation implies no releases, but consumes self"), failIfNotFound: false)
+//                results.checkWarning(.contains("mismatching function effects"), failIfNotFound: false)
+//                results.checkWarning(.contains("mismatching function effects"), failIfNotFound: false)
```
bnbarham, line 609: Whoops

bnbarham:
@swift-ci please test

jakepetroules (approved):

bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

bnbarham:
(just a bunch more test updates for tests using SDKs I didn't have locally)

PR #432, opened 2025-04-23
App Sandbox build setting ends up in iOS signed app

baujla:
If the app sandbox entitlement ends up in an iOS app, the app is not installable on device as the corresponding provisioning profile cannot include this entitlement. The app sandbox and hardened runtime entitlements are only applicable on macOS.


baujla:
@swift-ci please test

jakepetroules (approved):

PR #437, opened 2025-04-24
Add new libcxx flag for disabling container overflow checks for Asan.

thetruestblue:
libcxx 20 has changed the naming scheme of internal macros that indicate feature availability. This affects the name of the macro used to include sanitizer container annotations -- it was renamed from `_LIBCPP_HAS_NO_ASAN` -> `_LIBCPP_HAS_ASAN`.

libcxx change: https://github.com/llvm/llvm-project/pull/89178

This change adds the new flag name to maintain current behavior, which is to disable these checks by default.
Sanitizer container overflow checks are prone to false positives when vector objects are modified outside of instrumented code -- so we have historically defaulted to disabling this on Xcode projects via the libcxx macro.

rdar://149350723

### DIFF
--- a/Sources/SWBUniversalPlatform/Specs/Clang LLVM 1.0.xcspec
+++ b/Sources/SWBUniversalPlatform/Specs/Clang LLVM 1.0.xcspec
@@ -2736,6 +2736,7 @@
                     YES = ();
                     NO = (
                         "-D_LIBCPP_HAS_NO_ASAN",
+                        "-D_LIBCPP_HAS_ASAN=0",
                     );
                 };
                 Condition = "$(CLANG_ADDRESS_SANITIZER)";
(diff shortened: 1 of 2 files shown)


thetruestblue:
@swift-ci please test

owenv (approved):

neonichu (approved):

owenv:
@swift-ci test linux

thetruestblue:
> Thank you for finding and fixing this! Have you tested this change with both an old SDK (that doesn't support `_LIBCPP_HAS_ASAN`) and a new SDK (that doesn't support `_LIBCPP_HAS_NO_ASAN`) to make sure there are no macro-related warnings produced by the "extra" definition?

Yep I tested on both older and newer SDKs, there were no macro-related warnings in either case. 

I wouldn't expect unused or extra macro definitions to trigger a warning, but I did check and confirmed.

PR #438, opened 2025-04-25
Update README.md instructions for SwiftPM

pepicrft:
I noticed SwiftPM only supports `--build-system native/xcode/next` so the instructions in the README don't seem accurate. I'm updating them to `--build-system next` assuming that "next" refers to swift-build. Let me know otherwise.

### Environment

```
swift-driver version: 1.120.5 Apple Swift version 6.1 (swiftlang-6.1.0.110.21 clang-1700.0.13.3)
Target: arm64-apple-macosx15.0
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Usage
 
 ### With SwiftPM
 
-When building SwiftPM from sources which include Swift Build integration, passing `--build-system swiftbuild` will enable the new build-system. This functionality is not currently available in nightly toolchains.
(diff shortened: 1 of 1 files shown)


owenv:
thanks for the PR - since this section of the README is specifically referring to building SwiftPM from source, I'd prefer to keep it reflecting the naming used on `main` and `release/6.2`, which is `--build-system swiftbuild` (defined in https://github.com/swiftlang/swift-package-manager/blob/ed3046151ceea7d51726f595b64a6345a8fc7cf6/Sources/SPMBuildCore/BuildSystem/BuildSystem.swift#L133)

pepicrft:
That's a good idea. Thanks @owenv for the suggestion :)

PR #446, opened 2025-04-27
[CI] Add support for GitHub Actions

shahmishal:


.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,19 @@
...
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+
+jobs:
+  tests:
+    name: Test
+    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@main
+    with:
```
jakepetroules, line 11: Please set `linux_os_versions` to the full list of distributions; the coverage is rather important for this project as it has some distro-specific behaviors.

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,19 @@
...
+      linux_exclude_swift_versions: '[{"swift_version": "5.9"}, {"swift_version": "5.10"}]'
+      linux_pre_build_command: apt-get update && apt-get install -y libsqlite3-dev libncurses5-dev
```
jakepetroules, line 13: Can this be a reference to an external script file or does it have to be inline? Because you'll need something a bit longer to handle both Debian and RedHat lineages:

```sh
command -v apt >/dev/null 2>&1 && apt update && apt install -y libsqlite3-dev || (command -v yum >/dev/null 2>&1 && yum update -y && yum install -y sqlite-devel)
```

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,19 @@
...
+      windows_exclude_swift_versions: '[{"swift_version": "5.9"}, {"swift_version": "5.10"}]'
+  soundness:
+    name: Soundness
+    uses: swiftlang/github-workflows/.github/workflows/soundness.yml@main
+    with:
+       license_header_check_project_name: "Swift"
```
jakepetroules, line 19: YAML indentation issue

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,19 @@
...
+    types: [opened, reopened, synchronize]
+
+jobs:
+  tests:
+    name: Test
+    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@main
+    with:
+      linux_exclude_swift_versions: '[{"swift_version": "5.9"}, {"swift_version": "5.10"}]'
+      linux_pre_build_command: apt-get update && apt-get install -y libsqlite3-dev libncurses5-dev
+      windows_exclude_swift_versions: '[{"swift_version": "5.9"}, {"swift_version": "5.10"}]'
```
jakepetroules, line 14: We might just want `macos_xcode_versions: '["16.3"]'` as I'm not sure there's value in testing on earlier macOS versions. @owenv, what do you think?

jakepetroules (changes requested):

.github/workflows/pull_request.yml
owenv: yes, I think we can restrict this to just 16.3

owenv:
We should probably turn off api_breakage_check_enabled as well since we don't expose a long-term stable API

owenv:
6.0 jobs fix: https://github.com/swiftlang/swift-build/pull/447

owenv:
shellcheck fix: https://github.com/swiftlang/swift-build/pull/448

.licenseignore
```diff
@@ -0,0 +1,12 @@
+**/*.mlmodel
+**/*.pbxproj
+**/*.png
+**/*.strings
+**/*.xcbuildrules
```
jakepetroules, line 5: I don't think we should ignore license headers for strings or xcbuildrules or xcspec files. All of those should (and current do) have a valid license header. The rest make sense.

.github/workflows/pull_request.yml
shahmishal: macOS support will be added in the future.

.github/workflows/pull_request.yml
```diff
@@ -4,14 +4,19 @@ on:
...
 jobs:
   tests:
     name: Test
     uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@main
     with:
-      linux_exclude_swift_versions: '[{"swift_version": "5.9"}, {"swift_version": "5.10"}]'
-      linux_pre_build_command: apt-get update && apt-get install -y libsqlite3-dev libncurses5-dev
-      windows_exclude_swift_versions: '[{"swift_version": "5.9"}, {"swift_version": "5.10"}]'
+      linux_os_versions: '["noble", "jammy", "focal", "rhel-ubi9"]'
+      linux_pre_build_command: command -v apt >/dev/null 2>&1 && apt update && apt install -y libsqlite3-dev libncurses5-dev || (command -v yum >/dev/null 2>&1 && yum update -y && yum install -y sqlite-devel)
```
jakepetroules, line 17: On RedHat it's `libncurses-dev`; I'd add that too.

shahmishal:
```
No match for argument: libncurses-dev
Error: Unable to find a match: libncurses-dev
```

Going to revert back to not including it.

jakepetroules: Oh, sorry, checked again and it's `ncurses-devel`.

I was looking at the Debian metapackage before -- and on Debian we should use `libncurses-dev` rather than `libncurses5-dev`; the latter is a transitional package for `libncurses-dev`.

jakepetroules:
Try adding `--no-parallel` to `swift test` so we can narrow down which test is crashing?

jakepetroules (approved):
Do you want to land build-only as a first round?

shahmishal:
@swift-ci test

PR #449, opened 2025-04-28
Improve cmake-smoke-test command plugin

owenv:
- Make --sysroot-path optional
- Fail early if --disable-sandbox is not passed
- Use SwiftPM's checkouts of dependencies instead of assuming the update-checkout layout


owenv:
@swift-ci test

jakepetroules (approved):

PR #452, opened 2025-04-28
Inclusive language: rename master object file to prelinked object file

jakepetroules:
This renames occurrences of "master" in relation to this feature to "prelink" or "prelinked". There are 3 remaining exceptions, which is to default the GENERATE_PRELINK_OBJECT_FILE build setting to GENERATE_MASTER_OBJECT_FILE for backwards compatibility.

### DIFF
--- a/Sources/SWBCore/Specs/en.lproj/CoreBuildSystem.strings
+++ b/Sources/SWBCore/Specs/en.lproj/CoreBuildSystem.strings
@@ -409,8 +409,8 @@ Generally you should not specify an order file in Debug or Development configura
 "[LINKER_DISPLAYS_MANGLED_NAMES]-name" = "Display Mangled Names";
 "[LINKER_DISPLAYS_MANGLED_NAMES]-description" = "Activating this setting causes the linker to display mangled names for C++ symbols. Normally, this is not recommended, but turning it on can help to diagnose and solve C++ link errors.";
 
-"[GENERATE_MASTER_OBJECT_FILE]-name" = "Perform Single-Object Prelink";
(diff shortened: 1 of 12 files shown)


Sources/SWBCore/Specs/en.lproj/CoreBuildSystem.strings
```diff
@@ -409,8 +409,8 @@ Generally you should not specify an order file in Debug or Development configura
...
 
-"[GENERATE_MASTER_OBJECT_FILE]-name" = "Perform Single-Object Prelink";
-"[GENERATE_MASTER_OBJECT_FILE]-description" = "Activating this setting will cause the object files built by a target to be prelinked using `ld -r` into a single object file, and that object file will then be linked into the final product. This is useful to force the linker to resolve symbols and link the object files into a single module before building a static library. Also, a separate set of link flags can be applied to the prelink allowing additional control over, for instance, exported symbols.";
+"[GENERATE_PRELINK_OBJECT_FILE]-name" = "Perform Single-Object Prelink";
+"[GENERATE_PRELINK_OBJECT_FILE]-description" = "Activating this setting will cause the object files built by a target to be prelinked using `ld -r` into a single object file, and that object file will then be linked into the final product. This is useful to force the linker to resolve symbols and link the object files into a single module before building a static library. Also, a separate set of link flags can be applied to the prelink allowing additional control over, for instance, exported symbols.";
```
jakepetroules, line 413: "PRELINK" here rather than "PRELINKED" since there are two other build settings used for this spec which are already called "PRELINK" (see below)

owenv (approved):

PR #455, opened 2025-04-29
[Explicit Module Builds] Add support for creating a reproducer when clang process crashes.

vsapsai:
For the explicitly built modules generated script contains commands how to build all necessary .pcm files and how to compile the final translation unit.

rdar://59743925

### DIFF
--- a/Sources/SWBCSupport/CLibclang.cpp
+++ b/Sources/SWBCSupport/CLibclang.cpp
@@ -2183,7 +2183,7 @@ extern "C" {
             argc, const_cast<const char**>(argv), workingDirectory,
             /*ReproducerLocation=*/NULL, /*UseUniqueReproducerName=*/true, &messageString);
         if (message) {
-            *message = strdup(lib->fns.clang_getCString(messageString));
+            *message = strdup_safe(lib->fns.clang_getCString(messageString));
         }
         lib->fns.clang_disposeString(messageString);
         return result == LibclangFunctions::CXError_Success;
--- a/Sources/SWBCSupport/CLibclang.cpp
+++ b/Sources/SWBCSupport/CLibclang.cpp
@@ -859,6 +859,9 @@ extern "C" {
         typedef struct CXOpaqueDependencyScannerServiceOptions
             *CXDependencyScannerServiceOptions;
 
+        /**
+         * Generate a self-contained reproducer in a specified location to re-run the compilation.
+         */
         enum CXErrorCode
         (*clang_experimental_DependencyScanner_generateReproducer)(
             int argc, const char *const *argv, const char *WorkingDirectory,
--- a/Sources/SWBCSupport/CLibclang.cpp
+++ b/Sources/SWBCSupport/CLibclang.cpp
@@ -859,6 +859,12 @@ extern "C" {
         typedef struct CXOpaqueDependencyScannerServiceOptions
             *CXDependencyScannerServiceOptions;
 
+        enum CXErrorCode
+        (*clang_experimental_DependencyScanner_generateReproducer)(
+            int argc, const char *const *argv, const char *WorkingDirectory,
+            const char *ReproducerLocation, bool UseUniqueReproducerName,
+            CXString *messageOut);
+
         /**
          * Creates a default set of service options.
          * Must be disposed with \c
@@ -1423,6 +1429,7 @@ struct LibclangWrapper {
         LOOKUP_OPTIONAL(clang_experimental_cas_replayCompilation);
         LOOKUP_OPTIONAL(clang_experimental_cas_ReplayResult_dispose);
         LOOKUP_OPTIONAL(clang_experimental_cas_ReplayResult_getStderr);
+        LOOKUP_OPTIONAL(clang_experimental_DependencyScanner_generateReproducer);
         LOOKUP_OPTIONAL(clang_experimental_DependencyScannerServiceOptions_create);
         LOOKUP_OPTIONAL(clang_experimental_DependencyScannerServiceOptions_dispose);
         LOOKUP_OPTIONAL(clang_experimental_DependencyScannerServiceOptions_setDependencyMode);
@@ -1773,6 +1780,10 @@ extern "C" {
     bool libclang_has_current_working_directory_optimization(libclang_t lib) {
         return lib->wrapper->fns.clang_experimental_DepGraphModule_isCWDIgnored &&
                lib->wrapper->fns.clang_experimental_DependencyScannerServiceOptions_setCWDOptimization;
+    }
+
+    bool libclang_has_reproducer_feature(libclang_t lib) {
+        return lib->wrapper->fns.clang_experimental_DependencyScanner_generateReproducer;
     }
 
     libclang_casoptions_t libclang_casoptions_create(libclang_t lib) {
@@ -2157,6 +2168,22 @@ extern "C" {
         scanner->scanner->releaseWorker(std::move(worker));
 
         return depGraph != nullptr;
+    }
+
+    bool libclang_scanner_generate_reproducer(libclang_scanner_t scanner,
+                                              int argc, char *const *argv,
+                                              const char *workingDirectory,
+                                              const char **message) {
+        auto lib = scanner->scanner->lib;
+        LibclangFunctions::CXString messageString;
+        auto result = lib->fns.clang_experimental_DependencyScanner_generateReproducer(
+            argc, const_cast<const char**>(argv), workingDirectory,
+            /*ReproducerLocation=*/NULL, /*UseUniqueReproducerName=*/true, &messageString);
+        if (message) {
+            *message = strdup(lib->fns.clang_getCString(messageString));
+        }
+        lib->fns.clang_disposeString(messageString);
+        return result == LibclangFunctions::CXError_Success;
     }
 
     bool libclang_driver_get_actions(libclang_t wrapped_lib,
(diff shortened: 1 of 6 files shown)


vsapsai:
This is an initial version that is lacking comments and tests. At first I want to get the general feedback if such approach is reasonable or if it won't work on some basic level. If you have better ideas to implement it, that would be great too.

Sources/SWBCSupport/CLibclang.cpp
```diff
@@ -2159,6 +2170,22 @@ extern "C" {
...
+                                              int argc, char *const *argv,
+                                              const char *workingDirectory,
+                                              const char **message) {
+        auto lib = scanner->scanner->lib;
+        LibclangFunctions::CXString messageString;
+        auto result = lib->fns.clang_experimental_DependencyScanner_generateReproducer(
+            argc, const_cast<const char**>(argv), workingDirectory,
+            /*ReproducerLocation=*/NULL, /*UseUniqueReproducerName=*/true, &messageString);
+        if (message) {
+            *message = strdup(lib->fns.clang_getCString(messageString));
```
owenv, line 2183: Is libclang guaranteed to return a non null string in messageString here?

Sources/SWBTaskExecution/DynamicTaskSpecs/ClangModuleDependencyGraph.swift
```diff
@@ -153,33 +158,36 @@ package final class ClangModuleDependencyGraph {
...
         package func serialize<T>(to serializer: T) where T : Serializer {
-            serializer.serializeAggregate(9) {
+            serializer.serializeAggregate(10) {
                 serializer.serialize(kind)
                 serializer.serialize(files)
                 serializer.serialize(includeTreeID)
                 serializer.serialize(modules)
                 serializer.serialize(workingDirectory)
                 serializer.serialize(commands)
+                serializer.serialize(scanningCommandLine)
```
owenv, line 175: We should try to measure the impact of serializing an extra command line here, it should be shorter than the cc1 command but we do want to keep this data small...

Sources/SWBTaskExecution/TaskActions/ClangCompileTaskAction.swift
```diff
@@ -301,6 +301,19 @@ public final class ClangCompileTaskAction: TaskAction, BuildValueValidatingTaskA
                         outputDelegate.emitOutput("Failed frontend command:\n")
                         outputDelegate.emitOutput(ByteString(encodingAsUTF8: commandString) + "\n")
                     }
+
```
owenv, line 304: Related to the question about serialization overhead, we should consider whether this should be opt-in or opt-out. I'm not sure I have a strong opinion right now

owenv:
this approach lgtm, just had a few minor comments

Sources/SWBCSupport/CLibclang.h
```diff
@@ -126,6 +126,8 @@ bool libclang_has_cas_up_to_date_checks_feature(libclang_t lib);
 /// Whether the libclang has current working directory optimization support.
 bool libclang_has_current_working_directory_optimization(libclang_t lib);
 
+bool libclang_has_reproducer_feature(libclang_t lib);
```
owenv, line 129: Please rebase and tag this with CSUPPORT_EXPORT, and do the same for libclang_scanner_generate_reproducer so it gets the __declspec(dllexport) modifier on windows

Sources/SWBTaskExecution/TaskActions/ClangCompileTaskAction.swift
```diff
@@ -301,6 +301,19 @@ public final class ClangCompileTaskAction: TaskAction, BuildValueValidatingTaskA
...
+                    if case .some(.exit(.uncaughtSignal, _)) = outputDelegate.result {
```
jakepetroules, line 305: We treat some signals as cancellation; should we avoid creating reproducers in that case?

Sources/SWBCSupport/CLibclang.cpp
vsapsai: For success case I think we can make it mandatory for messageString to be present (you need to tell where to find the generated reproducer). Error cases might not populate messageString. Even though they should provide more details, it is fairly easy to return InvalidArgument error without any extra message.

But we can guarantee messageString isn't corrupted and in the worst case it would be an empty string.

I'll think about these cases and would document API contract on libclang side.

Sources/SWBTaskExecution/DynamicTaskSpecs/ClangModuleDependencyGraph.swift
vsapsai: Is the impact in the consumed storage space? Just to know what to measure.

Sources/SWBTaskExecution/TaskActions/ClangCompileTaskAction.swift
vsapsai: We also have plans to add reproducer-on-error. That one should be opt-in. But crashes seem to be rare enough that generating a reproducer can be don unconditionally. Honestly, the crashes seem to be rare enough that this entire feature in its current form isn't particularly useful. The value should come from iterating on the feature.

Sources/SWBTaskExecution/TaskActions/ClangCompileTaskAction.swift
vsapsai: If we have a cancellation, does `TaskProcessDelegate.commandResult == .cancelled`? We can generate a reproducer only when `lastResult == .failed` to avoid useless reproducers. I think in my testing I've encountered signals to some compilations after the build encountered crashes for a different file.

Sources/SWBTaskExecution/DynamicTaskSpecs/ClangModuleDependencyGraph.swift
vsapsai: I've tested the storage space changes on [Sparkle](https://github.com/sparkle-project/Sparkle) and here are my findings. The build directory has grown by 2.3MB (2,441,200 bytes) which is 0.5% of the build directory size (growth from 450,234,890 bytes to 452,676,090). Most of the growth (1.8MB) is in ExplicitPrecompiledModules where each .scan file (out of 647) has grown by 2.5-3KB. For the record, the smallest .pcm in my testing is 20KB, average is 366KB. I wasn't using short names or shallow directory structure, so the measurements should be fairly representative.

Is it a sufficient measurement or would you like to see more data?

vsapsai:
Are there tests that actually execute the build? I've found testing the build description is correct. But the reproducers aren't a part of the build description, we need to actually call the compiler to see if it crashes.

So should I write test for this change and if yes, can you give some examples of existing test cases, please?

Sources/SWBTaskExecution/TaskActions/PrecompileClangModuleTaskAction.swift
```diff
@@ -215,16 +215,30 @@ final public class PrecompileClangModuleTaskAction: TaskAction, BuildValueValida
...
+                    ).encode(commandLine)
+
+                    // <rdar://59354519> We need to find a way to use the generic infrastructure for displaying the command line in
+                    // the build log.
+                    outputDelegate.emitOutput("Failed frontend command:\n")
+                    outputDelegate.emitOutput(ByteString(encodingAsUTF8: commandString) + "\n")
+                }
+                if case .some(.exit(.uncaughtSignal, _)) = outputDelegate.result {
+                    do {
+                        if let reproducerMessage = try clangModuleDependencyGraph.generateReproducer(
```
cachemeifyoucan, line 232: Do you have a PR for how `generateReproducer` is implemented? I don't like a in-process generator reproducer because it is likely to crash again and take down the entire build system, e.g. it is a crash in preprocessor.

cachemeifyoucan (commented):
What I like to see if there is an official way to support crash reproducer for all tools in build system, where each tool can just implement a specification that if it crashes, it can create a new command to construct a reproducer into a directory specified by the build system.

Sources/SWBTaskExecution/TaskActions/PrecompileClangModuleTaskAction.swift
vsapsai: libclang part is https://github.com/swiftlang/llvm-project/pull/10577 We detect crashes in cc1 but when we generate a reproducer, we don't run those commands. So it is not likely to crash again. In fact, generating a reproducer mostly does dependency scanning that we've done already. It is possible that a successful action can crash when repeated again but I don't think I would describe it as "likely".

vsapsai:
> What I like to see if there is an official way to support crash reproducer for all tools in build system, where each tool can just implement a specification that if it crashes, it can create a new command to construct a reproducer into a directory specified by the build system.

How long do you think it would take to add reproducers for all other tools? If it will be done soon, I don't see a problem with adjusting reproducers for EBM to fit the general approach. Going from "no reproducers" to "reproducers for everything" is commendable and I admire your dedication. My idea was that implementing a reproducer for 1 case would help us to see what is useful and what is not, so we can base future work on actual usage experience and not on the guesswork. That is also a reason to limit the scope of changes.

Sources/SWBTaskExecution/TaskActions/PrecompileClangModuleTaskAction.swift
cachemeifyoucan: I see. This reproducer has no preprocessing steps more than the dependency scanning which is probably fine even I might still prefer an out of process tool for that, but that will need a bigger change to squeeze multiple tool execution into one task.

owenv (approved):

vsapsai:
@swift-ci please test

PR #474, opened 2025-04-30
Respect entitlements file above build setting for build settings affe…

baujla:
This makes it so the sandbox and hardened runtime build setting will never overwrite a setting present in the entitlements file. This is a bad state for a project to be, hence why we've added warnings for this case exactly. 

The entitlements file is the authority on what entitlements should be, while the build setting here is a convenience. In this case though, the build system is acting as an authority which based on feedback is quite confusing.


baujla:
@swift-ci please test

baujla:
We will not be making this change, but rather we will stop implicitly setting the value for ENABLE_APP_SANDBOX In tests and will soften the warning.

PR #475, opened 2025-05-01
Fixes inconsistent locations found for top level app sandbox conflict…

baujla:
The sub issue locations for this warning need to be consistent otherwise the top level warning takes the developer to inconsistent places. This feels odd as the warnings seem pretty similar.


baujla:
@swift-ci please test

baujla:
No longer relevant as we removed the second warning.

PR #476, opened 2025-05-01
Add a helper function for avoiding certain race conditions in Win32 API calls

jakepetroules:
For many Win32 APIs returning strings, the expected usage is to first call the API with a nil buffer and 0 capacity to receive the buffer count, then call it again with a buffer of the right size. However, some APIs refer to external state which can change between calls (such as GetEnvironmentVariableW and GetCurrentDirectoryW). This can lead to race conditions where the buffer doesn't end up being of sufficient size to hold the result, and the call fails. To protect against this, add a helper function with a reusable algorithm that continually doubles the buffer size until it's large enough to hold the result, up to a specified maximum to prevent denial of service attacks.

### DIFF
--- a/Sources/SWBUtil/CMakeLists.txt
+++ b/Sources/SWBUtil/CMakeLists.txt
@@ -101,6 +101,7 @@ add_library(SWBUtil
   VFS.swift
   WaitCondition.swift
   WeakRef.swift
+  Win32.swift
   Win32Error.swift
   XCBuildDataArchive.swift
   Xcode.swift)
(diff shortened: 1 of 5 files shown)


jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test macos

PR #478, opened 2025-05-02
Reverts change for test targets and softens warning

baujla:
We need to revert the change we made for test targets and soften the warning as it's valid to have entitlements set to true while the sandbox build setting is set to NO implicitly.


baujla:
@swift-ci please test

baujla:
@swift-ci please test

jakepetroules (dismissed):

jakepetroules (approved):

PR #480, opened 2025-05-02
Setup a basic run of cmake-smoke-test in CI

owenv:
Add test coverage of the Cmake build

### DIFF
--- a/Plugins/cmake-smoke-test/cmake-smoke-test.swift
+++ b/Plugins/cmake-smoke-test/cmake-smoke-test.swift
@@ -24,21 +24,24 @@ struct CMakeSmokeTest: CommandPlugin {
         }
 
         guard let cmakePath = args.extractOption(named: "cmake-path").last else { throw Errors.missingRequiredOption("--cmake-path") }
-        print("using cmake at \(cmakePath)")
+        Diagnostics.progress("using cmake at \(cmakePath)")
         let cmakeURL = URL(filePath: cmakePath)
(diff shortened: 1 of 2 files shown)


Plugins/cmake-smoke-test/cmake-smoke-test.swift
```diff
@@ -166,29 +168,18 @@ enum OS {
 }
 
 extension Process {
-    func run() async throws {
```
jakepetroules, line 169: why?

owenv: Process execution is hanging the CI action, haven't fully debugged it yet

owenv:
@swift-ci test

jakepetroules (approved):

Plugins/cmake-smoke-test/cmake-smoke-test.swift
```diff
@@ -182,7 +186,53 @@ extension Process {
...
+
+        posix_spawnattr_init(&attrs)
+        posix_spawnattr_setpgroup(&attrs, 0)
+        var noSignals = sigset_t()
+        sigemptyset(&noSignals)
+        posix_spawnattr_setsigmask(&attrs, &noSignals)
+
+        var mostSignals = sigset_t()
+        sigemptyset(&mostSignals)
+        for i in 1 ..< SIGSYS {
```
jakepetroules, line 210: `..<SIGRTMAX`?

owenv: I don't think anything is using the realtime signals, and there could be a gap between SIGSYS and SIGRTMIN on some implementations

PR #483, opened 2025-05-03
Make SpecRegistry Sendable and use Swift Concurrency

jakepetroules:
TSan found some races in the spec proxy registration code, switch that to Swift async to get better compile time assurance it's free of data races.


jakepetroules:
@swift-ci test

owenv (approved):

PR #485, opened 2025-05-03
Implement Statistic with atomics so it's thread safe

jakepetroules:
Prevents false positives in TSAN which obscure other real issues.


jakepetroules:
@swift-ci test

owenv (approved):

PR #488, opened 2025-05-05
Include xcspec resources in the CMake build

owenv:
Builds on a rebased https://github.com/swiftlang/swift-build/pull/243 by @compnerd 

- Add SwiftBuild_Bundle to handle assembly of the resource bundles
- Move resource bundle access generation inside so it can be shared among all the targets

### DIFF
--- a/Sources/CMakeLists.txt
+++ b/Sources/CMakeLists.txt
@@ -16,11 +16,38 @@ function(SwiftBuild_Bundle)
   cmake_parse_arguments(PARSE_ARGV 0 BundleXCSpecs
     "${Options}" "${OneValueArguments}" "${MultiValueArguments}")
 
+  list(TRANSFORM BundleXCSpecs_FILES PREPEND "${CMAKE_CURRENT_SOURCE_DIR}/")
   add_custom_command(TARGET ${BundleXCSpecs_MODULE} POST_BUILD
     COMMAND
-      ${CMAKE_COMMAND} -E make_directory "$<TARGET_PROPERTY:${BundleXCSpecs_MODULE},LOCATION>/SwiftBuild_${BundleXCSpecs_MODULE}.resources"
+      ${CMAKE_COMMAND} -E make_directory "${CMAKE_BINARY_DIR}/share/pm/SwiftBuild_${BundleXCSpecs_MODULE}.resources"
     COMMAND
-      ${CMAKE_COMMAND} -E copy_if_different ${BundleXCSpecs_FILES} "$<TARGET_PROPERTY:${BundleXCSpecs_MODULE},LOCATION>/SwiftBuild_${BundleXCSpecs_MODULE}.resources/")
+      ${CMAKE_COMMAND} -E copy_if_different ${BundleXCSpecs_FILES} "${CMAKE_BINARY_DIR}/share/pm/SwiftBuild_${BundleXCSpecs_MODULE}.resources/")
+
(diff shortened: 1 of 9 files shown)


neonichu (approved):

Sources/CMakeLists.txt
```diff
@@ -8,6 +8,47 @@ See http://swift.org/LICENSE.txt for license information
...
+
+  list(TRANSFORM BundleXCSpecs_FILES PREPEND "${CMAKE_CURRENT_SOURCE_DIR}/")
+  add_custom_command(TARGET ${BundleXCSpecs_MODULE} POST_BUILD
+    COMMAND
+      ${CMAKE_COMMAND} -E make_directory "${CMAKE_CURRENT_BINARY_DIR}/SwiftBuild_${BundleXCSpecs_MODULE}.resources"
+    COMMAND
+      ${CMAKE_COMMAND} -E copy_if_different ${BundleXCSpecs_FILES} "${CMAKE_CURRENT_BINARY_DIR}/SwiftBuild_${BundleXCSpecs_MODULE}.resources/")
+
+  file(CONFIGURE
+    OUTPUT resource_bundle_accessor.swift
```
compnerd, line 27: I think that the absolute path with the conversion is important. Otherwise, the output location is not very well defined.

jakepetroules (approved):

Sources/CMakeLists.txt
compnerd: Yes, it is documented to be relative to that, and I originally tried to use the relative path but ended up running into some issues. Furthermore, now that it is extracted into a helper, you no longer have a good idea of where `CMAKE_CURRENT_BINARY_DIR` is pointing to.

owenv: ok, that makes sense to me, thanks. Updated the patch to use an absolute path here and avoid use of CMAKE_CURRENT_BINARY_DIR in general

owenv:
@swift-ci test

Sources/CMakeLists.txt
```diff
@@ -8,6 +8,48 @@ See http://swift.org/LICENSE.txt for license information
...
+    "${Options}" "${OneValueArguments}" "${MultiValueArguments}")
+
+  list(TRANSFORM BundleXCSpecs_FILES PREPEND "${CMAKE_CURRENT_SOURCE_DIR}/")
+  add_custom_command(TARGET ${BundleXCSpecs_MODULE} POST_BUILD
+    COMMAND
+      ${CMAKE_COMMAND} -E make_directory "${CMAKE_BINARY_DIR}/share/pm/SwiftBuild_${BundleXCSpecs_MODULE}.resources"
+    COMMAND
+      ${CMAKE_COMMAND} -E copy_if_different ${BundleXCSpecs_FILES} "${CMAKE_BINARY_DIR}/share/pm/SwiftBuild_${BundleXCSpecs_MODULE}.resources/")
+
+  file(TO_NATIVE_PATH "${CMAKE_BINARY_DIR}/share/pm/SwiftBuild_${BundleXCSpecs_MODULE}.resources" _SWIFT_BUILD_RESOURCE_BUNDLE_BUILD_PATH)
```
compnerd, line 26: I wonder if we can use `$<TARGET_PROEPRTY:${BundleXCSpecs_MODULE},BINARY_DIR>` to get the module's binary dir rather than placing this at the top-level.

owenv: gave this a try but I was unable to use a generator expression in the output path when creating the resource accessor

owenv:
Going to look at landing this as-is and look for simplification opportunities as follow-ups

@swift-ci test

PR #490, opened 2025-05-06
Remove blocking_sync usages from TaskProducerContext

jakepetroules:
These just need simple synchronization; use a mutex instead.

This also plugs a couple occurrences of unsynchronized access.


jakepetroules:
@swift-ci test

PR #491, opened 2025-05-06
Remove more blocking_sync calls

jakepetroules:
This just removes a bunch of calls that can trivially become async.


jakepetroules:
@swift-ci test

neonichu (approved):

mirza-garibovic:
Nice! But I have to nit that `sync` is a funny name for this.

jakepetroules:
The idea was that you type the "bad" thing (sync) and you automatically get the "good" thing 🤷. I have no strong feelings about the name though.

All these bridging utilities were intended to be temporary until we can entirely eliminate Dispatch usage in favor of Swift concurrency primitives.

PR #494, opened 2025-05-07
[Swift C++ Interop] Propagate hardening build setting to Swift

susmonteiro:
Propagates the hardening mode set in Xcode build settings to the Swift compiler. 

These changes were previously added in and got removed because of rdar://132491728.

I've tested this on the project where these changes used to cause a build error and the project now seems to build successfully with these changes.

Resolves rdar://148029835


susmonteiro:
@swift-ci please test

PR #499, opened 2025-05-09
XCTest discovery support for non-Darwin platforms

owenv:
Add a new target type for SwiftPM test runner executables used on Linux/Windows. XCTest discovery generates an entry point based on index store data. The index store implementation is a lightly edited version of what's in TSC. This will require some adoption work to generate the new test runner target in PIF - we should think about the best way to minimize the diff between Darwin and non-Darwin here (fyi @cmcgee1024 )

### DIFF
--- a/Sources/SWBCore/SpecImplementations/ProductTypes.swift
+++ b/Sources/SWBCore/SpecImplementations/ProductTypes.swift
@@ -321,7 +321,7 @@ public class ProductTypeSpec : Spec, SpecType, @unchecked Sendable {
     }
 
     /// Returns whether the product type supports embedding Swift standard libraries inside it.
-    public var supportsEmbeddingSwiftStandardLibraries: Bool {
+    public func supportsEmbeddingSwiftStandardLibraries(producer: CommandProducer) -> Bool {
         // Most product types don't support having the Swift libraries embedded in them.
         return false
     }
@@ -381,7 +381,7 @@ public final class ApplicationProductTypeSpec : BundleProductTypeSpec, @unchecke
         return "PBXApplicationProductType"
     }
 
-    public override var supportsEmbeddingSwiftStandardLibraries: Bool {
+    public override func supportsEmbeddingSwiftStandardLibraries(producer: CommandProducer) -> Bool {
         return true
     }
 
@@ -602,8 +602,8 @@ public final class XCTestBundleProductTypeSpec : BundleProductTypeSpec, @uncheck
         super.init(parser, basedOnSpec)
     }
 
-    public override var supportsEmbeddingSwiftStandardLibraries: Bool {
-        return true
+    public override func supportsEmbeddingSwiftStandardLibraries(producer: CommandProducer) -> Bool {
+        return producer.isApplePlatform
     }
 
     public class func usesXCTRunner(_ scope: MacroEvaluationScope) -> Bool {
@@ -649,7 +649,7 @@ public final class XCTestBundleProductTypeSpec : BundleProductTypeSpec, @uncheck
         var (tableOpt, warnings, errors) = super.overridingBuildSettings(scope, platform: platform)
         var table = tableOpt ?? MacroValueAssignmentTable(namespace: scope.namespace)
 
-        let isDeviceBuild = platform?.isDeploymentPlatform == true && platform?.identifier != "com.apple.platform.macosx"
+        let isDeviceBuild = platform?.isDeploymentPlatform == true && platform?.name != scope.evaluate(BuiltinMacros.HOST_PLATFORM)
         if isDeviceBuild {
             // For tests running on devices (not simulators) we always want to generate dSYMs so that symbolication can give file and line information about test failures.
             table.push(BuiltinMacros.DEBUG_INFORMATION_FORMAT, literal: "dwarf-with-dsym")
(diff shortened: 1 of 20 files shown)


owenv:
@swift-ci test

Sources/SWBCore/SpecImplementations/ProductTypes.swift
```diff
@@ -649,7 +649,7 @@ public final class XCTestBundleProductTypeSpec : BundleProductTypeSpec, @uncheck
         var (tableOpt, warnings, errors) = super.overridingBuildSettings(scope, platform: platform)
         var table = tableOpt ?? MacroValueAssignmentTable(namespace: scope.namespace)
 
-        let isDeviceBuild = platform?.isDeploymentPlatform == true && platform?.identifier != "com.apple.platform.macosx"
+        let isDeviceBuild = platform?.isDeploymentPlatform == true && platform?.name != scope.evaluate(BuiltinMacros.HOST_PLATFORM)
```
jakepetroules, line 652: As an aside, this seems fundamentally flawed. There's no guarantee you're running the resulting binaries on the same host that built them, so it seems the host platform should enable this too.

Why simulators are excluded I'm not sure.

Sources/SWBProjectModel/PIFGenerationModel.swift
```diff
@@ -295,6 +295,7 @@ public enum PIF {
             case executable = "com.apple.product-type.tool"
             case hostBuildTool = "com.apple.product-type.tool.host-build"
             case unitTest = "com.apple.product-type.bundle.unit-test"
+            case swiftpmTestRunner = "com.apple.product-type.tool.swiftpm-test-runner"
```
jakepetroules, line 298: Why does this need to be called "SwiftPM" test runner?

Sources/SWBUniversalPlatform/TestEntryPointGenerationTaskAction.swift
```diff
@@ -23,10 +23,37 @@ class TestEntryPointGenerationTaskAction: TaskAction {
...
+                let testInfo = try store.listTests(in: objects)
+                tests.append(contentsOf: testInfo)
+            }
+
+            try executionDelegate.fs.write(options.output, contents: ByteString(encodingAsUTF8: """
             #if canImport(Testing)
             import Testing
             #endif
+            
+            \(testObservationFragment(testOutputPath: "TODO"))
```
jakepetroules, line 53: TODO still supposed to be here?

Sources/SWBUniversalPlatform/TestEntryPointGenerationTaskAction.swift
```diff
@@ -66,18 +103,562 @@ class TestEntryPointGenerationTaskAction: TaskAction {
...
+        fragment += """
+            ]
+        }
+        """
+        return fragment
+    }
+
+    private func testObservationFragment(testOutputPath: String) -> String {
+        """
+        #if !os(Windows) // Test observation is not supported on Windows
```
jakepetroules, line 165: Why? Do we have a GitHub issue for this?

Sources/SWBUniversalPlatform/TestEntryPointGenerationTool.swift
```diff
@@ -17,7 +17,59 @@ import SWBCore
...
+            }
+        }
+        return args
+    }
+
     override func createTaskAction(_ cbc: CommandBuildContext, _ delegate: any TaskGenerationDelegate) -> (any PlannedTaskAction)? {
         TestEntryPointGenerationTaskAction()
     }
+
+    public func constructTasks(_ cbc: CommandBuildContext, _ delegate: any TaskGenerationDelegate, indexStorePaths: [Path], indexUnitBasePaths: [Path]) async {
```
jakepetroules, line 43: Can we inject the necessary additional arguments without overriding constructTasks? Ever bare call to createTasks adds risk that we forget to update things as its signature or default behavior changes.

Sources/SWBProjectModel/PIFGenerationModel.swift
owenv: I called it that to differentiate it from other kinds of test runners like XCTRunner, but I'm not too attached to the name

Sources/SWBUniversalPlatform/TestEntryPointGenerationTaskAction.swift
owenv: Nope, fixed now so that this is provided via a CLI arg

Sources/SWBUniversalPlatform/TestEntryPointGenerationTaskAction.swift
owenv: This is pulled in verbatim from SwiftPM

Sources/SWBUniversalPlatform/TestEntryPointGenerationTool.swift
owenv: I tried to do this at first, but I couldn't find an elegant way to differentiate the input types, especially since the handling of the index store is a bit unusual since the build system doesn't track the content inside it

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test

PR #502, opened 2025-05-10
Rename specs to workaround windows whitespace-escaping issues

owenv:
Try to fix an issue on Windows when the generator collapses the post-build commands for SWBApplePlatform into a batch script


owenv:
@swift-ci test

neonichu (approved):

jakepetroules:
Meh, good riddance to spaces in filenames.

PR #505, opened 2025-05-13
[CAS][SwiftCompiler] We don't need to disable caching if there is a bridging header that is not precompiled

akyrtzi:
With later improvements we actually force precompiling the bridging header from the driver, when caching is enabled.


akyrtzi:
@swift-ci Please test

owenv (approved):

cachemeifyoucan (approved):

neonichu (approved):

PR #507, opened 2025-05-13
Allow task spec to filter command line for signature

ldaley:
This wires TaskTypeDescription#commandLineForSignature() into the default implementation of TaskAction#getSignature(), allowing task action implementations to get correct signature normalization by default.


neonichu (approved):

jakepetroules:
Nice catch, this was a bug waiting to happen!

PR #514, opened 2025-05-19
Fix quadratic behavior when GenericOutputParser is written to many times per line

owenv:
This was observed in a project which had very long linker response files and passed OTHER_LDFLAGS=-v, causing clang to emit them as output.

rdar://151481591


owenv:
@swift-ci test

jakepetroules (approved):

neonichu (approved):

b3ll:
Thank you for fixing this!! :D

PR #515, opened 2025-05-19
Read hardcoded deployment targets list from features.json instead

jakepetroules:
Eliminate the maintenance burden of having to manually maintain this list. Also, it wasn't correct for Android and QNX anyways, since those don't actually have deployment target environment variables recognized by clang.

rdar://91377944


jakepetroules:
@swift-ci test

owenv (approved):

jakepetroules:
@swift-ci test

PR #517, opened 2025-05-20
Add support for building Swift on Android targets

jakepetroules:
Swift Build currently has support for building C/C++ code on Android; this extends it to Swift.

Concretely, this fixes the position of the version number in the triple passed to swiftc for Android targets, and adds the relevant search paths to find Swift standard libraries and other content in the Swift SDK.

The sysroot still comes from an Android NDK found on the system; the sysroot in the Swift SDK is ignored based on the intended direction around supporting Android targets in Swift.


neonichu (approved):

jakepetroules:
@swift-ci test

PR #520, opened 2025-05-22
Ensure swift jobs with the same command line but different dependenci…

owenv:
…es aren't deduplicated

This can happen if the Swift Driver plans the same job in two different builds, but specifies a differtent set of input dependencies which either remove up to date jobs or add out of date ones. Without this change, this could rarely lead to looking up the wrong set of or nonexistent input jobs in an incremental build following a failed or cancelled one

Fix this by incorporating the job dependencies in the key which uniquely identifies a job's dynamic task

### DIFF
--- a/Sources/SWBCore/LibSwiftDriver/PlannedBuild.swift
+++ b/Sources/SWBCore/LibSwiftDriver/PlannedBuild.swift
@@ -80,8 +80,6 @@ public struct SwiftDriverJob: Serializable, CustomDebugStringConvertible {
     public let outputs: [Path]
     /// The command line to execute for this job
     public let commandLine: [SWBUtil.ByteString]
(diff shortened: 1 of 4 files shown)


owenv:
@swift-ci test

jakepetroules (approved):

cachemeifyoucan (approved):
LGTM

Sources/SWBCore/LibSwiftDriver/PlannedBuild.swift
```diff
@@ -196,6 +197,7 @@ extension LibSwiftDriver {
                 try driverJob = deserializer.deserialize()
```
jakepetroules, line 197: @owenv This missed changing 4 to 5 in deserialization

PR #521, opened 2025-05-22
Document setting for allowing use of explicit modules with c++ interop

owenv:
SWIFT_EXPLICIT_MODULES_ALLOW_CXX_INTEROP = YES can be used to opt in to EBM if cxx interop is also enabled

rdar://151806916


owenv:
@swift-ci test

owenv:
@swift-ci test

jakepetroules (approved):

PR #525, opened 2025-05-26
Implement __SKIP_BUILD setting to prune test runners from builds for apple platforms

owenv:
Broke this off of the test runner PR - it will be set on test runner targets on macOS to skip building them when they're not needed. We remove targets the same way dependency scopes are handled, preserving transitive dependencies through a removed target.


owenv:
@swift-ci test

PR #533, opened 2025-05-28
Update automerger PR workflow

owenv:
We'd like to stop including all changes from main in the release/6.2 branch. Flip the direction of the automerger so we can periodically review that all changes to 6.2 do make it to main


owenv:
@swift-ci test

neonichu (approved):

PR #535, opened 2025-05-29
Simplify fallback developer directory calculation

jakepetroules:
Have the fallback logic specify the type of developer directory as well as just the location, and eliminate the initialization distinction on whether a Core was initialized with a "fallback" path vs one explicitly given, since it shouldn't matter.


jakepetroules:
@swift-ci test windows

neonichu (approved):

PR #540, opened 2025-05-30
Correct reference to wrong OS in XCFramework tests

jakepetroules:
Noticed this while working on some other tests; probably doesn't matter but may be confusing so correct it.


jakepetroules:
@swift-ci test

owenv (approved):

PR #548, opened 2025-06-03
Improve printed descriptions of task backtraces

owenv:
- Clarify frame descriptions based on feedback
- expose public API for rendering a backtrace as text

rdar://152194638


owenv:
@swift-ci test

PR #550, opened 2025-06-03
Fix BuildRequest encoding of the recordBuildBacktraces key

owenv:
This was not being encoded properly because we were relying on the user default to enable the feature


owenv:
@swift-ci test

PR #554, opened 2025-06-04
Ignore `-no_exported_symbols` when the debug dylib is used (rdar://152244838)

jonathanpenn:
The stub executor needs to trampoline to the entry point in the debug dylib. But linking with `-no_exported_symbols` strips it and launching the app fails to find the symbol.

Handles both the `LD_EXPORT_SYMBOLS` build setting and manual flags added to `OTHER_LDFLAGS`.


Tests/SWBTaskConstructionTests/PreviewsTaskConstructionTests.swift
```diff
@@ -1019,9 +1019,9 @@ fileprivate struct PreviewsTaskConstructionTests: CoreBasedTests {
         }
     }
 
-    /// Test that any `-dyld_env` arguments end up in the stub executor instead of the preview dylib when is enabled.
+    /// Test that any `OTHER_LDFLAGS` arguments end up in the stub executor instead of the debug dylib when is enabled.
     @Test(.requireSDKs(.iOS))
-    func xOJITDyldEnv() async throws {
+    func xOJITOtherLDFlags() async throws {
```
jakepetroules: nit: could fix this to say "XOJIT". Some of the capitalization got messed up during the XCTest conversion.

neonichu (approved):

Tests/SWBTaskConstructionTests/PreviewsTaskConstructionTests.swift
jonathanpenn: Yup, and fixed two other tests in the same file to match.

jakepetroules: Thanks :)

jakepetroules:
@swift-ci test

jakepetroules (approved):

jonathanpenn:
Failing linux tests are crashes when running the macro evaluation tests, which are unrelated.

PR #556, opened 2025-06-04
Revert "XCTest discovery support for non-Darwin platforms"

owenv:
Reverts swiftlang/swift-build#499

This caused a failure in SwiftPM CI


owenv:
@swift-ci test

PR #562, opened 2025-06-05
[Preview] Fix preview support for swift caching build

cachemeifyoucan:
Add support for swift caching build when swift compiler supports the new option that allows swift compiler performs an uncached build but loading module dependencies from CAS.

It also does few adjustment to make sure preview build shares the same dependencies with regular build as caching build has a stricter rule for invalidation:

* The VFS overlay used by preview is applied after swift driver invocation and onto the frontend invocation directly. This makes sure the VFS overlay doesn't invalidate all the module dependencies.
* BridgingHeader PCH was disabled as a workaround when swift driver doesn't produce deterministic output path when planning for PCH jobs. Properly fix this issue by requesting the same path when planning both build and don't use this workaround in certain configurations.

This should allow preview to build thunk correctly when swift caching is enabled for the regular build.

rdar://152107465

### DIFF
--- a/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
@@ -337,6 +337,7 @@ public struct SwiftDriverPayload: Serializable, TaskPayload, Encodable {
     public let uniqueID: String
     public let compilerLocation: LibSwiftDriver.CompilerLocation
     public let moduleName: String
+    public let outputPrefix: String
     public let tempDirPath: Path
     public let explicitModulesTempDirPath: Path
     public let variant: String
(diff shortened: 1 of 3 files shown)


cachemeifyoucan:
This uses the flag introduced here: https://github.com/swiftlang/swift/pull/81990

cachemeifyoucan:
@swift-ci please test

owenv (approved):

cachemeifyoucan:
Re-request review after properly fixed the bridging header PCH path mismatch.

@swift-ci please test

Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
```diff
@@ -3397,10 +3401,12 @@ public final class SwiftCompilerSpec : CompilerSpec, SpecIdentifierType, SwiftDi
             removeWithParameter(arg)
         }
 
-        // We need to ignore precompiled headers due to:
+        // For some old version of swift driver, the output path for bridging header pch is not stable,
```
cachemeifyoucan, line 3404: Do we have any compatibility requirement for older swift-driver? If we only need to support recent swift-driver, this workaround can be removed to make preview build faster

cachemeifyoucan:
@swift-ci please test

Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
owenv: Generally speaking I try not to immediately break compatibility with old drivers to minimize revlock when building the package for local development. But supporting a driver that's more than a few weeks old is a non-goal.

owenv:
The latest changes seem reasonable to me

Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
cachemeifyoucan: The change in swift-driver I am talking about is: https://github.com/swiftlang/swift-driver/pull/1832

It has been few months but not in official release yet.

PR #563, opened 2025-06-06
Improve error message when build description attachment saving fails

owenv:
Capture and emit the underlying error

### DIFF
--- a/Sources/SWBTaskExecution/BuildDescriptionManager.swift
+++ b/Sources/SWBTaskExecution/BuildDescriptionManager.swift
@@ -753,7 +753,7 @@ private final class BuildSystemTaskPlanningDelegate: TaskPlanningDelegate {
             try fileSystem.createDirectory(path.dirname, recursive: true)
             try fileSystem.write(path, contents: contents)
         } catch {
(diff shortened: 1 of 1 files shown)


owenv:
@swift-ci test

jakepetroules (approved):

jakepetroules:
I wonder if the failures are due to Windows path length limitations.

Recursive directory creation (and only recursive directory creation) runs afoul of the path length limit because recursive mode uses SHCreateDirectoryExW with the full path and lets Win32 handle the intermediate directory creation, while non-recursive uses plain CreateDirectoryW. The `\\?\` path prefix doesn't help with SHCreateDirectoryExW and long paths just fail anyways.

I documented this in the tests: https://github.com/swiftlang/swift-foundation/pull/1257/files#diff-1a0c3769fb737fce906a053a96566757ffd15da8a63224db73acd87592a563aaR1134

We should probably just switch swift-foundation to doing the recursion manually and always use CreateDirectoryW.

owenv:
I think we understand the underlying issue now but this is still nice to have

PR #564, opened 2025-06-06
Remove references to obsolete '.swiftPlaceholder' module type. 

artemcm:
This is a leftover from a prior design which had Swift Build specify paths to to-be-built modules to the driver. This design turned out to be not viable for multiple reasons and was abandoned a couple of years ago. We are now working towards removing this code.


owenv:
@swift-ci test

PR #565, opened 2025-06-06
Update LocalFS to use FileManager for all operations

daveinglis:
LocalFS was using some posix APIs (stat, mkdir, chmod, chown, etc) with some being used in Windows which mean long filenames would fail in some cases, this cleans that up making all platforms that same

### DIFF
--- a/Sources/SWBLLBuild/LowLevelBuildSystem.swift
+++ b/Sources/SWBLLBuild/LowLevelBuildSystem.swift
@@ -11,11 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 public import SWBUtil
-#if os(Windows)
-private import SWBLibc
-#else
 public import SWBLibc
-#endif
 
 // Re-export all APIs from llbuild bindings.
 @_exported public import llbuild
@@ -25,7 +21,34 @@ public import SWBLibc
 #endif
 
 // Filesystem adaptors for SWBLLBuild.FileSystem.
-extension SWBUtil.FileInfo: SWBLLBuild.FileInfo {}
+extension SWBUtil.FileInfo: SWBLLBuild.FileInfo {
+
+    public init(_ statBuf: stat) {
+        // This should be remove from llbuild FileInfo protocol as it just not needed, would also be nice to remove the stat requirement too.
+        preconditionFailure()
+    }
+    
+    public var statBuf: stat {
+        var statBuf: stat = stat()
+
+        statBuf.st_dev = numericCast(self.deviceID)
+        statBuf.st_ino = numericCast(self.iNode)
+        statBuf.st_mode = numericCast(self.permissions)
+        statBuf.st_size = numericCast(self.size)
+        #if canImport(Darwin)
+        statBuf.st_mtimespec.tv_sec = numericCast(self.modificationTimestamp)
+        statBuf.st_mtimespec.tv_nsec = self.modificationNanoseconds
+        #elseif os(Windows)
+        statBuf.st_mtime = self.modificationTimestamp
+        #elseif canImport(Glibc) || canImport(Musl) || canImport(Android)
+        statBuf.st_mtim.tv_sec = numericCast(self.modificationTimestamp)
+        statBuf.st_mtim.tv_nsec = self.modificationNanoseconds
+        #else
+        #error("Not implemented for this platform")
+        #endif
+        return statBuf
+    }
+}
 
 public final class FileSystemImpl: FileSystem {
 
--- a/Sources/SWBUtil/FSProxy.swift
+++ b/Sources/SWBUtil/FSProxy.swift
@@ -10,7 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-public import SWBLibc
+import SWBLibc
 
 #if canImport(System)
 public import System
@@ -18,6 +18,7 @@ public import System
 public import SystemPackage
 #endif
 
+public import struct Foundation.CocoaError
 public import struct Foundation.Data
 public import struct Foundation.Date
 public import struct Foundation.FileAttributeKey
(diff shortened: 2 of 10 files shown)


Sources/SWBUtil/FSProxy.swift
```diff
@@ -28,82 +29,86 @@ public import struct Foundation.URL
...
+        return _readFileAttributePrimitive(fileAttrs[.systemFileNumber], as: UInt64.self) ?? 0
+    }
+
+    public var deviceID: Int32 {
+        return _readFileAttributePrimitive(fileAttrs[.systemNumber], as: Int32.self) ?? 0
     }
 
     public static func ==(lhs: FileInfo, rhs: FileInfo) -> Bool {
-        return lhs.statBuf == rhs.statBuf
+        return NSDictionary(dictionary: lhs.fileAttrs).isEqual(to: rhs.fileAttrs)
```
daveinglis: I thought dictionaries where Equatable but when I just test them with '==' it fails (`Candidate requires that 'any Sendable' conform to 'Equatable' (requirement specified as 'any Sendable' : 'Equatable') (Swift.Dictionary))` even tried adding Equatable to the type specifier for fileAttr but still similar error.

daveinglis:
@swift-ci test

Sources/SWBUtil/FSProxy.swift
daveinglis: hmm, clearly Im doing something wrong as this doesn't work on linux...test fails but shows the dictionaries are the same....

owenv: I think you're comparing an NSDictionary on the lhs to a native Swift Dictionary on the rhs, which won't be equal even if both have the same contents

daveinglis:
@swift-ci test

Sources/SWBUtil/FSProxy.swift
daveinglis: ok updated, though the `to:` is suppose to take `[Hashable: Any].`..), but it seem to work locally now... I thought you where suppose to just be able to use '==' (https://developer.apple.com/documentation/swift/dictionary/==(_:_:)) ?

Sources/SWBLLBuild/LowLevelBuildSystem.swift
```diff
@@ -25,7 +21,39 @@ public import SWBLibc
...
+        statBuf.st_mode = self.permissions
+        statBuf.st_size = Int32(self.size)
+        statBuf.st_mtime = self.modificationTimestamp
+        #elseif canImport(Glibc) || canImport(Musl) || canImport(Android)
+        statBuf.st_dev = UInt(self.deviceID)
+        statBuf.st_ino = UInt(self.iNode)
+        statBuf.st_mode = UInt32(self.permissions)
+        statBuf.st_size = Int(self.size)
+        statBuf.st_mtim.tv_sec = Int(self.modificationTimestamp)
+        statBuf.st_mtim.tv_nsec = self.modificationNanoseconds
```
jakepetroules, line 45: I'd add:

```swift
#else
#error "Not implemented for this platform
```

...because otherwise new platform bringup will silently return an empty stat structure.

Also, I can test later but I bet you want `|| os(FreeBSD) || os(OpenBSD)` on the `canImport(Darwin)` branch because the BSDs and macOS are probably structured the same.

Sources/SWBUtil/FSProxy.swift
```diff
@@ -28,82 +29,86 @@ public import struct Foundation.URL
...
+        return nil
     }
 
     public var isFile: Bool {
-        #if os(Windows)
-        return (statBuf.st_mode & UInt16(ucrt.S_IFREG)) != 0
-        #else
-        return (statBuf.st_mode & S_IFREG) != 0
-        #endif
+        return (fileAttrs[.type] as! FileAttributeType == .typeRegular)
```
jakepetroules, line 66: Can you express this as `return fileAttrs[.type] == FileAttributeType.typeRegular` to avoid the force cast?

Sources/SWBUtil/FSProxy.swift
```diff
@@ -28,82 +29,86 @@ public import struct Foundation.URL
...
+#if os(macOS)
+import struct ObjectiveC.ObjCBool
 #endif
 
 /// File system information for a particular file.
 ///
 /// This is a simple wrapper for stat() information.
 public struct FileInfo: Equatable, Sendable {
-    public let statBuf: stat
+    public let fileAttrs: [FileAttributeKey: any Sendable]
```
jakepetroules, line 49: Can the value be `any Sendable & any Equatable`? I think that might allow simplifying:

```swift
return NSDictionary(dictionary: lhs.fileAttrs).isEqual(NSDictionary(dictionary: rhs.fileAttrs))
```

to

```swift
return lhs.fileAttrs == rhs.fileAttrs
```

Sources/SWBUtil/FSProxy.swift
```diff
@@ -348,28 +353,30 @@ class LocalFS: FSProxy, @unchecked Sendable {
...
     }
 
     /// Check whether the given path is a directory.
     ///
     /// If the given path is a symlink to a directory, then this will return true if the destination of the symlink is a directory.
     func isDirectory(_ path: Path) -> Bool {
-        var statBuf = stat()
-        if stat(path.str, &statBuf) < 0 {
-            return false
+#if os(macOS)
```
jakepetroules: Unfortunate conditional... I guess this is swiftlang/swift-foundation#909?

However, you want `canImport(Darwin)` here rather than `os(macOS)` because the iOS path will be the same as macOS.

Tests/SWBTaskExecutionTests/FileCopyTaskTests.swift
```diff
@@ -122,7 +122,11 @@ fileprivate struct FileCopyTaskTests {
             #expect(result == .failed)
 
             // Examine the error messages.
-            XCTAssertMatch(outputDelegate.errors, [.suffix("MissingFile.bogus): No such file or directory (2)")])
+            #if os(macOS)
```
jakepetroules: `canImport(Darwin)`

Sources/SWBUtil/FSProxy.swift
```diff
@@ -18,6 +18,7 @@ public import System
 public import SystemPackage
```
jakepetroules, line 18: commenting on wrong line because GitHub... is the SWBLibc import needed anymore?

Sources/SWBUtil/FSProxy.swift
```diff
@@ -28,82 +29,86 @@ public import struct Foundation.URL
...
-
-#if os(Windows)
-// Windows' POSIX layer does not have S_IFLNK, so define it.
-// We only need it for PseudoFS.
-fileprivate let S_IFLNK: Int32 = 0o0120000
+public import struct Foundation.FileAttributeType
+public import struct Foundation.FileAttributeKey
+public import struct Foundation.TimeInterval
+public import class Foundation.NSDictionary
+#if os(macOS)
```
jakepetroules: `canImport(Darwin)`

Sources/SWBUtil/FSProxy.swift
daveinglis: I get `Cannot convert value of type '(any Sendable)?' to expected argument type 'FileAttributeType'`

Sources/SWBUtil/FSProxy.swift
daveinglis: ya, I have tried this, I get `Type 'any Equatable & Sendable' cannot conform to `'Equatable'``. I'm not sure what Im missing...

Sources/SWBUtil/FSProxy.swift
daveinglis: doesn't need to be public anymore, but still required.

Tests/SWBTaskExecutionTests/FileCopyTaskTests.swift
daveinglis: fixed

Sources/SWBUtil/FSProxy.swift
daveinglis: fixed

Sources/SWBLLBuild/LowLevelBuildSystem.swift
daveinglis: fixed

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test linux

daveinglis:
@swift-ci test

owenv (approved):

daveinglis:
@swift-ci test

Sources/SWBUtil/FSProxy.swift
jakepetroules: :(

Sources/SWBLLBuild/LowLevelBuildSystem.swift
```diff
@@ -25,7 +21,41 @@ public import SWBLibc
...
+        #elseif os(Windows)
+        statBuf.st_dev = UInt32(self.deviceID)
+        statBuf.st_ino = UInt16(self.iNode)
+        statBuf.st_mode = self.permissions
+        statBuf.st_size = Int32(self.size)
+        statBuf.st_mtime = self.modificationTimestamp
+        #elseif canImport(Glibc) || canImport(Musl) || canImport(Android)
+        statBuf.st_dev = UInt(self.deviceID)
+        statBuf.st_ino = UInt(self.iNode)
+        statBuf.st_mode = UInt32(self.permissions)
```
jakepetroules: I forgot `canImport(Glibc)` is true on FreeBSD. However, mode_t is UInt16 there.

Maybe use this instead?

```swift
statBuf.st_mode = numericCast(self.permissions)
```

With that change I was able to successfully build for FreeBSD with the latest main + this PR.

daveinglis:
@swift-ci test

Sources/SWBLLBuild/LowLevelBuildSystem.swift
daveinglis: updated

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test linux

daveinglis:
@swift-ci test linux

PR #569, opened 2025-06-09
Support Cancellation in AsyncOperationQueue

plemarquand:
The original `AsyncOperationQueue` was ported in to `swift-package-manager`, where it was updated to support cancellation. This patch backports it to `swift-build` so that the two implementations are the same.


plemarquand:
@swift-ci please test

plemarquand:
@swift-ci please test

PR #571, opened 2025-06-10
Adapt test code to SE-0481

owenv:
These weak captures are now treated as immutable, and I don't see any reason we need to nil them out here


owenv:
@swift-ci test

PR #574, opened 2025-06-10
Ensure the path to Swift Testing's macro plugin is specified correctly when using a non-default Xcode toolchain

stmontgomery:
This resolves an issue which can occur when using Xcode 26 Beta with the downloadable Metal toolchain installed, if a target imports Swift Testing. A toolchain override is specified in this scenario, and this causes the existing logic modified by this PR to incorrectly believe it's using a non-Xcode toolchain, which in turn causes the wrong Swift macro plugin flags to be passed, ultimately leading to a failure locating the `TestingMacros` plugin.

The fix is to recognize all toolchains which have Xcode's prefix, and ensure the default toolchain prefix is used whenever any Xcode toolchain is in use.

Fixes rdar://152440128

### DIFF
--- a/Sources/SWBCore/Settings/Settings.swift
+++ b/Sources/SWBCore/Settings/Settings.swift
@@ -4273,17 +4273,20 @@ private class SettingsBuilder {
         }
 
         let toolchainPath = Path(scope.evaluateAsString(BuiltinMacros.TOOLCHAIN_DIR))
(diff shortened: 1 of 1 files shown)


stmontgomery:
@swift-ci test

jakepetroules (approved):

briancroom (approved):

PR #575, opened 2025-06-11
[6.2] WatchKitTests.watchBackDeploymentArtifacts(): Make the check for the build version in arm64/arm64e slices in the WatchKit stub binary more permissive

mhrawdon:
Also tidy up the logic of these checks since they all are implicitly assuming a single build version per arch, and that's also needed to use the `>=` comparison.

rdar://152870849


mhrawdon:
@swift-ci test

owenv (approved):

jakepetroules (approved):

mhrawdon:
@swift-ci test

PR #580, opened 2025-06-12
[6.2][Preview] Fix preview support for swift caching build (#562)

cachemeifyoucan:
Add support for swift caching build when swift compiler supports the new option that allows swift compiler performs an uncached build but loading module dependencies from CAS.

It also does few adjustment to make sure preview build shares the same dependencies with regular build as caching build has a stricter rule for invalidation:

* The VFS overlay used by preview is applied after swift driver invocation and onto the frontend invocation directly. This makes sure the VFS overlay doesn't invalidate all the module dependencies.
* BridgingHeader PCH was disabled as a workaround when swift driver doesn't produce deterministic output path when planning for PCH jobs. Properly fix this issue by requesting the same path when planning both build and don't use this workaround in certain configurations.

This should allow preview to build thunk correctly when swift caching is enabled for the regular build.

rdar://152107465
(cherry picked from commit 6c6309e4b37a0775db191f8bcc503f90c7c9edc7)


cachemeifyoucan:
@swift-ci test

owenv (approved):

PR #596, opened 2025-06-17
Use -Isystem and -Fsystem for swift-api-digester

ian-twilightcoder:
swift-api-digester now supports -Isystem and -Fsystem in the driver and tool. Use those arguments instead of -I and -F so that system paths are used for both baselines and checks.

rdar://152747822


ian-twilightcoder:
https://github.com/swiftlang/swift/pull/82283

@swift-ci test

owenv:
@swift-ci test Windows

PR #597, opened 2025-06-18
[6.2] Skip Windows tests triggering hangs in CI

owenv:
Cherrypick these workarounds since we're observing issues on the release branch as well


owenv:
@swift-ci test

PR #599, opened 2025-06-19
[#bundle support] Generate internal bundle access class for mergeable libraries

matthewseaman:
Mergeable libraries use a re-exported library in Debug mode and merge into the client in Release builds.
In both cases, we cannot simply rely on the `#dsohandle` for bundle lookup as `#bundle` does by default.

Instead, `dyld` includes special bundle loading hooks that allows `BundleForClass` to return the current resource bundle at runtime.

In this PR, we have mergeable libraries codegen a simple `__BundleLookupHelper` class and define the `SWIFT_BUNDLE_LOOKUP_HELPER_AVAILABLE` compilation condition.
The implementation of `#bundle` in Foundation will then lookup `Bundle(for: __BundleLookupHelper.self)` under this condition.

This should allow `#bundle` to work correctly for mergeable libraries.

### DIFF
--- a/Sources/SWBCore/Settings/Settings.swift
+++ b/Sources/SWBCore/Settings/Settings.swift
@@ -1554,7 +1554,7 @@ private class SettingsBuilder {
         }
 
         // Push the target derived overriding settings.
-        addTargetDerivedSettings(self.target, boundProperties.platform, boundProperties.sdk, boundProperties.sdkVariant)
+        addTargetDerivedSettings(self.target, boundProperties.platform, boundProperties.sdk, boundProperties.sdkVariant, specLookupContext)
 
         if boundDeploymentTarget.platformDeploymentTargetMacro == BuiltinMacros.DRIVERKIT_DEPLOYMENT_TARGET, let deploymentTarget = boundDeploymentTarget.platformDeploymentTarget, deploymentTarget < Version(20) {
             var table = MacroValueAssignmentTable(namespace: userNamespace)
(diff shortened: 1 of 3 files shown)


Sources/SWBCore/Settings/Settings.swift
```diff
@@ -2173,13 +2173,22 @@ private class SettingsBuilder {
...
 
             // If the product is being built as mergeable, then that overrides certain other settings.
             if scope.evaluate(BuiltinMacros.MAKE_MERGEABLE) {
-                var table = MacroValueAssignmentTable(namespace: core.specRegistry.internalMacroNamespace)
                 table.push(BuiltinMacros.STRIP_INSTALLED_PRODUCT, literal: false)
-                push(table, .exportedForNative)
             }
+
+            // Even if not being merged in this build, a mergeable library still uses a generated bundle lookup helper to power #bundle support.
+            if scope.evaluate(BuiltinMacros.MERGEABLE_LIBRARY) {
```
dmcyk, line 2189: I wonder if this effectively evaluates to true in a debug build when we're not building as mergeable? cc @mhrawdon

matthewseaman: I think this is the difference between `MERGEABLE_LIBRARY` and `MAKE_MERGEABLE`.
If `MERGEABLE_LIBRARY` is true then this is a library that is opting into being mergeable (either automatically or manually), while `MAKE_MERGEABLE` is only true in optimized builds per the lines above.

So `MERGEABLE_LIBRARY` should be what we need here.

matthewseaman:
@swift-ci test

matthewseaman:
@swift-ci test

owenv (approved):

PR #603, opened 2025-06-22
Fix the criteria for enabling modulewrap in release builds

owenv:
Closes https://github.com/swiftlang/swift-build/issues/602

By default, many WMO compiles don't emit the module separately, and won't produce modulewrap outputs. Fix the criteria used to register them as outputs to pass to the linker.


owenv:
@swift-ci test

PR #604, opened 2025-06-22
Define an RPATH_ORIGIN variable for constructing valid cross-platform paths

owenv:
Also, fix up the install name of linux unit tests


owenv:
@swift-ci test

owenv:
@swift-ci test

PR #612, opened 2025-06-24
Lookup swift-api-digester relative to the swiftc in use

owenv:
This ensures that it is always able to load the modules from the current build.


owenv:
@swift-ci test

jakepetroules (approved):

PR #613, opened 2025-06-25
rdar://154310191 Enable compilation caching for Swift packages

jansvoboda11:
This PR enables compilation caching for Swift package targets whenever normal targets have it enabled.

I originally wanted to hook into the `SuperimposedProperties` machinery and propagate it from normal targets similar to the setting for mergeable libraries. However, some aspects of `TargetDependencyResolver` made that very hard:
* treating package targets differently from normal targets (see `DiscoveredTargetInfo.immediateDependencies` and `DiscoveredTargetInfo.packageProductDependencies`),
* not propagating superimposed properties transitively,
* only visiting nodes in the build graph once (even when the specializations differ).


jansvoboda11:
@swift-ci test

jansvoboda11:
The build log says the only test failure that's not known is coming from `FileHandleTests.asyncReadFileDescriptor()`. Let's try again.

@swift-ci test macOS platform

owenv (approved):

Issue #614, opened 2025-06-11
SPM error using latest package from Xcode beta and swift-protobuf


renewedvision-kh:
When using the latest Xcode beta there is an issue with swift-protobuf:
```
error: unsupported configuration: the aggregate target 'SwiftProtobufPlugin' has package dependencies, but targets that build for different platforms depend on it
```
Is there a fix or known workaround?

Using Xcode-beta Version 26.0 (24141.23) - Build 17A5241e and SwiftProtobuf 1.30.0. The messages are show during the "Prepare build" step. The every first action after asking Xcode to build.

The SwiftProtobufPlugin target is referenced from another Swift Packages via dependency. For example,

dependencies: [
    .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.30.0")
],
and

plugins: [
    .plugin(name: "SwiftProtobufPlugin", package: "swift-protobuf")
]
In the other package's Package.swift file.

jdisho:
Interesting, I currently have the same issue in another framework. 
Reported here: https://github.com/apple/swift-openapi-generator/issues/786

Sam0711er:
+1 here with swiftgen

```
unsupported configuration: the aggregate target 'Localization' has package dependencies, but targets that build for different platforms depend on it

unsupported configuration: the aggregate target 'Localization' has package dependencies, but targets that build for different platforms depend on it

unsupported configuration: the aggregate target 'Localization' has package dependencies, but targets that build for different platforms depend on it

unsupported configuration: the aggregate target 'Localization' has package dependencies, but targets that build for different platforms depend on it

unsupported configuration: the aggregate target 'Assets' has package dependencies, but targets that build for different platforms depend on it

unsupported configuration: the aggregate target 'Files' has package dependencies, but targets that build for different platforms depend on it

unsupported configuration: the aggregate target 'Localization' has package dependencies, but targets that build for different platforms depend on it
```

michal-tomlein:
I’m hitting this issue with my [Swift CIKernel Plugin](https://github.com/in2core/swift-cikernel-plugin/).

FranckLetellier:
Same here for swiftGen

owenv:
@renewedvision-kh  thanks for the report. Have you done any of the following?

- Installed the optional Metal toolchain content (You can check this under "Components" in Xcode settings)
- Selected a Swift open source toolchain in Xcode
- Overridden the TOOLCHAINS build setting in any part of your project

[...]

Issue #570, opened 2025-06-10
Failing to build a package with a swift plugin dependency


maximkrouk:
I'm getting the following error and it seems like it's coming from swift-build when building a package that depends on an spm plugin
> _`unsupported configuration: the aggregate target 'spmgen-plugin' has package dependencies, but targets that build for different platforms depend on it`_

Simplified structure from build log output is
```
// my package
// platforms: [.iOS(.v18)]
Target 'MyTarget' in project 'ios-package'
        ➜ Explicit dependency on target 'MyClildTarget' in project 'ios-package'
Target 'MyClildTarget' in project 'ios-package'
        ➜ Explicit dependency on target 'spmgen-plugin' in project 'spmgen'
        ➜ Explicit dependency on target 'PackageResources' in project 'swift-package-resources'
Target 'MaybeSomeOtherTarget' in project 'ios-package'
        ➜ Explicit dependency on target 'spmgen-plugin' in project 'spmgen'
        ➜ Explicit dependency on target 'PackageResources' in project 'swift-package-resources'

// resources package
// generalizes models for both plugin and the app
Target 'PackageResources' in project 'swift-package-resources'
        ➜ Explicit dependency on target 'PackageResourcesCore' in project 'swift-pacakage-resources'

// my plugin (separate package)
// platforms: [.macOS(.v11)]
// even with a local dependency with platforms declaration removed the error persists
// `swift-argument-parser` package dependency is kinda required for the plugin cli
Target 'spmgen-plugin' in project 'spmgen'
        ➜ Explicit dependency on target 'spmgen' in project 'spmgen'
Target 'spmgen' in project 'spmgen'
        ➜ Explicit dependency on target 'SPMGenClient' in project 'spmgen'
        ➜ Explicit dependency on target 'ArgumentParser' in project 'swift-argument-parser'
Target 'SPMGenClient' in project 'spmgen'
        ➜ Explicit dependency on target 'PackageResourcesCore' in project 'swift-package-resources'
        ➜ Explicit dependency on target 'Prelude' in project 'swift-prelude'
```

The error message isn't really indicating why does this happen and how to fix it (everything was compiling fine with Xcode 16.x, and even with Xcode 26 beta, but after opening SwiftUI Preview in Xcode 26 beta I'm getting the compilation error, clearing caches and derived data didn't help) 🌚

[...]

PR #615, opened 2025-06-26
Fix toolchain stack equality checks when checking if specialized targets are compatible

owenv:
When the downloadable Metal toolchain is installed in Xcode, it could cause specialization compatibility checks to fail, causing downstream errors specializing targets representing package plugins. This happened because the compatibility check compared the `toolchains` of `SpecializationParameters` with the `TOOLCHAINS` macro in the configured target's settings. We should use `settings.toolchains` on the RHS instead to ensure both sides of the comparison have a chance to inject the default toolchain during settings construction.

Closes https://github.com/swiftlang/swift-build/issues/614
Closes https://github.com/swiftlang/swift-build/issues/570


owenv:
@swift-ci test

owenv:
@swift-ci test macOS

owenv:
@swift-ci test Linux

owenv:
@swift-ci test macOS

jakepetroules (approved):

PR #616, opened 2025-06-26
Don't require objc in modulemaps generated for Swift compatibility headers

xymus:
Let's remove the `requires objc` from the module map file generated for the compatibility header. This condition triggered an error when importing the compatibility header from a C source file.

The `require objc` isn't necessary, the compatibility header is already printed in a way where the Objective-C code is protected behind a language check. C clients can safely import the current compatibility header even if they may not see any content.

Let's lift this restriction. It isn't currently necessary and we're adding C content to the compatibility header with the official support for `@cdecl` that is independent of Objective-C.

swiftpm equivalent change: https://github.com/swiftlang/swift-package-manager/pull/8736


owenv (approved):

xymus:
@swift-ci please test

jakepetroules (approved):

xymus:
@swift-ci please test macOS

PR #620, opened 2025-06-27
Dependencies: validate Swift imports against MODULE_DEPENDENCIES and provide fix-its (rdar://150314567)

mirza-garibovic:
Dependencies: validate Swift imports against MODULE_DEPENDENCIES and provide fix-its (rdar://150314567)

Also includes fixes for:

- Macro: fix an issue where locations from xcconfig files are lost across condition binding
- launch-xcode: honor DEVELOPER_DIR

### DIFF
--- a/Plugins/launch-xcode/launch-xcode.swift
+++ b/Plugins/launch-xcode/launch-xcode.swift
@@ -40,8 +40,8 @@ struct LaunchXcode: CommandPlugin {
 
         print("Launching Xcode...")
         let process = Process()
-        process.executableURL = URL(fileURLWithPath: "/usr/bin/open")
-        process.arguments = ["-n", "-F", "-W", "--env", "XCBBUILDSERVICE_PATH=\(buildServiceURL.path())", "-b", "com.apple.dt.Xcode"]
+        process.executableURL = URL(fileURLWithPath: "/bin/sh")
+        process.arguments = ["-c", "open -n -F -W --env 'XCBBUILDSERVICE_PATH=\(buildServiceURL.path(percentEncoded: false))' $(xcode-select -p)/../.."]
         process.standardOutput = nil
         process.standardError = nil
         try await process.run()
--- a/Sources/SWBCore/Dependencies.swift
+++ b/Sources/SWBCore/Dependencies.swift
@@ -1,0 +1,183 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+public import SWBUtil
+import SWBMacro
+
+public struct ModuleDependency: Hashable, Sendable, SerializableCodable {
+    public let name: String
+    public let accessLevel: AccessLevel
+
+    public enum AccessLevel: String, Hashable, Sendable, CaseIterable, Codable, Serializable {
+        case Private = "private"
+        case Package = "package"
+        case Public = "public"
+    }
+
+    public init(name: String, accessLevel: AccessLevel) {
+        self.name = name
+        self.accessLevel = accessLevel
+    }
+
+    public init(entry: String) throws {
+        let components = entry.split(separator: " ")
+        guard (1...2).contains(components.count) else {
+            throw StubError.error("expected 1 or 2 space-separated components in: \(entry)")
+        }
+
+        let accessLevel: AccessLevel
+        if components.count > 1 {
+            if let a = AccessLevel(rawValue: String(components[0])) {
+                accessLevel = a
+            }
+            else {
+                throw StubError.error("unexpected access modifier '\(components[0])', expected one of: \(AccessLevel.allCases.map { $0.rawValue }.joined(separator: ", "))")
+            }
+        }
+        else {
+            accessLevel = .Private
+        }
+
+        self.name = String(components.last!)
+        self.accessLevel = accessLevel
+    }
+
+    public var asBuildSettingEntry: String {
+        "\(accessLevel == .Private ? "" : "\(accessLevel.rawValue) ")\(name)"
+    }
+
+    public var asBuildSettingEntryQuotedIfNeeded: String {
+        let e = asBuildSettingEntry
+        return e.contains(" ") ? "\"\(e)\"" : e
+    }
+}
+
+public struct ModuleDependenciesContext: Sendable, SerializableCodable {
+    var validate: BooleanWarningLevel
+    var moduleDependencies: [ModuleDependency]
+    var fixItContext: FixItContext?
+
+    init(validate: BooleanWarningLevel, moduleDependencies: [ModuleDependency], fixItContext: FixItContext? = nil) {
+        self.validate = validate
+        self.moduleDependencies = moduleDependencies
+        self.fixItContext = fixItContext
+    }
+
+    public init?(settings: Settings) {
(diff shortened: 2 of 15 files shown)


mirza-garibovic:
@swift-ci test

Sources/SWBCore/LibSwiftDriver/LibSwiftDriver.swift
```diff
@@ -12,7 +12,7 @@
 
 import Foundation
 
-import SwiftDriver
+public import SwiftDriver
```
owenv: Ideally we should avoid exposing SwiftDriver types in public API of SWBCore

Sources/SWBCore/Dependencies.swift
```diff
@@ -0,0 +1,181 @@
...
+        case Public = "public"
+    }
+
+    public init(name: String, accessLevel: AccessLevel) {
+        self.name = name
+        self.accessLevel = accessLevel
+    }
+
+    public init(entry: String) throws {
+        let components = entry.split(separator: " ")
```
jakepetroules, line 32: I suggest using Swift regex for this. Makes it much easier to read and understand the code flow (and no force unwraps that risk crashing if this becomes more complex in future).

```swift
guard let match = try #/(?<accessLevel>.+)? (?<moduleName>)/#.wholeMatch(in: entry)?.output else {
    throw StubError.error("expected 1 or 2 space-separated components in: \(entry)")
}

self.accessLevel = try match.accessLevel.map { accessLevel in
    if let a = AccessLevel(rawValue: String(accessLevel)) {
        return a
    } else {
        throw StubError.error("unexpected access modifier '\(accessLevel)', expected one of: \(AccessLevel.allCases.map { $0.rawValue }.joined(separator: ", "))")
    }
} ?? .Private
self.name = String(match.moduleName)
```

Sources/SWBCore/LibSwiftDriver/LibSwiftDriver.swift
owenv: Could this expose the list of Module dependencies instead of constructing that in the TaskAction?

Sources/SWBCore/Dependencies.swift
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        guard !missingDeps.isEmpty else { return [] }
+
+        // This is structured as one umbrella diagnostic with child diagnostics for each missing dependency
+        //
+        // 1. The umbrella allows a single fixit to resolve all missing dependencies, which makes initial adoption easier
+        // 2. The children allow a user to see where dependencies were introduced, which is sometimes in generated code
+        // 3. The umbrella is currently required because if the children were top-level diagnostics with locations inside source files, they would get replaced by live issues / sourcekit
+        // 4. In Xcode, accepting a fixit from any child removes the top-level diagnostic, so for a given target, any child fixit will actually resolve all deps
```
jakepetroules: Should we really be documenting Xcode UI behaviors here? Xcode isn't part of the Swift project and those behaviors could change over time.

Sources/SWBMacro/MacroValueAssignmentTable.swift
```diff
@@ -511,3 +511,21 @@ private extension MacroValueAssignment {
         return (expression.isLiteral && conditions == nil) || (next?.containsUnconditionalLiteralInChain ?? false)
     }
 }
+
+// MARK: - Sequence Utilities
+
+extension MacroValueAssignment {
+    /// Returns a sequence that iterates through the linked list of `next` assignments starting from this node
+    public var sequence: some Sequence<MacroValueAssignment> {
```
jakepetroules, line 519: nice!

jakepetroules:
If there's a build setting assignment in xcconfig, you get a diagnostic with a fixit, otherwise if it's set on a project model entity directly (project/target), you get a diagnostic only -- right?

Sources/SWBCore/LibSwiftDriver/LibSwiftDriver.swift
mirza-garibovic: Thanks, fixed that!

mirza-garibovic:
> If there's a build setting assignment in xcconfig, you get a diagnostic with a fixit, otherwise if it's set on a project model entity directly (project/target), you get a diagnostic only -- right?

Yup

Sources/SWBCore/Dependencies.swift
mirza-garibovic: I would like a better way to define this syntax instead of "manually" splitting like this, but I'm not sure about Swift Regex here because of https://forums.swift.org/t/should-regex-be-sendable/69529 (you can't construct the regex literal once and reuse it if it isn't Sendable). I think my preference is to keep it simple until it gets more complicated and then re-evaluate. WDYT?

owenv (approved):

Sources/SWBCore/Dependencies.swift
mirza-garibovic: Fair, I've removed this

mirza-garibovic:
@swift-ci test

mirza-garibovic:
@swift-ci test

artemcm (approved):

jakepetroules:
Looks like there's still some test failures on Linux.

mirza-garibovic:
> Looks like there's still some test failures on Linux.

I changed the test to use `host` instead of `macOS` so I can get Linux and Windows coverage, going to work through the issues.

mirza-garibovic:
@swift-ci test

mirza-garibovic:
Checks are green! @jakepetroules, any remaining concerns from you?

Plugins/launch-xcode/launch-xcode.swift
```diff
@@ -40,8 +40,8 @@ struct LaunchXcode: CommandPlugin {
 
         print("Launching Xcode...")
         let process = Process()
-        process.executableURL = URL(fileURLWithPath: "/usr/bin/open")
-        process.arguments = ["-n", "-F", "-W", "--env", "XCBBUILDSERVICE_PATH=\(buildServiceURL.path())", "-b", "com.apple.dt.Xcode"]
+        process.executableURL = URL(fileURLWithPath: "/bin/sh")
```
jakepetroules, line 43: Not sure about this one.`xcode-select -p` does not necessarily return an Xcode instance, it could be Developer Command Line Tools, in which case this will just try to open /Developer and fail.

Could you run `xcode-select -p` in a separate `Process()` invocation, collect its output, and call `open` with that path if the return value ends in `.app`, otherwise fall back to `-b com.apple.dt.Xcode`?

Also suggest splitting into another PR so that concerns with that don't block your primary changes.

Sources/SWBCore/Dependencies.swift
jakepetroules: Fair enough but then I'd at least consider doing this:

```swift
let components = entry.split(separator: " ")
switch components.count {
case 1:
    accessLevel = .Private
    self.name = components[0]
case 2:
    accessLevel = try AccessLevel(string: components[0])
    self.name = components[1]
default:
    throw StubError.error("expected 1 or 2 space-separated components in: \(entry)")
}
```

(elsewhere)

```swift
fileprivate extension AccessLevel {
    init(string: String) throws {
        guard let accessLevel = AccessLevel(rawValue: string) else {
            throw StubError.error("unexpected access modifier '\(string)', expected one of: \(AccessLevel.allCases.map { $0.rawValue }.joined(separator: ", "))")
        }
        self = accessLevel
    }
}
```

The main point being to localize checking in a way that makes it as obvious as possible at a glance that it's correct (and avoiding a force unwrap).

mirza-garibovic:
Closing this PR and moving to https://github.com/swiftlang/swift-build/pull/632 because we're locking down the repo and PRs have to come from forks now, I can't update this PR.

Plugins/launch-xcode/launch-xcode.swift
mirza-garibovic: Removing this for now

PR #631, opened 2025-07-02
Add initial support for object library product type

owenv:
These are similar to the CMake concept of object libraries. They're a directory with a .objlib extension containing object files for the target. They also contain an args.resp file with the linker arguments required by consumers. This adds initial support for creating and consuming object libraries. The intent is to use these to move towards properly support the Windows linkage model in SwiftPM

### DIFF
--- a/Sources/SWBCore/SpecImplementations/Tools/LinkerTools.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/LinkerTools.swift
@@ -918,7 +918,7 @@ public final class LdLinkerSpec : GenericLinkerSpec, SpecIdentifierType, @unchec
         )
 
         // Create the task.
-        delegate.createTask(type: self, payload: payload, ruleInfo: defaultRuleInfo(cbc, delegate), commandLine: commandLine, environment: environmentFromSpec(cbc, delegate), workingDirectory: cbc.producer.defaultWorkingDirectory, inputs: inputs, outputs: outputs, action: nil, execDescription: resolveExecutionDescription(cbc, delegate), enableSandboxing: enableSandboxing)
+        delegate.createTask(type: self, payload: payload, ruleInfo: defaultRuleInfo(cbc, delegate), commandLine: commandLine, environment: environmentFromSpec(cbc, delegate), workingDirectory: cbc.producer.defaultWorkingDirectory, inputs: inputs, outputs: outputs, action: createTaskAction(cbc, delegate), execDescription: resolveExecutionDescription(cbc, delegate), enableSandboxing: enableSandboxing)
     }
 
     public func constructPreviewsBlankInjectionDylibTask(
@@ -992,7 +992,7 @@ public final class LdLinkerSpec : GenericLinkerSpec, SpecIdentifierType, @unchec
             workingDirectory: cbc.producer.defaultWorkingDirectory,
             inputs: [],
             outputs: outputs,
-            action: nil,
+            action: createTaskAction(cbc, delegate),
             execDescription: resolveExecutionDescription(cbc, delegate),
             enableSandboxing: enableSandboxing
         )
@@ -1246,6 +1246,9 @@ public final class LdLinkerSpec : GenericLinkerSpec, SpecIdentifierType, @unchec
             case .object:
                 // Object files are added to linker inputs in the sources task producer.
                 return ([], [])
+            case .objectLibrary:
+                let pathFlags = specifier.absolutePathFlagsForLd()
+                return (pathFlags, [specifier.path])
             }
         }.reduce(([], [])) { (lhs, rhs) in (lhs.args + rhs.args, lhs.inputs + rhs.inputs) }
     }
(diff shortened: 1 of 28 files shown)


owenv:
@swift-ci test

owenv:
@swift-ci test

Sources/SWBCore/SpecImplementations/Tools/LinkerTools.swift
```diff
@@ -1500,6 +1506,8 @@ fileprivate extension LinkerSpec.LibrarySpecifier {
         case (.object, _):
             // Object files are added to linker inputs in the sources task producer.
             return []
+        case (.objectLibrary, _):
+            return ["@\(path.join("args.resp").str)"]
```
jakepetroules, line 1533: Not all tools on all platforms support response files; we already have CLANG_USE_RESPONSE_FILE / LIBTOOL_USE_RESPONSE_FILE for that reason. I think we need some checks here?

Sources/SWBCore/SpecImplementations/Tools/LinkerTools.swift
```diff
@@ -1601,6 +1609,10 @@ public final class LibtoolLinkerSpec : GenericLinkerSpec, SpecIdentifierType, @u
                 // Object files are added to linker inputs in the sources task producer and so end up in the link-file-list.
                 return []
 
+            case .objectLibrary:
+                inputPaths.append(specifier.path)
+                return ["@\(specifier.path.join("args.resp").str)"]
```
jakepetroules, line 1642: This should respect LIBTOOL_USE_RESPONSE_FILE (see above)

Sources/SWBTaskConstruction/TaskProducers/BuildPhaseTaskProducers/SourcesTaskProducer.swift
```diff
@@ -1624,20 +1633,30 @@ final class SourcesTaskProducer: FilesBasedBuildPhaseTaskProducerBase, FilesBase
...
 
-            // FIXME: Emit a warning here.
+                // FIXME: Emit a warning here.
+            }
         }
 
         // Return the default linker.
-        identifier = isStaticLib ? LibtoolLinkerSpec.identifier : LdLinkerSpec.identifier
+        let identifier: String
+        if isObjectLibrary {
```
jakepetroules: Use a `switch (isObjectLibrary, isStaticLib)` to explicitly handle all cases? It's a bit ambiguous what happens if both are true; actually I think that should probably an error because a while ago we switched to MACH_O_TYPE consistently being the source of truth (since people set it, so there are dylib product types with MACH_O_TYPE set to staticlib, and static lib product types with it set to mh_dylib).

That said, should we also consider driving this entirely off a new MACH_O_TYPE value like "objectlib" for consistency and code simplicity? It's not like "staticlib" is an actual Mach-O object type anyways, so it's already overloaded and this doesn't correspond 1:1 with the Mach-O file format header's file type field (and is has way more values anyways).

Sources/SWBTaskExecution/TaskActions/ObjectLibraryAssemblerTaskAction.swift
```diff
@@ -0,0 +1,49 @@
...
+public import SWBCore
+import SWBUtil
+import ArgumentParser
+
+public final class ObjectLibraryAssemblerTaskAction: TaskAction {
+    public override class var toolIdentifier: String {
+        return "assemble-object-library"
+    }
+
+    private struct Options: ParsableArguments {
```
jakepetroules, line 22: Been waiting forever to be able to use this in task action implementations ❤️

Sources/SWBTestSupport/TestWorkspaces.swift
```diff
@@ -1028,6 +1031,8 @@ package final class TestStandardTarget: TestInternalTarget, Sendable {
                 return "lib\(name).a"
             case .objectFile:
                 return "\(name).o"
+            case .objectLibrary:
+                return "\(name).objlib"
```
jakepetroules, line 1037: Is .objlib what CMake uses? Doesn't particularly matter, just wondering if there's any prior art at all

owenv: No precedent, I just made this up

Sources/SWBCore/SpecImplementations/Tools/LinkerTools.swift
owenv: hmm, this is going to be annoying because determining the full list of object files a dependency target will produce from the dependent is tricky

owenv: I reworked this with a linker task action which expands the response files if needed before invoking the linker

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test

jakepetroules (approved):

PR #633, opened 2025-07-02
Ignore `-no_exported_symbols` when the debug dylib is used (rdar://152244838)

jonathanpenn:
The stub executor needs to trampoline to the entry point in the debug dylib. But linking with `-no_exported_symbols` strips it and launching the app fails to find the symbol.
    
Handles both the `LD_EXPORT_SYMBOLS` build setting and manual flags added to `OTHER_LDFLAGS`.

Cherry-picked from `b133772`.


owenv (approved):

jakepetroules (approved):

jonathanpenn:
@swift-ci test

PR #635, opened 2025-07-03
Validate MODULE_DEPENDENCIES using tracing from Clang

bob-wilson:
This currently only checks Clang compilations with modules because there were some issues with introducing a new task action for non-modular Clang compilations.

rdar://150313957

### DIFF
--- a/Sources/SWBCore/SpecImplementations/Tools/CCompiler.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/CCompiler.swift
@@ -1344,12 +1344,16 @@ public class ClangCompilerSpec : CompilerSpec, SpecIdentifierType, GCCCompatible
         }
 
         if let moduleDependenciesContext {
-            guard let jsonData = try? JSONEncoder(outputFormatting: [.prettyPrinted, .sortedKeys, .withoutEscapingSlashes]).encode(moduleDependenciesContext),
-               let signature = String(data: jsonData, encoding: .utf8) else {
-                delegate.error("failed to serialize 'MODULE_DEPENDENCIES' context information")
+            do {
+                let jsonData = try JSONEncoder(outputFormatting: [.prettyPrinted, .sortedKeys, .withoutEscapingSlashes]).encode(moduleDependenciesContext)
+                guard let signature = String(data: jsonData, encoding: .utf8) else {
+                    throw StubError.error("non-UTF-8 data")
+                }
+                additionalSignatureData += "|\(signature)"
+            } catch {
+                delegate.error("failed to serialize 'MODULE_DEPENDENCIES' context information: \(error)")
                 return
             }
-            additionalSignatureData += "|\(signature)"
         }
 
         // Finally, create the task.


Sources/SWBCore/Dependencies.swift
```diff
@@ -181,4 +223,9 @@ public struct ModuleDependenciesContext: Sendable, SerializableCodable {
             return Diagnostic.FixIt(sourceRange: sourceRange, newText: newText)
         }
     }
+
+    func signatureData() -> String {
```
mirza-garibovic: I don't think this quite belongs here. For one, the type is serializable/codable, so clients can serialize it to get a signature blob. And two, what actually goes into a signature depends on each specific client task, so there can't be a central definition like this. That said, I think the clang task signature actually should contain the full value of ModuleDependenciesContext (and probably the entire task Payload tbh? @owenv is that not already the case?) instead of just validate+names, because things like access level and fixit context do influence the diagnostics the clang task emits, so we'd want to re-run when the user changes any of those to avoid ending up with stale diagnostics that give you incorrect fixits.

Sources/SWBCore/SpecImplementations/Tools/CCompiler.swift
```diff
@@ -432,7 +432,10 @@ public struct ClangTaskPayload: ClangModuleVerifierPayloadType, DependencyInfoEd
 
     public let fileNameMapPath: Path?
 
-    fileprivate init(serializedDiagnosticsPath: Path?, indexingPayload: ClangIndexingPayload?, explicitModulesPayload: ClangExplicitModulesPayload? = nil, outputObjectFilePath: Path? = nil, fileNameMapPath: Path? = nil, developerPathString: String? = nil) {
+    public let moduleDependenciesContext: ModuleDependenciesContext?
+    public let traceFile: Path?
```
mirza-garibovic: Should have the `Path` suffix like `fileNameMapPath`

Sources/SWBCore/SpecImplementations/Tools/CCompiler.swift
```diff
@@ -1316,8 +1343,12 @@ public class ClangCompilerSpec : CompilerSpec, SpecIdentifierType, GCCCompatible
...
+        delegate.createTask(type: self, dependencyData: dependencyData, payload: payload, ruleInfo: ruleInfo, additionalSignatureData: additionalSignatureData, commandLine: commandLine, additionalOutput: additionalOutput, environment: environmentBindings, workingDirectory: compilerWorkingDirectory(cbc), inputs: inputNodes + extraInputs, outputs: [outputNode], action: action ?? delegate.taskActionCreationDelegate.createClangNonModularCompileTaskAction(), execDescription: resolveExecutionDescription(cbc, delegate), enableSandboxing: enableSandboxing, additionalTaskOrderingOptions: [.compilationForIndexableSourceFile], usesExecutionInputs: usesExecutionInputs, showEnvironment: true, priority: .preferred)
```
mirza-garibovic: I think this PR is just for modular builds right?

Sources/SWBTaskExecution/TaskActions/ClangCompileTaskAction.swift
```diff
@@ -248,6 +248,20 @@ public final class ClangCompileTaskAction: TaskAction, BuildValueValidatingTaskA
...
+            var traceFile: Path? = nil
+            var moduleDependenciesContext: ModuleDependenciesContext? = nil
+            if let payload = task.payload as? ClangTaskPayload {
+                traceFile = payload.traceFile
+                moduleDependenciesContext = payload.moduleDependenciesContext
+            }
+            if let traceFile {
+                // Remove the trace output file if it already exists.
+                if executionDelegate.fs.exists(traceFile) {
+                    try executionDelegate.fs.remove(traceFile)
```
mirza-garibovic: Does clang create or append to this particular trace file just like the ld trace we looked at?

Sources/SWBTaskExecution/TaskActions/ClangCompileTaskAction.swift
```diff
@@ -304,6 +318,24 @@ public final class ClangCompileTaskAction: TaskAction, BuildValueValidatingTaskA
                     return lastResult ?? .failed
                 }
             }
+
+            if let moduleDependenciesContext, let traceFile, lastResult == .succeeded {
+                // Verify the dependencies from the trace data.
+                let fs = executionDelegate.fs
+                let traceData = try JSONDecoder().decode(Array<TraceData>.self, from: fs.readMemoryMapped(traceFile))
```
mirza-garibovic: Why use readMemoryMapped here? I think we don't have a lot of clarity around when to use read vs readMemoryMapped and should probably default to read.

Tests/SWBBuildSystemTests/DependencyVerificationBuildOperationTests.swift
```diff
@@ -0,0 +1,99 @@
...
+
+import SWBCore
+import SWBTestSupport
+@_spi(Testing) import SWBUtil
+
+import SWBTaskExecution
+import SWBProtocol
+
+@Suite(.requireSDKs(.macOS))
+fileprivate struct DependencyVerificationBuildOperationTests: CoreBasedTests {
```
mirza-garibovic: I suggest colocating this with `validateModuleDependencies()`, rename that one to `validateModuleDependenciesSwift()`, and name this one `validateModuleDependenciesClang()`.

Sources/SWBTaskExecution/TaskActions/ClangCompileTaskAction.swift
```diff
@@ -248,6 +248,20 @@ public final class ClangCompileTaskAction: TaskAction, BuildValueValidatingTaskA
                 casDBs = nil
             }
 
+            // Check if verifying dependencies from trace data is enabled.
+            var traceFile: Path? = nil
+            var moduleDependenciesContext: ModuleDependenciesContext? = nil
```
mirza-garibovic: Nit but typically we'd make these two `let` and have `else` set them to `nil`, just like what you did with `let traceFile: Path?` in `CCompiler.swift`

Sources/SWBTaskExecution/TaskActions/ClangCompileTaskAction.swift
```diff
@@ -304,6 +318,24 @@ public final class ClangCompileTaskAction: TaskAction, BuildValueValidatingTaskA
                     return lastResult ?? .failed
                 }
             }
+
+            if let moduleDependenciesContext, let traceFile, lastResult == .succeeded {
```
mirza-garibovic: If clang doesn't have the feature, traceFile will be nil here, so we'll skip checking altogether, even if the user requested `VALIDATE_MODULE_DEPENDENCIES`. The Swift version on the other hand emits an error if `VALIDATE_MODULE_DEPENDENCIES` is set but the toolchain doesn't have the required support, mainly so that the test can check for that when it's running against older toolchains.

Sources/SWBCore/SpecImplementations/Tools/CCompiler.swift
mirza-garibovic: I went ahead and pushed a fix for this one so it'll build

Sources/SWBCore/Dependencies.swift
owenv: Probably not the entire payload because it's fairly large and a lot of the content doesn't need to be included because it's redundant with the command line. But yeah, I would just serialize ModuleDependenciesContext to form a signature for it

Tests/SWBBuildSystemTests/DependencyVerificationBuildOperationTests.swift
mirza-garibovic: Went ahead and did this and aligned them a bit

Sources/SWBCore/Dependencies.swift
mirza-garibovic: Oops, that means that my Swift change is incorrect for incremental builds, since it doesn't add this context to its signature. Going to fix that.

Sources/SWBTaskExecution/TaskActions/ClangCompileTaskAction.swift
bob-wilson: It will create the file if necessary and append to any existing content.

Sources/SWBTaskExecution/TaskActions/ClangCompileTaskAction.swift
bob-wilson: If it is going to emit a diagnostic, then it needs to be a warning, not an error. (The same is true for the Swift version of the check.) Otherwise, enabling `VALIDATE_MODULE_DEPENDENCIES` will break your build unless you are certain that the toolchain will always be new enough. I'm not really convinced that a warning adds much value, though. The test for the Clang version is checking directly whether the compiler supports the necessary tracing feature. Is that harder to do for the Swift version?

Sources/SWBCore/Dependencies.swift
bob-wilson: OK, I added the JSON serialization of the ModuleDependenciesContext. I renamed `signatureData` to `jsonData` but the function is still here in the same place, so I'm not sure if that fully addresses your concern. Also, is JSON the right thing to use in this context?

mirza-garibovic: It includes everything so it addresses my functional concerns, thanks! For nitty concerns, I think you can inline the call to JSONEncoder in CCompiler.swift since it doesn't add much to have that on ModuleDependenciesContext. I do think CCompiler.swift should handle errors by emitting a diagnostic instead of `try?`.

Sources/SWBTaskExecution/TaskActions/ClangCompileTaskAction.swift
mirza-garibovic: As a general rule, if a user requests a feature but the build system is unable to fulfill it, it shouldn't silently ignore the request and proceed incorrectly, it should fail or warn. TBH we aren't very consistent about this... I do think an error or warning has value, because users should know that they aren't getting the validation they expect. If they're building in an environment with a toolchain which doesn't support the feature, they should conditionalize out `VALIDATE_MODULE_DEPENDENCIES` in that environment. How about we make it emit warnings and I'll change the Swift side?

mirza-garibovic:
@swift-ci test

Sources/SWBTaskExecution/TaskActions/ClangCompileTaskAction.swift
bob-wilson: OK. I already made it a warning for the Clang part here.

Sources/SWBCore/Dependencies.swift
bob-wilson: OK

bob-wilson:
@swift-ci test

Sources/SWBCore/SpecImplementations/Tools/CCompiler.swift
```diff
@@ -1316,6 +1343,15 @@ public class ClangCompilerSpec : CompilerSpec, SpecIdentifierType, GCCCompatible
             extraInputs = []
         }
 
+        if let moduleDependenciesContext {
+            guard let jsonData = try? JSONEncoder(outputFormatting: [.prettyPrinted, .sortedKeys, .withoutEscapingSlashes]).encode(moduleDependenciesContext),
+               let signature = String(data: jsonData, encoding: .utf8) else {
+                delegate.error("failed to serialize 'MODULE_DEPENDENCIES' context information")
```
mirza-garibovic: There's a `try?` on 1347 that this isn't handling

bob-wilson: Maybe I'm missing something obvious, but I think it does. If the `encode` throws, the `try?` will produce a `nil` and the guard will fail. No?

mirza-garibovic: Sorry, yes the guard will fail, but it's discarding the specific error detail about why it's failing. It should capture the error and emit it to help us triage failures.

bob-wilson: Ah, OK. I changed it to capture the error details.

mirza-garibovic (approved):

mirza-garibovic:
@swift-ci test

PR #636, opened 2025-07-03
Adapt QNX support to the introduction of LINKER_DRIVER

jakepetroules:
_(migrated from #398 in accordance with our new policy to require PRs from forks)_


jakepetroules:
@swift-ci test

PR #584, opened 2025-06-13
Conditionally adopt swift-subprocess

jakepetroules:
This fully resolves working directory thread safety issues with subprocess spawning across all platforms. For now, subprocess is adopted conditionally in order to continue building in certain environments where the Subprocess module may not be available, in which case we fall back to Foundation Process.

Closes #441

PR #638, opened 2025-07-03
Conditionally adopt swift-subprocess

### LINKED ISSUE #441: Amazon Linux 2 Support for Swift Build
Amazon Linux 2 is a little more challenging to support than other Linux distributions. In particular, there is a missing implementation of [posix_spawn_file_actions_addchdir](https://pubs.opengroup.org/onlinepubs/9799919799/functions/posix_spawn_file_actions_addchdir.html) for the version of glibc available there. There could be other issues with Swift Build support for that platform too, but this is the most immediate problem.

jakepetroules:
This fully resolves working directory thread safety issues with subprocess spawning across all platforms. For now, subprocess is adopted conditionally in order to continue building in certain environments where the Subprocess module may not be available, in which case we fall back to Foundation Process.

Closes #441

_(migrated from #584 in accordance with our new policy to require PRs from forks)_


jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

Sources/SWBUtil/ProcessController.swift
```diff
@@ -0,0 +1,177 @@
...
+            let task = Task<Void, Never>.detached { [path, arguments, environment, workingDirectory, done] in
+                defer { done.signal() }
+                let result = await Result.catching {
+                    #if !canImport(Darwin) || os(macOS)
+                    #if canImport(Subprocess)
+                    var platformOptions = PlatformOptions()
+                    platformOptions.teardownSequence = [.gracefulShutDown(allowedDurationToNextStep: .seconds(5))]
+                    #if os(macOS)
+                    if highPriority {
+                        platformOptions.qualityOfService = .userInitiated
```
jakepetroules, line 85: @compnerd Would it seem reasonable to call SetPriorityClass with HIGH_PRIORITY_CLASS for Windows (REALTIME_PRIORITY_CLASS definitely sounded excessive)? Currently this is used to elevate the build service process so that it can serve high-priority requests more efficiently (e.g. with Swift Concurrency task priority set higher than normal).

compnerd: Hmm, we should try it - it _can_ saturate the system to the point of being unusable though, so I am a little concerned about that.

jakepetroules:
@swift-ci test

neonichu (approved):

Sources/SWBUtil/Process.swift
```diff
@@ -81,6 +84,23 @@ extension Process {
...
 extension Process {
     public static func getOutput(url: URL, arguments: [String], currentDirectoryURL: URL? = nil, environment: Environment? = nil, interruptible: Bool = true) async throws -> Processes.ExecutionResult {
+        #if canImport(Subprocess)
+        #if !canImport(Darwin) || os(macOS)
+        var platformOptions = PlatformOptions()
+        if interruptible {
+            platformOptions.teardownSequence = [.gracefulShutDown(allowedDurationToNextStep: .seconds(5))]
+        }
+        let result = try await Subprocess.run(.path(FilePath(url.filePath.str)), arguments: .init(arguments), environment: environment.map { .custom(.init($0)) } ?? .inherit, workingDirectory: (currentDirectoryURL?.filePath.str).map { FilePath($0) } ?? nil, platformOptions: platformOptions, body: { execution, inputWriter, outputReader, errorReader in
+            try await inputWriter.finish()
```
weissi, line 103: potential deadlock here. You can't assume that you can fully unload the stdin w/o reading stdout/err

Sources/SWBUtil/Process.swift
```diff
@@ -81,6 +84,23 @@ extension Process {
...
+            async let stdoutBytes = outputReader.collect().flatMap { $0.withUnsafeBytes(Array.init) }
```
weissi: two `*Unsafe*`s here

Sources/SWBUtil/Process.swift
```diff
@@ -81,6 +84,23 @@ extension Process {
...
+            async let stderrBytes = errorReader.collect().flatMap { $0.withUnsafeBytes(Array.init) }
```
weissi: `collect` will load all of this potentially infinite stream into memory and I don't even see a limit, seems dangerous. If anything, limit it to something high like 32 MB and throw if we're blowing through that.

Sources/SWBUtil/Process.swift
jakepetroules: so? we can't use Span yet, this still has to build with 6.1

Sources/SWBUtil/Process.swift
jakepetroules: There's no deadlock here, all this is doing is closing a file descriptor. We have no input. That said, I'll move it below the output reading just to be correct.

Sources/SWBUtil/Process.swift
jakepetroules: This is existing behavior; we can address it separately.

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

jakepetroules:
Moving to draft -- SwiftSubprocess is on its way to 1.0 with some API changes, so we might as well hold off until then.

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

PR #639, opened 2025-07-03
Add register app groups build setting

baujla:
Documentation is needed for the register app groups build setting.


baujla:
@swift-ci test

owenv (approved):

PR #642, opened 2025-07-03
Add support for the libswiftCompatibilitySpan.dylib backward deployment library

DougGregor:
The Swift library `libswiftCompatibilitySpan.dylib` allows uses of `Span`, `MutableSpan`, etc. to deploy to earlier OS versions. Detect when it is needed and copy it into the build artifact.


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

jakepetroules:
Note that we need this in `release/6.2`. You can either cherry pick it there afterwards, or land it in `release/6.2` first and it'll be automatically forwarded to `main`.

Also, @slavapestov opened a PR for this a couple months ago (#359), I guess we should close that one now.

Sources/SWBCore/PlatformRegistry.swift
```diff
@@ -287,6 +295,29 @@ extension Platform {
 
         return version >= minimumSwiftConcurrencyVersion
     }
+
+    /// Determines if the platform natively supports Swift 6.2's Span type. If `false`, then the Swift Span back-compat lib needs to be copied into the app/framework's bundle.
+    public func supportsSwiftSpanNatively(_ scope: MacroEvaluationScope, forceNextMajorVersion: Bool = false, considerTargetDeviceOSVersion: Bool = true) -> Bool? {
```
jakepetroules: Why add default values for the arguments? This is only ever called in one place.

DougGregor:
Ah, I can replace this with one that targets release/6.2. Didn't realize that's how swift-build is handling branching.

Sources/SWBCore/SDKRegistry.swift
```diff
@@ -445,12 +449,12 @@ public final class SDKVariant: PlatformInfoProvider, Sendable {
...
 
-    private static func fallbackSwiftVersions(variantName name: String) -> (String?, String?) {
+    private static func fallbackSwiftVersions(variantName name: String) -> (os: String?, concurrency: String?, span: String?) {
         switch name {
         case "macos", "macosx":
-            return ("10.14.4", "12.0")
+            return ("10.14.4", "12.0", "26.0")
         default:
-            return (nil, nil)
+            return (nil, nil, "26.0")
```
jakepetroules, line 457: Just remove the `case "macos", "macosx"` now; that was a compatibility polyfill and isn't needed anymore.

Sources/SWBTestSupport/SkippedTestSupport.swift
```diff
@@ -283,6 +283,15 @@ extension Trait where Self == Testing.ConditionTrait {
         })
     }
 
+    package static func requireXcode26(sourceLocation: SourceLocation = #_sourceLocation) -> Self {
```
jakepetroules, line 278: The body of this function can just be implemented as `requireMinimumXcodeBuildVersion("17A1", sourceLocation: sourceLocation)`

jakepetroules (changes requested):

jakepetroules:
> Ah, I can replace this with one that targets release/6.2. Didn't realize that's how swift-build is handling branching.

Yeah, we have a custom job set up... bit different from the rest of swiftlang.

FYI, you can just change the PR's target branch and force push without needing to open a new PR.

jakepetroules:
The Windows failures:

> × Test debuggableCommandLineTool() recorded an issue at BuildOperationTests.swift:390:17: Expectation failed: (executionResult.exitStatus → terminated with uncaught signal 309) == (.exit(0) → exited with status 0)

...look like STATUS_DLL_NOT_FOUND, meaning the binary that's running is failing to find some dependent DLL.

EDIT: not a blocker, this is showing up everywhere.

owenv (approved):
This is looking good to me. There's a chance some more tests might need tweaks since CI is on Xcode 16.2 and associated SDKs, but we'll see.

owenv:
@swift-ci test

Sources/SWBCore/PlatformRegistry.swift
DougGregor: I've dropped them.

Sources/SWBCore/SDKRegistry.swift
DougGregor: I'd prefer to do that as a separate PR

DougGregor:
@swift-ci please test

DougGregor:
> FYI, you can just change the PR's target branch and force push without needing to open a new PR.

Yep, done!

Sources/SWBCore/SDKRegistry.swift
jakepetroules: OK, that's fine

jakepetroules:
@swift-ci test macos

jakepetroules:
I misread the CI results -- there's a few which still need adjustment. Here's a link to the last failing run: https://ci.swift.org/job/pr-swift-build-macos/1004/console

DougGregor:
@owenv has a suggestion for making the tests run properly on older Xcodes, I'll get that pushed soon

DougGregor:
@swift-ci test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Linux

DougGregor:
@swift-ci please test

PR #643, opened 2025-07-04
Use SwiftSystem.Errno instead of calling strerror directly

jakepetroules:
This resolves some builds warnings on Windows, which helps log readability.

(Yes I know we could alternatively just disable those warnings, but technically strerror does have thread safety issues and if we're going to fix that it would be nice to just do it in SwiftSystem in one place only)


jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test windows

PR #644, opened 2025-07-05
Fix flakiness in asyncReadFileDescriptor test

jakepetroules:
It's not safe to "modify" a file descriptor while it's in use by DispatchIO, per the documentation, which is the likely cause of flakiness here.


jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test windows

owenv (approved):

PR #646, opened 2025-07-08
Cherry-pick: Revert change to avoid dead code striping with debug dylib (rdar://154656898)

jonathanpenn:
We don't need this right now. Turned out to be more restrictive than necessary.

Cherry-picked from release/6.2


jonathanpenn:
@swift-ci test

PR #648, opened 2025-07-10
[6.2] Update metal tests to account for a downloaded metal toolchain

owenv:
Update metal tests to account for a downloaded metal toolchain. If needed, run xcodebuild -showComponent metalToolchain --json to fetch the external toolchain info and inject it into tests which rely on the Metal compiler,

rdar://146431505


owenv:
@swift-ci test

owenv:
@swift-ci test

jakepetroules (approved):

PR #652, opened 2025-07-11
Upstream: Extract Extension Point Definition and Extension's target extension point identifier from swift const values

jakepetroules:


jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

PR #654, opened 2025-07-14
Pass -platform_version to the linker when prelinking a target on Darwin platforms

owenv:
This ensures we don't fail the build in situations where it can't be inferred from the inputs

rdar://154346861


owenv:
@swift-ci test

owenv:
@swift-ci test

PR #655, opened 2025-07-15
Loosen test expectations in framework pruning tests

owenv:
These tests aren't intended to check for code signing behavior, so filter out the _CodeSignature folder to simplify comparisons

rdar://155531620


owenv:
@swift-ci test

PR #657, opened 2025-07-16
Add launcher binary that allows for debugging missing DLL dependencies at load time

jakepetroules:
Background: on Unix platforms, running a binary with a missing dependent shared library produces a human-readable error message on stderr containing the name of the missing library. On Windows, the loader may show a graphical dialog box with missing the DLL names when a process is run from a UI context, but in most cases the process will simply exit with `STATUS_DLL_NOT_FOUND` and no other information. This can make debugging missing dependent shared libraries very difficult.

The program in this PR uses Loader Snaps and documented SPI to replicate the Unix experience on Windows, printing nice human readable messages like `This application has failed to start because \(missingDLL) was not found.` to stderr. It's implemented as a launcher (think ccache), where its arguments are the command line that would be originally executed.

Mostly putting this up to illustrate the technique, not immediately planning to land this. Not sure if this is the right place for it either, or if Swift is even a good choice of implementation language (due to the fact that the launcher itself will have DLL dependences of its own on the Swift runtime).

It also has to be a launcher binary and not a set of helper APIs, since some of the APIs called in here really only make sense in the context of a main executable and not in library code.

Thanks to https://github.com/TimMisiak/LoaderLog/blob/main/LoaderLog.cpp for some learnings behind this.

### DIFF
--- a/Sources/swblauncher/main.swift
+++ b/Sources/swblauncher/main.swift
@@ -1,0 +1,291 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import WinSDK
+
+// Also see winternl.h in the Windows SDK for the definitions of a number of these structures.
+
+// https://learn.microsoft.com/en-us/windows/win32/api/winternl/nf-winternl-ntqueryinformationprocess#parameters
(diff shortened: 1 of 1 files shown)


Sources/swblauncher/main.swift
```diff
@@ -0,0 +1,291 @@
...
+        }
+    }
+    return Int32(bitPattern: exitCode)
+}
+
+func main() -> Int32 {
+    do {
+        var commandLine = String(decodingCString: GetCommandLineW(), as: UTF16.self)
+
+        // FIXME: This could probably be more robust
```
jakepetroules, line 270: I think I'll eliminate the command line parsing here and just re-invoke the exact command line string, requiring clients to invoke _this_ process with the lpApplicationName parameter to CreateProcessW set to the path to the swblauncher so that this doesn't need to do any parsing.

jakepetroules:
I will probably try to land this once @compnerd lands the static linking support for the Swift runtime on Windows.

PR #659, opened 2025-07-18
Make Android NDK discovery more robust, and add tests

jakepetroules:
Add support for Android NDK versions 23 through 26, and emit an explicit error for older versions. Add test coverage to verify parsing of the metadata.

I chose 23 as the cutoff for now simply because that's the next last time that the abis.json schema changed. That version was released in  August 2021. We can add older versions if anyone really wants them.


jakepetroules:
@swift-ci test

neonichu (approved):

jakepetroules:
@swift-ci test linux

jakepetroules:
@swift-ci test linux

PR #662, opened 2025-07-21
Tighten up some logic around file permissions

jakepetroules:
Fix a regression (from #565) in PbxCp which caused files to inappropriately gain the execute bit. `isExecutable` is conceptually a higher level operation which (per POSIX) can return true even if the file has no execute permission bits set:

"If execute permission is requested, access shall be granted if execute permission is granted to at least one user by the file permission bits or by an alternate access control mechanism"

No test because it's not clear what the concrete conditions are under which this occurs. That said, copying now checks against the owner permission bit specifically, as that's the actual intent of what the code is meaning to do and matches the original behavior.

Additionally, fix some missing permission information in getFileInfo for the PseudoFS implementation and increase test coverage a bit. Avoids it coming back to bite later.

rdar://154663442

### DIFF
--- a/Sources/SWBUtil/FSProxy.swift
+++ b/Sources/SWBUtil/FSProxy.swift
@@ -160,7 +160,11 @@ public protocol FSProxy: AnyObject, Sendable {
     // FIXME: Need to document behavior w.r.t. error handling.
     func isDirectory(_ path: Path) -> Bool
 
-    /// Checks whether the given path has the execute bit (which on Windows is determined by the file extension).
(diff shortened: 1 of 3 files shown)


jakepetroules:
@swift-ci test

neonichu (approved):

jakepetroules:
@swift-ci test linux

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test linux

jakepetroules:
@swift-ci test linux

PR #665, opened 2025-07-22
Fix incorrect error code propagation in CodeSignatureInfo.load

jakepetroules:
In the CodeSignatureInfo.load method, when SecCodeCopySigningInformation fails, the code was incorrectly throwing MacError(result) instead of MacError(result2). This caused misleading error messages since it propagated the error code from the previous SecStaticCodeCreateWithPath call rather than the actual failing SecCodeCopySigningInformation call.

The fix ensures accurate error reporting for code signing verification failures.


neonichu (approved):

jakepetroules:
@swift-ci test

PR #671, opened 2025-07-24
Only check `isTargetSuitableForPlatformForIndex` for workspace description

hamishknight:
For the package and target build description we can end up incorrectly dropping build tool dependencies since not all clients pass in the correct dependency information. It's not clear that we actually gain much from doing this check for the target and package build descriptions though, its primary purpose is to avoid configuring unsupported targets in the workspace build description where we try to configure for all available platforms. As such, switch to only checking for the workspace build description. Note we don't encounter this issue in the workspace case since there we override the build parameters for host build tools to always build for the host platform.

rdar://152012769

### DIFF
--- a/Sources/SWBCore/DependencyResolution.swift
+++ b/Sources/SWBCore/DependencyResolution.swift
@@ -630,11 +630,13 @@ extension SpecializationParameters {
     ///
     /// Note there's an exception for this for host build tools, which are required for compilation
     /// and must therefore be configured (and registered as a dependency) regardless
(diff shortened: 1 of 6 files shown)


hamishknight:
Linux failure is unrelated, same as e.g https://ci.swift.org/job/pr-swift-build-linux/1074/

hamishknight:
> Just checking I understand this - we override the build parameters for the host tool and then those parameters are passed down to all its dependencies and thus the platform check in isTargetSuitableForPlatformForIndex still passes?

Yes that's correct

hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

jakepetroules:
@swift-ci test linux

jakepetroules:
Ignore Focal CI failures, that was removed in #676.

owenv (approved):

PR #673, opened 2025-07-25
Fix hang in ServiceConsoleTests.serviceShutdown

jakepetroules:
This hang occurred only in CI environments and only on Linux. Here's the sequence of events:

- Test terminates swbuild using SIGKILL
- OS reparents SWBBuildService (a subprocess of swbuild) to launchd (Darwin) / init (others)
- OS closes the file descriptors for the I/O pipes swbuild has connected to SWBBuildService
- SWBBuildService's read() loop indicates EOF due to the broken pipe
- SWBBuildService causes itself to exit

At this point, the getpgid loop should return ERSCH and terminate the test. However, SWBBuildService is sticking around as a zombie for an extended period of time without init reaping the pid, causing getpgid to never hit the termination state. This causes the test to hang indefinitely.

To fix this, there are two aspects:

- A timeout is added around the termination monitoring loop that forces the exit promise to be fulfilled with an error if a 30-second interval elapses without the process exiting
- We switch from using a getpgid loop to using a waitid loop, where the terminal state is that the process has _exited_... we don't care if the zombie hasn't been collected by init, only that it's not in a running state

This fixes the hang for both the Jenkins based CI as well as GitHub actions, and also insulates us against future hangs by ensuring the test will terminate with a timeout error instead of hanging indefinitely, so that we at least know _which_ test is the problem.

### DIFF
--- a/Tests/SwiftBuildTests/ConsoleCommands/ServiceConsoleTests.swift
+++ b/Tests/SwiftBuildTests/ConsoleCommands/ServiceConsoleTests.swift
@@ -19,7 +19,13 @@ import SWBUtil
 import WinSDK
 #endif
 
-@Suite(.skipInGitHubActions("failing in the GitHub actions runner environment"), .skipHostOS(.windows))
+#if canImport(System)
+import System
+#else
+import SystemPackage
+#endif
+
+@Suite(.skipHostOS(.windows))
 fileprivate struct ServiceConsoleTests {
     @Test
     func emptyInput() async throws {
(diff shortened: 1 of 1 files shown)


jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

Tests/SwiftBuildTests/ConsoleCommands/ServiceConsoleTests.swift
```diff
@@ -19,7 +19,13 @@ import SWBUtil
...
 #endif
 
-@Suite(.skipInGitHubActions("failing in the GitHub actions runner environment"), .skipHostOS(.windows))
+#if canImport(System)
+import System
+#else
+import SystemPackage
+#endif
+
+@Suite(.skipHostOS(.windows))
```
cmcgee1024, line 28: question: Will this cause the entire suite to be skipped on Windows, including individual test functions that aren't marked as skip for Windows?

jakepetroules: Yes, but note that this isn't new code. Many of these were broken on Windows which is why they were skipped en masse. We should work on getting them passing.

PR #679, opened 2025-07-27
[SWUtil] PropertyList: Clarify use of `init` overloads that take arra…

xedin:
…y and dictionary

The original comment that these overloads can be removed was incorrect because the existential of `PropertyListItemConvertible` doesn't self conform. The overloads are still needed to support collections with heterogeneous values but nothing else and so they should be disfavored.

Turns out that some of the expressions that used to take advantage of these overloads only work due to the solver producing a valid solution from the "diagnostic" mode.

For example:

```
public var propertyListItem: PropertyListItem {
    return .init(entries
        .sorted { $0.identifier < $1.identifier }
        .map {
            [
                "bundle-id": .plString($0.identifier),
                "tags": PropertyListItem($0.tags.sorted()),
                "bundle-path": .plString($0.path.str),
            ]
        })
}
```

This expression should have been ambiguous because it can match both `.init(any ...)` and `.init([any ...])` overloads due to an existential conversion that is bi-directional for the closure.

For `.init(any ...)`, result of the closure is going to be erased to existential inside of the body of the `.map` closure and for `.init([any ...])`, this erasure is going to happen outside but both are equally valid.


xedin:
@swift-ci please test

xedin:
@swift-ci please test

xedin:
@swift-ci please test Linux

jakepetroules (approved):

xedin:
@swift-ci please test Linux

jakepetroules:
The Linux job is broken; you need to wait until #673 is merged to unblock the CI.

xedin:
Sounds good, there is no rush with this.

neonichu (approved):

jakepetroules:
@swift-ci please test Linux

PR #682, opened 2025-07-29
Pr/update usage description keys

baujla:
Many usage strings are missing from Swift Build. This means that if a developer needs to configure one of these, or if Xcode's signing editor in Xcode 26 adds a usage string, they need to directly add an info plist file. This is a low risk change that just adds data to the existing pattern of defining usage strings.


jakepetroules:
@swift-ci test

PR #686, opened 2025-07-29
[Swift] Add crash reproducer support

cachemeifyoucan:
Add support for swift crash reproducer so a crash reproducer is created if supported when a swift compiler crash happened.

### DIFF
--- a/Sources/SWBCore/LibSwiftDriver/PlannedBuild.swift
+++ b/Sources/SWBCore/LibSwiftDriver/PlannedBuild.swift
@@ -529,6 +529,16 @@ extension LibSwiftDriver {
             }
         }
 
+        public func getCrashReproducerCommand(for job: PlannedSwiftDriverJob, output dir: Path) async throws -> [String]? {
+            try await dispatchQueue.sync {
+                let driverJob = try self.driverJob(for: job)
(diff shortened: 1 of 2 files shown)


cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:
@swift-ci please test

Sources/SWBCore/LibSwiftDriver/PlannedBuild.swift
```diff
@@ -529,6 +529,16 @@ extension LibSwiftDriver {
             }
         }
 
+        public func getCrashReproducerCommand(for job: PlannedSwiftDriverJob, output dir: Path) throws -> [String]? {
+            try dispatchQueue.blocking_sync {
```
jakepetroules: `getCrashReproducerCommand` is only called from asynchronous contexts, so you should use `sync` instead of `blocking_sync`, and turn this into an async function. Whenever possible, we want to avoid blocking the Swift Concurrency thread pool.

neonichu (approved):

cachemeifyoucan:
@swift-ci please test

owenv (approved):

cachemeifyoucan:
@swift-ci please test macOS platform

cachemeifyoucan:
@jakepetroules Ping~

PR #698, opened 2025-08-01
Allow command line tool specs to be extended by BuildSettingsExtension specs and an additional task producer extension point

jakepetroules:
This can allow more decoupling of platform specific concepts and the like from the core build specs.


jakepetroules:
@swift-ci test

PR #702, opened 2025-08-04
Handle the OS version in the triple for FreeBSD

jakepetroules:
We're going to keep the versioned triple, so it needs to be passed through after all.


jakepetroules:
@swift-ci test

neonichu (approved):

jakepetroules:
@swift-ci test

PR #705, opened 2025-08-04
Generalize relocating source-generating resource files to the sources phase

matthewseaman:
Continue decoupling Asset Catalog and String Catalog symbol generation from the core build system by automatically moving Resource build phase files to the Sources build phase when they generate Swift code.


matthewseaman:
@swift-ci please test

PR #718, opened 2025-08-08
Minor modifications to Process._getOutput to avoid extra Task creation

rconnell9:
Minor modifications to Process._getOutput to avoid extra Task creation


rconnell9:
@swift-ci please test

PR #722, opened 2025-08-12
Allow opting into Swift Testing entrypoints for macOS test bundles when building with the SwiftPM CLI

owenv:
SwiftPM has historically generated entrypoints for test bundles on macOS in order to support using a copy of the Swift Testing library other than the one shipped in Xcode. Allow builds to opt into this behavior using a GENERATE_TEST_ENTRYPOINTS_FOR_BUNDLES override which is off by default.


owenv:
@swift-ci test

grynspan:
When building with the SwiftPM CLI, this functionality needs to be enabled. It should not be off by default.

owenv:
@grynspan and I talked offline - this is off by default in Swift Build but SwiftPM CLI will unconditionally enable it

grynspan:
All good, ignore me!

owenv:
@swift-ci test

owenv:
@swift-ci test

PR #724, opened 2025-08-13
Alter usage of async withExtendedLifetime that is causing crash in swift concurrency

rconnell9:
Alter usage of async withExtendedLifetime that is causing crash in swift concurrency


rconnell9:
@swift-ci test

PR #725, opened 2025-08-13
Fix crash when running swift test for swift-build

rconnell9:
Fix crash when running swift test for swift-build

### DIFF
--- a/Sources/SWBUtil/Process.swift
+++ b/Sources/SWBUtil/Process.swift
@@ -82,33 +82,42 @@ extension Process {
 extension Process {
     public static func getOutput(url: URL, arguments: [String], currentDirectoryURL: URL? = nil, environment: Environment? = nil, interruptible: Bool = true) async throws -> Processes.ExecutionResult {
         if #available(macOS 15, iOS 18, tvOS 18, watchOS 11, visionOS 2, *) {
+            let stdoutPipe = Pipe()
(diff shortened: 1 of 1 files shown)


rconnell9:
@swift-ci test

jakepetroules:
@swift-ci test

owenv (approved):

jakepetroules:
> Ahh, nice catch. I think it's a compiler bug that `withExtendedLifetime` can contain async code despite its closure parameter not being annotated `async`?
> 
> Do we have a bug on the compiler to diagnose this as a compile-time error instead?

I mis-remembered, I was thinking of `withoutActuallyEscaping`. The async `withExtendedLifetime` is something coming from Swift Build's own codebase, in Misc+Async.swift:

```swift
@_alwaysEmitIntoClient
public func withExtendedLifetime<T: ~Copyable, E: Error, Result: ~Copyable>(
    _ x: borrowing T,
    _ body: (borrowing T) async throws(E) -> Result
) async throws(E) -> Result {
    defer { withExtendedLifetime(x) { } }
    return try await body(x)
}
```

(making this crash even more puzzling, since this patch is effectively just inlining the code)

ktoso:
Yeah it's a bit puzzling but it seems something about this pattern was upsetting 🤔 
Might be worth a follow up investigation in Swift itself

PR #728, opened 2025-08-13
[6.2.0] Fix crash when running swift test for swift-build

rconnell9:
Fix crash when running swift test for swift-build.

Cherry-pick from: 35c3303


owenv:
@swift-ci test

owenv (approved):

PR #735, opened 2025-08-14
Add VALIDATE_DEPENDENCIES_DOWNGRADE_ERRORS build setting

bob-wilson:
When VALIDATE_MODULE_DEPENDENCIES or VALIDATE_HEADER_DEPENDENCIES is set to YES_ERROR, this new setting can downgrade those errors to warnings. Since the dependency validation is a new feature that is still evolving, we need a way to keep it from breaking builds in some contexts.

rdar://157863457


bob-wilson:
@swift-ci test

neonichu (approved):

jakepetroules (approved):

PR #746, opened 2025-08-21
Use `Invoke-Program` for the Windows build command

bnbarham:
Windows does not stop scripts when native commands exit with non-zero. Instead, their exit status has to be checked manually. This can be done through an `Invoke-Program` function that is added to the script running `windows_build_command`, which also previously prefixed the given command. This is changing in
https://github.com/swiftlang/github-workflows/pull/154 since it doesn't help with multi-line commands - update our modified `windows_build_command` to use `Invoke-Program` instead.


neonichu (approved):

jakepetroules:
@swift-ci test

jakepetroules (approved):

PR #748, opened 2025-08-25
[TEST-ONLY][swift][Caching] Fix some swift caching tests

cachemeifyoucan:
Fix two problems from now enabled swift caching tests:
* Some destinations/requirements are not consistent across the test so the test can fail with certain test configuration.
* Make the test more reliable for different SDK configurations by abstracting out the number of module dependencies.

rdar://158600194


cachemeifyoucan:
@swift-ci please test

PR #753, opened 2025-08-29
Remove FILETextOutputStream, etc.

jakepetroules:
This needed to be tweaked recently to migrate SWBUtil to Swift 6, and caused a build breakage in the process. Turns out this is dead code and not needed, so just eliminate it.


jakepetroules:
@swift-ci test

PR #755, opened 2025-08-29
Add some documentation on the Windows compiler flags

jakepetroules:
Thought this would be useful to document as it's somewhat complex.

### DIFF
--- a/Sources/SWBWindowsPlatform/README.md
+++ b/Sources/SWBWindowsPlatform/README.md
@@ -1,0 +1,56 @@
+# Windows Platform
+
+In most cases, the clang and swift compilers will automatically detect the requisite components installed on the system by searching the environment and/or registry. However, Swift Build contains its own lookup logic and explicitly passes paths and versions of dependencies to the compiler in order to have more direct control over the build process and version selection.
+
+## System Requirements
+
+There are a number of components required for building for Windows...
+
+### Visual C++
+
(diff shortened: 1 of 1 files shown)


jakepetroules:
@swift-ci test

neonichu (approved):

Sources/SWBWindowsPlatform/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+In most cases, the clang and swift compilers will automatically detect the requisite components installed on the system by searching the environment and/or registry. However, Swift Build contains its own lookup logic and explicitly passes paths and versions of dependencies to the compiler in order to have more direct control over the build process and version selection.
+
+## System Requirements
+
+There are a number of components required for building for Windows...
+
+### Visual C++
+
+The most basic requirement is the Visual C++ compiler and runtime. To get this, you must install any edition of Visual Studio (Build Tools, Community, Professional, Enterprise). The Visual Studio installer also places a tool called vswhere at the fixed location of `%PROGRAMFILES(X86)%\Microsoft Visual Studio\Installer\vswhere.exe`, which Swift Build uses to look up the Visual Studio installations on the system, and their properties.
```
compnerd, line 11: I wonder if we really should be using `vswhere`. Does this path exist on ARM64? What happens if there is a native ARM64 binary? What happens with say WoW? I think that it would be better if we just directly interact with the COM interface.

jakepetroules: > Does this path exist on ARM64?

Yes

> https://github.com/microsoft/vswhere/wiki#installing
> **"This is a fixed location that will be maintained."**

Going to merge this for now since it's just documenting the implementation details of what's in place today. Happy to explore the COM interface for the future if you'd like to help with that, though.

PR #751, opened 2025-08-28
Export swb_stdout and swb_stderr on Windows

Steelskin:
This is preventing the Windows toolchain build from succeeding.

PR #750, opened 2025-08-27
Upgrade SWBUtil to Swift 6

jakepetroules:

PR #757, opened 2025-08-29
Run the CMake smoke test on Windows

jakepetroules:
This would have caught a recent regression (#750 / #751).

### DIFF
--- a/.github/scripts/windows_pre_build.ps1
+++ b/.github/scripts/windows_pre_build.ps1
@@ -1,0 +1,40 @@
+##===----------------------------------------------------------------------===##
+##
+## This source file is part of the Swift open source project
+##
+## Copyright (c) 2025 Apple Inc. and the Swift project authors
+## Licensed under Apache License v2.0 with Runtime Library Exception
+##
+## See http://swift.org/LICENSE.txt for license information
+## See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+##
+##===----------------------------------------------------------------------===##
+
+param (
+    [switch]$SkipAndroid,
+    [switch]$InstallCMake
+)
+
+# winget isn't easily made available in containers, so use chocolatey
+Set-ExecutionPolicy Bypass -Scope Process -Force; [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072; iex ((New-Object System.Net.WebClient).DownloadString('https://community.chocolatey.org/install.ps1'))
+
+if ($InstallCMake) {
+    choco install -y cmake --installargs 'ADD_CMAKE_TO_PATH=System' --apply-install-arguments-to-dependencies
+    choco install -y ninja
+
+    Import-Module $env:ChocolateyInstall\helpers\chocolateyProfile.psm1
+    refreshenv
+
+    # Let swiftc find the path to link.exe in the CMake smoke test
+    $env:Path += ";$(Split-Path -Path "$(& "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe" "-latest" -products Microsoft.VisualStudio.Product.BuildTools -find VC\Tools\MSVC\*\bin\HostX64\x64\link.exe)" -Parent)"
+}
+
+if (-not $SkipAndroid) {
+    choco install android-ndk
+
+    Import-Module $env:ChocolateyInstall\helpers\chocolateyProfile.psm1
+    refreshenv
+
+    # Work around a bug in the package causing the env var to be set incorrectly
+    $env:ANDROID_NDK_ROOT = $env:ANDROID_NDK_ROOT.replace('-windows.zip','')
+}
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -29,7 +29,10 @@ jobs:
       linux_pre_build_command: SKIP_ANDROID=1 INSTALL_CMAKE=1 ./.github/scripts/linux_pre_build.sh
       linux_build_command: 'swift package -Xbuild-tools-swiftc -DUSE_PROCESS_SPAWNING_WORKAROUND cmake-smoke-test --disable-sandbox --cmake-path `which cmake` --ninja-path `which ninja` --extra-cmake-arg -DCMAKE_C_COMPILER=`which clang` --extra-cmake-arg -DCMAKE_CXX_COMPILER=`which clang++` --extra-cmake-arg -DCMAKE_Swift_COMPILER=`which swiftc`'
       linux_swift_versions: '["nightly-main"]'
-      windows_swift_versions: '[]'
+      windows_pre_build_command: 'Invoke-Program .\.github\scripts\windows_pre_build.ps1 -SkipAndroid -InstallCMake'
+      windows_swift_versions: '["nightly-main"]'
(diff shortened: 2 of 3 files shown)


jakepetroules:
@compnerd Any idea what's up with this?

> LINK : fatal error LNK1104: cannot open file 'libTSCLibc.lib'

It seems TSC can't build itself because TSCBasic is trying to link libTSCLibc.lib when it should actually be TSCLibc.lib, but the wrong spelling is not referenced anywhere on the command line or in response files! Does it somehow get embedded in the .obj files by the compiler?

How is Swift CI currently building TSC successfully with CMake?

compnerd:
> @compnerd Any idea what's up with this?
> 
> > LINK : fatal error LNK1104: cannot open file 'libTSCLibc.lib'
> 
> It seems TSC can't build itself because TSCBasic is trying to link libTSCLibc.lib when it should actually be TSCLibc.lib, but the wrong spelling is not referenced anywhere on the command line or in response files! Does it somehow get embedded in the .obj files by the compiler?
> 
> How is Swift CI currently building TSC successfully with CMake?

Yes, the spelling is embedded into the object file by the compiler. The new prefix of `lib` for the static libraries is only available as of 6.2 I believe. For the time being, you can remove `-D CMAKE_STATIC_LIBRARY_PREFIX_Swift=lib` until we update to a new enough compiler where it will break in the inverse way: `libTSCLibc.lib not found` as the file on disk will be `TSCLibc.lib` as written out by CMake/Ninja, but the compiler expects it to be `libTSCLibc.lib` to indicate that that it is a (Swift) static library.

jakepetroules:
> Yes, the spelling is embedded into the object file by the compiler. The new prefix of lib for the static libraries is only available as of 6.2 I believe

This is all with nightly (6.3) toolchains...

> remove -D CMAKE_STATIC_LIBRARY_PREFIX_Swift=lib 

"remove"? I was able to get past this error by _adding_ that flag. 

That said, I now get this:

```
clang++: error: no such file or directory: '\INCREMENTAL:NO'
```

...due to this flag being passed manually [here](https://github.com/swiftlang/swift-llbuild/blob/main/products/libllbuild/CMakeLists.txt#L25) and [here](https://github.com/swiftlang/swift-llbuild/blob/main/products/llbuild/CMakeLists.txt#L14). Should we actually remove those? It seems `/INCREMENTAL:NO` makes more sense to pass for the build as a whole (which Swift CI does) rather than having it on individual targets. These flags were added by swiftlang/swift-llbuild#475 back in 2019. The justification was "Disable incremental linking so llbuild.ilk's don't conflict", which made sense as the library and tool had the same OUTPUT_NAME back then, which you recently fixed in swiftlang/swift-llbuild#1005, so I think that workaround seems no longer necessary.

If I instead set CLANG_C_COMPILER / CLANG_CXX_COMPILER to clang-cl to work around the unrecognized flags issue, I'm getting this:

```
error: processError(terminationReason: Foundation.Process.TerminationReason.exit, terminationStatus: 1)
C:\Windows\system32\cmd.exe /C "cd . && C:\Users\ContainerAdministrator\AppData\Local\Programs\Swift\Toolchains\0.0.0+Asserts\usr\bin\swiftc.exe -j 4 -num-threads 4 -emit-library -module-cache-path C:\source\.build\plugins\cmake-smoke-test\outputs\module-cache -Onone -g -libc MD  /machine:x64 /debug /INCREMENTAL -cxx-interoperability-mode=default -Xlinker -implib:lib\SWBCSupport.lib   -o bin\SWBCSupport.dll Sources\SWBCSupport\CMakeFiles\SWBCSupport.dir\CLibclang.cpp.obj Sources\SWBCSupport\CMakeFiles\SWBCSupport.dir\CLibRemarksHelper.c.obj Sources\SWBCSupport\CMakeFiles\SWBCSupport.dir\empty.swift.obj  -lBlocksRuntime.lib && cd ."
clang++: error: no such file or directory: '\machine:x64'
clang++: error: no such file or directory: '\debug'
clang++: error: no such file or directory: '\INCREMENTAL'
```

jakepetroules:
It seems fairly clear that those llbuild flags should be removed, so I went ahead and put that up for review: swiftlang/swift-llbuild#1011

compnerd:
> > Yes, the spelling is embedded into the object file by the compiler. The new prefix of lib for the static libraries is only available as of 6.2 I believe
> 
> This is all with nightly (6.3) toolchains...
> 
> > remove -D CMAKE_STATIC_LIBRARY_PREFIX_Swift=lib
> 
> "remove"? I was able to get past this error by _adding_ that flag.

Ah, okay. I always get confused about the directionality :(. Its just a question of matching the output from CMake to match what the compiler wants.

> That said, I now get this:
> 
> ```
> clang++: error: no such file or directory: '\INCREMENTAL:NO'
> ```
> 
> ...due to this flag being passed manually [here](https://github.com/swiftlang/swift-llbuild/blob/main/products/libllbuild/CMakeLists.txt#L25) and [here](https://github.com/swiftlang/swift-llbuild/blob/main/products/llbuild/CMakeLists.txt#L14). Should we actually remove those? It seems `/INCREMENTAL:NO` makes more sense to pass for the build as a whole (which Swift CI does) rather than having it on individual targets. These flags were added by [swiftlang/swift-llbuild#475](https://github.com/swiftlang/swift-llbuild/pull/475) back in 2019. The justification was "Disable incremental linking so llbuild.ilk's don't conflict", which made sense as the library and tool had the same OUTPUT_NAME back then, which you recently fixed in [swiftlang/swift-llbuild#1005](https://github.com/swiftlang/swift-llbuild/pull/1005), so I think that workaround seems no longer necessary.

We need to disable incremental linking as that will break everything as the Swift runtime is unable to deal with the additional padding in the protocol tables which is required for incremental linking. We should be passing `/INCREMENTAL:NO` as `-Xlinker /INCREMENTAL:NO`. This is just a matter of missing `-Xlinker` prefixes (this is behavioural differences between various CMake releases).

> If I instead set CLANG_C_COMPILER / CLANG_CXX_COMPILER to clang-cl to work around the unrecognized flags issue, I'm getting this:
> 
> ```
> error: processError(terminationReason: Foundation.Process.TerminationReason.exit, terminationStatus: 1)
> C:\Windows\system32\cmd.exe /C "cd . && C:\Users\ContainerAdministrator\AppData\Local\Programs\Swift\Toolchains\0.0.0+Asserts\usr\bin\swiftc.exe -j 4 -num-threads 4 -emit-library -module-cache-path C:\source\.build\plugins\cmake-smoke-test\outputs\module-cache -Onone -g -libc MD  /machine:x64 /debug /INCREMENTAL -cxx-interoperability-mode=default -Xlinker -implib:lib\SWBCSupport.lib   -o bin\SWBCSupport.dll Sources\SWBCSupport\CMakeFiles\SWBCSupport.dir\CLibclang.cpp.obj Sources\SWBCSupport\CMakeFiles\SWBCSupport.dir\CLibRemarksHelper.c.obj Sources\SWBCSupport\CMakeFiles\SWBCSupport.dir\empty.swift.obj  -lBlocksRuntime.lib && cd ."
> clang++: error: no such file or directory: '\machine:x64'
> clang++: error: no such file or directory: '\debug'
> clang++: error: no such file or directory: '\INCREMENTAL'
> ```

I would say just write this as `-Xlinker /machine:x64 -Xlinker /debug -Xlinker /INCREMENTAL:NO`.

We should pass in `-g -debug-info-format=codeview` to generate the proper debug information, or we should consider dropping `-Xlinker /debug` if we do not want to generate debug information.

jakepetroules:
@swift-ci test

owenv (approved):

PR #758, opened 2025-08-29
Refactor the plugin loading a bit to minimize compile-time differences

jakepetroules:
This narrows what's guarded by USE_STATIC_PLUGIN_INITIALIZATION as tightly as possible, to avoid unintentional build breakages during refactoring, for clients where this flag is turned off. It also fixes an issue which resulted from this during Core initialization where the wrong type of plugin manager was being passed into the plugin initialization function.


jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

PR #759, opened 2025-08-30
Assert that no two message types with the same name are registered

ahoppen:
This simplifies debugging issues that arise from accidentally trying to register multiple message types with the same name.

### DIFF
--- a/Sources/SWBProtocol/Message.swift
+++ b/Sources/SWBProtocol/Message.swift
@@ -1222,6 +1222,14 @@ public struct IPCMessage: Serializable, Sendable {
     /// Reverse name mapping.
     static let messageNameToID: [String: any Message.Type] = {
         var result = [String: any Message.Type]()
+        #if DEBUG
+        var seenMessageNames: Set<String> = []
+        for messageType in messageTypes {
+            if !seenMessageNames.insert(messageType.name).inserted {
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

owenv (approved):

Sources/SWBProtocol/Message.swift
```diff
@@ -1222,6 +1222,14 @@ public struct IPCMessage: Serializable, Sendable {
...
+        #if DEBUG
+        var seenMessageNames: Set<String> = []
+        for messageType in messageTypes {
+            if !seenMessageNames.insert(messageType.name).inserted {
+                assertionFailure("Multiple message types registered for same name: \(messageType.name)")
+            }
+        }
+        #endif
         for type in IPCMessage.messageTypes {
             result[type.name] = type
```
jakepetroules, line 1234: @ahoppen  nit: I would have suggested just doing it here, unconditionally (not behind DEBUG), rather than having a second loop:

```swift
if let oldValue = result.updateValue(type, forKey: type.name) {
    fatalError("Multiple message types registered for same name: \(type.name)")
}
```

ahoppen: Here we go: https://github.com/swiftlang/swift-build/pull/764

PR #760, opened 2025-08-30
Add request to get preferred run destination for a platform

ahoppen:
This can be useful when implementing a system that needs to configure the active run destination of a build request but only knows the platform that the user is requesting a build for, but no architecture etc.

### DIFF
--- a/Sources/SWBBuildService/Messages.swift
+++ b/Sources/SWBBuildService/Messages.swift
@@ -1528,6 +1528,31 @@ private struct ClearAllCaches: MessageHandler {
     }
 }
 
+private struct GetPlatformPreferredRunDestinationMsg: MessageHandler {
+    struct PlatformNotFoundError: Error, CustomStringConvertible {
+        let platform: String
+        var description: String { "Unknown platform '\(platform)'" }
+    }
+    struct UnknownPreferredArchError: Error, CustomStringConvertible {
+        let platform: String
+        var description: String { "Platform '\(platform)' does not specify a preferred architecture" }
+    }
+
+    func handle(request: Request, message: GetPlatformPreferredRunDestinationRequest) async throws -> GetPlatformPreferredRunDestinationResponse {
+        let session = try request.session(for: message)
+        guard let platform = session.core.platformRegistry.lookup(name: message.platform) else {
+            throw PlatformNotFoundError(platform: message.platform)
+        }
+        // All relevant platforms define a preferredArch, so the undefined_arch fallback case should never happen
+        // in practice, and indicates a serious issue occurred during plugin loading.
+        guard let targetArchitecture = platform.preferredArch else {
+            throw UnknownPreferredArchError(platform: message.platform)
+        }
+        let runDestination = RunDestinationInfo(platform: platform.name, sdk: platform.name, sdkVariant: nil, targetArchitecture: targetArchitecture, supportedArchitectures: [targetArchitecture], disableOnlyActiveArch: false, hostTargetedPlatform: nil)
+        return GetPlatformPreferredRunDestinationResponse(info: runDestination)
+    }
+}
+
 // MARK: ServiceExtension Support
 
 public struct ServiceSessionMessageHandlers: ServiceExtension {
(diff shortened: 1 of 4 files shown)


Sources/SwiftBuild/SWBBuildServiceSession.swift
```diff
@@ -582,6 +582,10 @@ public final class SWBBuildServiceSession: Sendable {
         try await service.send(request: DeveloperPathRequest(sessionHandle: uid)).value
     }
 
+    public func preferredRunDestination(forPlatform platformName: String) async throws -> RunDestinationInfo? {
```
jakepetroules: This should return an `SWBRunDestinationInfo`, which is the public API cover type for `RunDestinationInfo`.

Sources/SwiftBuild/SWBBuildServiceSession.swift
```diff
@@ -12,7 +12,7 @@
 
 import Foundation
 
-import SWBProtocol
+public import SWBProtocol
```
jakepetroules: Not needed. (see https://github.com/swiftlang/swift-build/pull/760/files#r2312690176)

Sources/SWBBuildService/Messages.swift
```diff
@@ -1528,6 +1528,25 @@ private struct ClearAllCaches: MessageHandler {
...
+    }
+
+    func handle(request: Request, message: GetPlatformPreferredRunDestinationRequest) async throws -> GetPlatformPreferredRunDestinationResponse {
+        let session = try request.session(for: message)
+        guard let platform = session.core.platformRegistry.lookup(name: message.platform) else {
+            throw PlatformNotFoundError(platform: message.platform)
+        }
+        // All relevant platforms define a preferredArch, so the undefined_arch fallback case should never happen
+        // in practice, and indicates a serious issue occurred during plugin loading.
+        let targetArchitecture = platform.preferredArch ?? "undefined_arch"
```
jakepetroules: Maybe just return nil / throw error in this case? since "undefined_arch" isn't actually usable

Sources/SwiftBuild/SWBBuildServiceSession.swift
jakepetroules: Also, why can it return nil? It doesn't seem like it can in practice.

jakepetroules (commented):
Seems reasonable to me, if the API contract is basically "give me a suggested destination given the specified input parameters", without defining/promising the semantics used to define that.

Sources/SwiftBuild/SWBBuildServiceSession.swift
ahoppen: OK, I see. I wasn’t sure what the rationale for not using the protocol types in `SwiftBuild` was and tried to avoid unnecessary work but now using `SWBRunDestinationInfo`. And good catch with `nil`.

Sources/SWBBuildService/Messages.swift
ahoppen: Good idea, I copied this code from https://github.com/swiftlang/swift-build/blob/f59c406e5ed0d3c2395175c6717d571b6fdfd39e/Sources/SWBTaskExecution/ProjectPlanner.swift#L136-L139 without many second thoughts.

ahoppen:
> Seems reasonable to me, if the API contract is basically "give me a suggested destination given the specified input parameters", without defining/promising the semantics used to define that.

That’s the idea. Do you think this needs to be clarified somewhere in a doc comment or is it clean enough from the current ones?

ahoppen:
Added a few test cases as well based on Owen’s comment here: https://github.com/swiftlang/swift-build/pull/761. Would be good to get a review for those as well.

ahoppen:
@swift-ci Please test

jakepetroules:
@swift-ci test

Sources/SwiftBuild/SWBBuildServiceSession.swift
jakepetroules: You still need to revert this part.

ahoppen: Thanks for catching this. I completely forgot about it.

ahoppen:
@swift-ci Please test

PR #761, opened 2025-08-30
Add requests that can be used to implement a BSP server based on swift-builds build description

ahoppen:
The design of these requests closely resemble the BSP requests required by SourceKit-LSP and should be useful to implement a BSP server based on swift-build.

Open tasks: 
- [ ] I didn’t find a way to test my new code in `SWBBuildService`. If there is a way to test it, I’d appreciate some guidance
- [ ] The new requests currently assume that the build description with the given ID hasn’t been purged from the `BuildDescriptionManager` but there’s no way to ensure that. I was thinking to add a request to retain or release a build description. As long as the build description has a retain count greater than 0, it won’t be purged from the `BuildDescriptionManager`. But I’d like to hear feedback about this idea before implementing it.


Sources/SwiftBuild/SWBBuildServiceSession.swift
```diff
@@ -12,8 +12,8 @@
 
 import Foundation
 
-import SWBProtocol
-import SWBUtil
+public import SWBProtocol
```
owenv: Generally we've tried not to expose SWBProtocol and SWBUtil as public imports, though here it might require duplicating some of the response types

owenv:
> The new requests currently assume that the build description with the given ID hasn’t been purged from the BuildDescriptionManager but there’s no way to ensure that. I was thinking to add a request to retain or release a build description. As long as the build description has a retain count greater than 0, it won’t be purged from the BuildDescriptionManager. But I’d like to hear feedback about this idea before implementing it.

Right now we will never evict the most recent index workspace description from the cache, but it would be nice to have something better. Being able to treat the build description ID as a handle clients can retain makes sense to me

owenv:
> I didn’t find a way to test my new code in SWBBuildService. If there is a way to test it, I’d appreciate some guidance

I think e can put a test for the new API in the SwiftBuildTests target. IndexingInfoTests.indexingInfoDetailed is a good example of an existing test that stands up a test session with a mock project. I think it should be possible to copy that and adapt it to call into the new API you've added here

Sources/SwiftBuild/SWBBuildServiceSession.swift
jakepetroules: I think it's a hard requirement that we don't expose them.

ahoppen:
@swift-ci Please test

ahoppen:
I added a few tests now. Before being able to use this in production, I think we need a way to create build descriptions and keep them in memory without evicting them from the cache but I’d like to make that a follow-up change.

owenv (approved):

neonichu (approved):

Sources/SWBBuildService/BuildDescriptionMessages.swift
```diff
@@ -0,0 +1,218 @@
...
+        )
+
+        let session = try request.session(for: message)
+
+        let targetInfos = buildDescription.allConfiguredTargets.map { configuredTarget in
+            let toolchain: Path?
+            if let toolchainID = toolchainIDs(in: configuredTarget, of: buildDescription)?.first {
+                toolchain = session.core.toolchainRegistry.lookup(toolchainID)?.path
+                if toolchain == nil {
+                    log("Unable to find path for toolchain with identifier \(toolchainID)", isError: true)
```
jakepetroules, line 110: @ahoppen We don't conventionally have this level of logging in the product; it seems like debug printing that should've been removed.

Sources/SWBBuildService/BuildDescriptionMessages.swift
```diff
@@ -0,0 +1,218 @@
...
+    typealias SourceFileInfo = BuildDescriptionConfiguredTargetSourcesResponse.SourceFileInfo
+    typealias ConfiguredTargetSourceFilesInfo = BuildDescriptionConfiguredTargetSourcesResponse.ConfiguredTargetSourceFilesInfo
+
+    func handle(request: Request, message: BuildDescriptionConfiguredTargetSourcesRequest) async throws -> BuildDescriptionConfiguredTargetSourcesResponse {
+        let buildDescription = try await request.buildDescription(for: message)
+
+        let configuredTargetsByID = Dictionary(
+            buildDescription.allConfiguredTargets.map { ($0.guid, $0) }
+        ) { lhs, rhs in
+            log("Found conflicting targets for the same ID: \(lhs.guid)", isError: true)
```
jakepetroules, line 159: @ahoppen These type of scenarios should throw an error, not "pick one and log"; it tends to lead to unpredictable and nondeterministic behavior

jakepetroules:
Why is it the right approach to add more low-level requests at the Swift Build layer for a higher level layer to implement the BSP, versus Swift Build itself hosting the BSP server?

PR #763, opened 2025-09-01
WIP: Prototype for a `TemporaryDirectoryTrait` and `AsyncDeferrableTrait`

ahoppen:
Depends on https://github.com/swiftlang/swift-build/pull/760.

---

I just want to raise the idea of using test traits over `with` function to avoid nesting levels due to `with*` functions in tests.

### DIFF
--- a/Sources/SWBBuildService/Messages.swift
+++ b/Sources/SWBBuildService/Messages.swift
@@ -1528,6 +1528,31 @@ private struct ClearAllCaches: MessageHandler {
     }
 }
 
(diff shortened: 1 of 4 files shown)


jakepetroules:
I kind of like this, but at the same time it would be a lot of effort to go change it all..

bnbarham:
> but at the same time it would be a lot of effort to go change it all..

I volunteer as tribute if it means we could switch to this. Could also just do it one test file at a time or something similar 😅

PR #764, opened 2025-09-01
Do not guard check for duplicate message names behind `DEBUG` compilation condition

ahoppen:
As suggested in https://github.com/swiftlang/swift-build/pull/759#discussion_r2312689408


owenv (approved):

jakepetroules (approved):

jakepetroules:
@swift-ci test

PR #766, opened 2025-09-02
Skip XCTest discovery if tests are not built with -enable-testing

owenv:
Otherwise, entry point code will be generated using @testable imports and fail to compile.


neonichu (approved):

owenv:
@swift-ci test

jakepetroules (approved):

PR #768, opened 2025-09-02
Properly quote linker file list contents on Linux and Windows

owenv:


owenv:
@swift-ci test

jakepetroules:
This really needs to depend on the specific linker in use; what about cross compilation scenarios where we are using lld on macOS or Windows?

owenv:
> This really needs to depend on the specific linker in use; what about cross compilation scenarios where we are using lld on macOS or Windows?

The general idea here is that this change establishes the correct platform-default behavior, and then we'll need to inspect -fuse-ld flags at some point to override that behavior when manually selecting a different linker

owenv:
@swift-ci test

owenv:
@swift-ci test

Sources/SWBUtil/ArgumentSplitting.swift
```diff
@@ -280,3 +280,39 @@ public final class LLVMStyleCommandCodec: CommandSequenceEncodable, CommandSeque
     private static let space = Character(" ")
     private static let dollarSign = Character("$")
 }
+
+/// Suitable for escaping Windows process arguments, but NOT command lines which will be interpreted by a shell
```
jakepetroules, line 284: Excellent call out :)

PR #771, opened 2025-09-03
Dependencies: handle empty clang trace (rdar://159767983)

mirza-garibovic:
Used multiple commits to factor out the common code here.


mirza-garibovic:
@swift-ci test

PR #776, opened 2025-09-04
Add a mechanism to clear the global module cache at the end of a build for testing

owenv:
This is similar to the existing setting which allows clearing the global compiler cache


Sources/SWBBuildSystem/BuildOperation.swift
```diff
@@ -1601,6 +1602,33 @@ internal final class OperationSystemAdaptor: SWBLLBuild.BuildSystemDelegate, Act
...
+        signatureCtx.add(string: "CleanupGlobalModuleCache")
+        signatureCtx.add(string: cachePath.str)
+        let signature = signatureCtx.signature
+
+        withActivity(ruleInfo: "CleanupGlobalModuleCache \(cachePath.str)", executionDescription: "Cleanup global module cache at \(cachePath)", signature: signature, target: nil, parentActivity: nil) { activity in
+            do {
+                try operation.fs.removeDirectory(cachePath)
+            } catch {
+                // Log error but do not fail the build.
+                emit(diagnostic: Diagnostic.init(behavior: .warning, location: .unknown, data: DiagnosticData("Error cleaning up \(cachePath): \(error.localizedDescription)")), for: activity, signature: signature)
```
jakepetroules: Phrase as "failed to remove" rather than "error cleaning up"?

Using the word "error" in a warning message could be confusing.

owenv: sure, makes sense

owenv:
@swift-ci test

PR #777, opened 2025-09-04
Require the macOS SDK for `InspectBuildDescriptionTests`

ahoppen:
We need the macOS SDK in order to generate the build graph containing app targets.


ahoppen:
@swift-ci Please test

PR #778, opened 2025-09-05
Dependencies: support clang modules when parsing dependency validation info

sina-mahdavi:
This PR adds support for the new Clang `-header-include-format=direct-per-file` output, which now tracks Clang modules. The logic should still support the old format for backwards compatibility if it doesn't detect a version number in the trace file. The related LLVM PR is: https://github.com/llvm/llvm-project/pull/156756


sina-mahdavi:
@swift-ci Please test

sina-mahdavi:
@swift-ci Please test

Sources/SWBTaskExecution/TaskActions/ClangCompileTaskAction.swift
```diff
@@ -610,9 +616,65 @@ public final class ClangNonModularCompileTaskAction: TaskAction {
     }
 }
 
+fileprivate func parseTraceData(_ data: Data) throws -> TraceData? {
+    if let jsonObject = try JSONSerialization.jsonObject(with: data) as? [String: Any],
```
jakepetroules: We don't use JSONSerialization directly in Swift Build, as it is Any-typed. When you need to decode _arbitrary_ JSON structures, use:

```swift
if let jsonObject = try PropertyList.fromJSONData(data)
```

...which will give you a strongly-typed JSON object tree. You can then access the version like:

```swift
jsonObject.dictValue?["version"]?.stringValue
```

Sources/SWBTaskExecution/TaskActions/ClangCompileTaskAction.swift
```diff
@@ -610,9 +616,65 @@ public final class ClangNonModularCompileTaskAction: TaskAction {
...
+        return nil
+    } else {
+        // The initial unversioned format (v1) of the trace file generated from clang
+        // is a JSON array so deserializing it as a dictionary will fail.
+        return .V1(try JSONDecoder().decode(TraceData.TraceFileV1.self, from: data))
+    }
+}
+
+fileprivate func parseTraceSourceLocation(_ locationStr: String) -> SWBUtil.Diagnostic.Location {
+    guard let match = locationStr.wholeMatch(of: #/(.+):(\d+):(\d+)/#) else {
```
jakepetroules: Use named capture groups so you're not accessing the captures by index:

```suggestion
    guard let match = locationStr.wholeMatch(of: #/(?<filename>.+):(?<line>\d+):(?<line>\d+)/#) else {
```

then:

```swift
let filename = Path(match.filename)
let line = Int(match.line)
let column = Int(match.column)
```

This also helps avoids accessing the wrong element if the regex adds/removes capture groups over time.

Tests/SWBBuildSystemTests/DependencyValidationTests.swift
```diff
@@ -542,7 +542,7 @@ fileprivate struct DependencyValidationTests: CoreBasedTests {
 
             // Expect complaint about undeclared dependency
             try await tester.checkBuild(parameters: BuildParameters(configuration: "Debug"), runDestination: .host, persistent: true) { results in
-                results.checkError(.contains("Missing entries in MODULE_DEPENDENCIES: Foundation (for task"))
+                results.checkError(.contains("Missing entry in MODULE_DEPENDENCIES: Foundation (for task"))
```
mirza-garibovic, line 545: Hmm why would this change? I thought the top level diag was always pluralized.

sina-mahdavi: The test fails for me locally with "entries", I think it's because the "(for task" substring is at the end of the diagnostic which will have the singular messages. So the whole diagnostic is like:
```
Missing entries in MODULE_DEPENDENCIES: Foundation Foundation
/private/var/folders/5h/9y9ls7ms0fs8gdmlfwklkwpm0000gn/T/swbuild.tmp.x8Tuqtfi/Data.noindex/Test/aProject/Sources/CoreFoo.m: Missing entry in MODULE_DEPENDENCIES: Foundation
/private/var/folders/5h/9y9ls7ms0fs8gdmlfwklkwpm0000gn/T/swbuild.tmp.x8Tuqtfi/Data.noindex/Test/aProject/Sources/CoreBar.m: Missing entry in MODULE_DEPENDENCIES: Foundation (for task: [\"ValidateDependencies\"])
```

If you want I could change this to look for `Missing entries in MODULE_DEPENDENCIES: Foundation Foundation\n`?

bob-wilson: The missing entries are supposed to be de-duplicated (which is why the test checks for "(for task" at the end). It sounds like that's not working.

Sources/SWBTaskExecution/TaskActions/ClangCompileTaskAction.swift
```diff
@@ -483,29 +481,39 @@ public final class ClangCompileTaskAction: TaskAction, BuildValueValidatingTaskA
         isModular: Bool
     ) throws {
         let payload: DependencyValidationInfo.Payload
-        if let traceFilePath {
-            let traceFileContent = try fileSystem.read(traceFilePath)
+        if let traceFilePath,
+           let traceData = try parseTraceData(Data(fileSystem.read(traceFilePath))) {
 
-            let traceData: Array<TraceData>
-            // clang will emit an empty file instead of an empty array when there's nothing to trace
```
bob-wilson, line 492: Did you check that the empty file case works? I don't see an explicit check for that, but maybe it's handled by your parsing code.

sina-mahdavi: You're right I need to check that, I'll add that check.

Tests/SWBBuildSystemTests/DependencyValidationTests.swift
sina-mahdavi: Actually I think there's an issue with the deduplication in my implementation of parsing v1 traces that I need to fix. However, even after fixing that the diagnostic looks like this:
```
Missing entries in MODULE_DEPENDENCIES: Foundation
Missing entry in MODULE_DEPENDENCIES: Foundation (for task: [\"ValidateDependencies\"])
```

owenv:
@swift-ci test

sina-mahdavi:
@swift-ci test

Tests/SWBBuildSystemTests/DependencyValidationTests.swift
sina-mahdavi: Ok I think my last commit should fix the deduplication issue, the first line of the diagnostic has only one mention of `Foundation`. However, `Foundation (for task` is still only in the last line which has singular `entry`.

bob-wilson:
@swift-ci test

PR #779, opened 2025-09-05
Fixup StackedSearchPathLookupError error message

plemarquand:
The target file that could not be found was printing `Path(_str: "libclang.so")` instead of `libclang.so`.

Add the `.str` to the `Path` so just the filename is printed.

### DIFF
--- a/Sources/SWBCore/Settings/StackedSearchPaths.swift
+++ b/Sources/SWBCore/Settings/StackedSearchPaths.swift
@@ -104,7 +104,7 @@ extension StackedSearchPathLookupError: CustomStringConvertible {
         switch self {
         case let .unableToFind(subject, operatingSystem, searchPaths):
             let candidates = searchPaths.flatMap { $0.paths.map { $0.join(subject.fileName(operatingSystem: operatingSystem)).str }}
-            return "unable to find \(subject.fileName(operatingSystem: operatingSystem)) among search paths: \(candidates.joined(separator: ", "))"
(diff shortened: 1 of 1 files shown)


bkhouri (approved):

Sources/SWBCore/Settings/StackedSearchPaths.swift
```diff
@@ -104,7 +104,7 @@ extension StackedSearchPathLookupError: CustomStringConvertible {
         switch self {
         case let .unableToFind(subject, operatingSystem, searchPaths):
             let candidates = searchPaths.flatMap { $0.paths.map { $0.join(subject.fileName(operatingSystem: operatingSystem)).str }}
-            return "unable to find \(subject.fileName(operatingSystem: operatingSystem)) among search paths: \(candidates.joined(separator: ", "))"
+            return "unable to find \(subject.fileName(operatingSystem: operatingSystem).str) among search paths: \(candidates.joined(separator: ", "))"
```
bkhouri, line 107: suggestion: could we add a `small` test to ensure we don't regress here?

owenv (approved):

owenv:
@swift-ci test

Sources/SWBCore/Settings/StackedSearchPaths.swift
jakepetroules: I don't know if it's worth it because the string form of error messages is not considered stable -- if we changed it again we'd just change the test here too and break downstream anyways.

jakepetroules (approved):

PR #780, opened 2025-09-05
Support migratable features when diagnosing adoption

owenv:
If a listed feature isn't boolean, eval it as a string as a fallback and check for a "YES" case

rdar://159684792


owenv:
@swift-ci test

jakepetroules (approved):

PR #781, opened 2025-09-05
Add session API for registering additional toolchains

owenv:
This can be used by SwiftPM to register the toolchain it intends to use with the build service if it's at a nonstandard location

### DIFF
--- a/Sources/SWBBuildService/Messages.swift
+++ b/Sources/SWBBuildService/Messages.swift
@@ -302,6 +302,13 @@ private struct SetSessionUserPreferencesMsg: MessageHandler {
     }
 }
 
+private struct RegisterToolchainHandler: MessageHandler {
+    func handle(request: Request, message: RegisterToolchainRequest) async throws -> StringResponse {
+        let session = try request.session(for: message)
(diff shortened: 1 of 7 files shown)


owenv:
@swift-ci test

owenv:
@swift-ci test

jakepetroules:
Why do we need to register toolchains dynamically versus freezing them at Core initialization time? Can you detail here under what circumstances SwiftPM would use this API?

Also note that there is one Core instance in a service per developer-directory, so they can be shared across multiple sessions, and a service level API would seem more appropriate than a session-level API if we're going to add this functionality.

owenv:
@jakepetroules the main problem I'm trying to solve is that right now, SwiftPM is relying on CC and SWIFT_EXEC overrides instead of properly setting up the toolchainRegistry stack to be `selected-toolchain default-toolchain`. So, given a SwiftPM UserToolchain, we need to 1. register the toolchain if it's in a nonstandard location and 2. determine its identifier to include in TOOLCHAINS.

I could potentially register the toolchain when the service is brought up but it ends up working a little oddly since in theory a service might have >1 Core. I'll explore some alternatives a bit

PR #782, opened 2025-09-08
Expose a low-level BSP interface wrapping SWBBuildServiceSession

owenv:
SWBBuildServer implements BSP message handling and is intended to back a higher level BSP which manages the project model and generates PIF and a build request (SwiftPM).

As part of this change, many of the core BSP types are vendored into the new SWBBuildServerProtocol target from SourceKitLSP with minor modifications. The only significant changes to that code are:
- Replacing some use of atomics with SWBMutex to cut dependencies on C atomics shims since the deployment target is not yet high enough to adopt the atomics in Synchronization
- Removing use of the sourcekit-lsp logging mechanism

The new code in this PR is largely in SWBBuildServer and SWBBuildServerTests

### DIFF
--- a/Sources/SwiftBuild/SWBBuildServer.swift
+++ b/Sources/SwiftBuild/SWBBuildServer.swift
@@ -1,0 +1,521 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+#if !os(iOS)
+import BuildServerProtocol
+public import LanguageServerProtocol
+public import LanguageServerProtocolTransport
+public import ToolsProtocolsSwiftExtensions
+import SWBProtocol
+import SWBUtil
+import Foundation
+
+/// Wraps a `SWBBuildServiceSession` to expose Build Server Protocol functionality.
+public actor SWBBuildServer: QueueBasedMessageHandler {
+    /// The session used for underlying build system functionality.
+    private let session: SWBBuildServiceSession
+    enum PIFSource {
+        // PIF should be loaded from the container at the given path
+        case container(String)
+        // PIF will be transferred to the session externally
+        case session
+    }
+    /// The source of PIF describing the workspace for this build server.
+    private let pifSource: PIFSource
+    /// The build request representing preparation.
+    private let buildRequest: SWBBuildRequest
+    /// The currently planned build description used to fulfill requests.
+    private var buildDescriptionID: SWBBuildDescriptionID? = nil
+
+    private var indexStorePath: String? {
+        buildRequest.parameters.arenaInfo?.indexDataStoreFolderPath.map {
+            Path($0).dirname.join("index-store").str
+        }
+    }
+    private var indexDatabasePath: String? {
+        buildRequest.parameters.arenaInfo?.indexDataStoreFolderPath
+    }
+
+    public let messageHandlingHelper = QueueBasedMessageHandlerHelper(
+        signpostLoggingCategory: "build-server-message-handling",
+        createLoggingScope: false
+    )
+    public let messageHandlingQueue = AsyncQueue<BuildServerMessageDependencyTracker>()
+    /// Used to serialize workspace loading.
+    private let workspaceLoadingQueue = AsyncQueue<Serial>()
+    /// Used to serialize preparation builds, which cannot run concurrently.
+    private let preparationQueue = AsyncQueue<Serial>()
+    /// Connection used to send messages to the client of the build server (an LSP or higher-level BSP implementation).
+    private let connectionToClient: any Connection
+
+    /// Represents the lifetime of the build server implementation..
+    enum ServerState: CustomStringConvertible {
+        case waitingForInitializeRequest
+        case waitingForInitializedNotification
+        case running
+        case shutdown
+
+        var description: String {
+            switch self {
+            case .waitingForInitializeRequest:
+                "waiting for initialization request"
+            case .waitingForInitializedNotification:
+                "waiting for initialization notification"
+            case .running:
+                "running"
+            case .shutdown:
+                "shutdown"
+            }
+        }
+    }
+    var state: ServerState = .waitingForInitializeRequest
+    /// Allows customization of server exit behavior.
+    var exitHandler: (Int) async -> Void
+
+    public static let sessionPIFURI = DocumentURI(.init(string: "swift-build://session-pif")!)
+
+    public init(session: SWBBuildServiceSession, containerPath: String, buildRequest: SWBBuildRequest, connectionToClient: any Connection, exitHandler: @escaping (Int) async -> Void) {
+        self.init(session: session, pifSource: .container(containerPath), buildRequest: buildRequest, connectionToClient: connectionToClient, exitHandler: exitHandler)
+    }
+
+    public init(session: SWBBuildServiceSession, buildRequest: SWBBuildRequest, connectionToClient: any Connection, exitHandler: @escaping (Int) async -> Void) {
+        self.init(session: session, pifSource: .session, buildRequest: buildRequest, connectionToClient: connectionToClient, exitHandler: exitHandler)
+    }
+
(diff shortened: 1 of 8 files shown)


owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test

Sources/SwiftBuild/SWBBuildServer.swift
```diff
@@ -0,0 +1,462 @@
...
+    private let workspaceLoadingQueue = AsyncQueue<Serial>()
+    /// Used to serialize preparation builds, which cannot run concurrently.
+    private let preparationQueue = AsyncQueue<Serial>()
+    /// Connection used to send messages to the client of the build server (an LSP or higher-level BSP implementation).
+    private let connectionToClient: any Connection
+
+    /// Represents the lifetime of the build server implementation..
+    enum ServerState: CustomStringConvertible {
+        case waitingForInitializationRequest
+        case waitingForInitializationNotification
```
ahoppen: I’d suggest sticking closer to the BSP request/notification names here and making this


```suggestion
        case waitingForInitializeRequest
        case waitingForInitializedNotification
```

Sources/SwiftBuild/SWBBuildServer.swift
```diff
@@ -0,0 +1,462 @@
...
+                exitHandler(1)
+            }
+        case is OnBuildInitializedNotification:
+            guard state == .waitingForInitializationNotification else {
+                log(.error, "Build initialized notification received while the build server is \(state.description)")
+                break
+            }
+            state = .running
+        case is OnWatchedFilesDidChangeNotification:
+            guard state == .running else {
```
ahoppen: I don’t think there’s any harm in handling file change notifications while the server is initializing, right? I would log an error if we are in the wrong state but still continue handling it, just makes the BSP server more resilient for badly behaving clients.

Sources/SwiftBuild/SWBBuildServer.swift
```diff
@@ -0,0 +1,462 @@
...
+        updatedBuildRequest.enableIndexBuildArena = true
+        updatedBuildRequest.continueBuildingAfterErrors = true
+        var overridesTable = SWBSettingsTable()
+        overridesTable.set(value: "YES", for: "ONLY_ACTIVE_ARCH")
+        updatedBuildRequest.parameters.overrides.commandLine = overridesTable
+        return updatedBuildRequest
+    }
+
+    public func handle(notification: some NotificationType) async {
+        switch notification {
```
ahoppen, line 126: Another notification to think of early on would be `CancelRequestNotification`.

Sources/SwiftBuild/SWBBuildServer.swift
```diff
@@ -0,0 +1,462 @@
...
+    public func handle<Request: RequestType>(
+        request: Request,
+        id: RequestID,
+        reply: @Sendable @escaping (LSPResult<Request.Response>) -> Void
+    ) async {
+        let request = RequestAndReply(request, reply: reply)
+        if !(request.params is InitializeBuildRequest) {
+            let state = self.state
+            guard state == .running else {
+                await request.reply { throw ResponseError.unknown("Request received while the build server is \(state.description)") }
```
ahoppen, line 172: Similarly here, if we are in the `waitingForInitializationNotification` state, we should be able to handle those requests, I would personally try and make the BSP server handle requests even if they don’t quite come in the right order.

Sources/SwiftBuild/SWBBuildServer.swift
```diff
@@ -0,0 +1,462 @@
...
+                    nil
+                }
+
+                return BuildTarget(
+                    id: try BuildTargetIdentifier(configuredTargetIdentifier: targetInfo.identifier),
+                    displayName: targetInfo.name,
+                    baseDirectory: nil,
+                    tags: tags,
+                    capabilities: BuildTargetCapabilities(),
+                    languageIds: [.c, .cpp, .objective_c, .objective_cpp, .swift],
```
ahoppen, line 297: If there was an easy way to tell which languages are present in the target, it would be nice to only report those but if not, it’s fine. SourceKit-LSP doesn’t end up caring about this. I’d just add a comment explaining why we return all languages here.

Related, worth reporting Metal as well?

Sources/SwiftBuild/SWBBuildServer.swift
```diff
@@ -0,0 +1,462 @@
...
+                of: request.targets.map { try $0.configuredTargetIdentifier },
+                buildDescription: buildDescriptionID,
+                buildRequest: buildRequest
+            )
+            let sourcesItems = try response.compactMap { (swbSourcesItem) -> SourcesItem? in
+                let sources = swbSourcesItem.sourceFiles.map { sourceFile in
+                    SourceItem(
+                        uri: DocumentURI(URL(filePath: sourceFile.path.pathString)),
+                        kind: .file,
+                        // Should `generated` check if the file path is a descendant of OBJROOT/DERIVED_SOURCES_DIR?
```
ahoppen, line 325: FWIW, SourceKit-LSP doesn’t care about it, so 🤷🏽

Sources/SwiftBuild/SWBBuildServer.swift
```diff
@@ -0,0 +1,462 @@
...
+                log(.log, info.message, .report(.init(taskID: buildTaskID)))
+            case .diagnostic(let info):
+                log(.log, info.message, .report(.init(taskID: buildTaskID)))
+            case .backtraceFrame, .reportPathMap, .reportBuildDescription, .preparedForIndex, .buildOutput, .targetStarted, .targetComplete, .targetOutput, .targetUpToDate, .taskUpToDate, .taskOutput, .output:
+                break
+            }
+        }
+    }
+
+    private func log(_ kind: BuildServerMessageType, _ message: String, _ structure: StructuredLogKind? = nil) {
```
ahoppen: Oh, I assumed this logged to the system log. Maybe `logToClient` would be a better term?

Sources/SwiftBuild/SWBBuildServer.swift
```diff
@@ -0,0 +1,462 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+public import SWBBuildServerProtocol
+import SWBProtocol
+import SWBUtil
+import Foundation
+
+/// Wraps a `SWBBuildServiceSession` to expose Build Server Protocol functionality.
+public actor SWBBuildServer: QueueBasedMessageHandler {
```
jakepetroules, line 22: Can we have a nested type that conforms to `QueueBasedMessageHandler` instead? I don't think clients should be calling didReceive, handle, etc., on this.

Probably also allows SWBBuildServerProtocol to become an internal import?

owenv: At a minimum we need to expose a type with a MessageHandler conformance so we can hand it off to a connection object like we do in the tests

Sources/SwiftBuild/SWBBuildServer.swift
owenv: My understanding was that QueueBasedMessageHandlerHelper already handled cancelling Tasks when a CancelRequestNotification is received, do I need more changes here? But admittedly the code could use some more task cancellation checks.

Sources/SwiftBuild/SWBBuildServer.swift
owenv: I just added a comment to that effect for now

Sources/SwiftBuild/SWBBuildServer.swift
owenv: yeah, sgtm

owenv:
@ahoppen did a first pass through your feedback, thanks. I'm waiting to address the rest until I've had more time to think about the logging strategy

Sources/SwiftBuild/SWBBuildServer.swift
ahoppen: Oh, true. Totally forgot about that.

ahoppen:
> did a first pass through your feedback, thanks. I'm waiting to address the rest until I've had more time to think about the logging strategy

Thanks. There was nothing blocking in my comments. Mostly just thoughts on how things could be improved.

owenv:
@swift-ci test

owenv:
@swift-ci test

neonichu (approved):

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test

PR #784, opened 2025-09-08
ModuleVerifier: Add blocklist for builtin module verifier 

cyndyishida:
This patches broken into two commits. 

First change refactors the clang-based blocklist a bit to more easily support extending it. The second change leverages this to add a blocklist for the built-in module verifier. This is to stage in turning this on by default against preexisting verification violations.

### DIFF
--- a/Sources/SWBCore/ToolInfo/ClangToolInfo.swift
+++ b/Sources/SWBCore/ToolInfo/ClangToolInfo.swift
@@ -27,6 +27,17 @@ public struct ClangBlocklists : Sendable {
 
     var caching: CachingBlocklistInfo? = nil
 
+    public struct BuiltinModuleVerifierInfo : ProjectFailuresBlockList, Codable, Sendable {
+        /// A blocklist of project names that do not support the `MODULE_VERIFIER_KIND=builtin` build setting.
+        let KnownFailures: [String]
+        enum CodingKeys: String, CodingKey {
+            case KnownFailures
+        }
+    }
+
+    var builtinModuleVerify: BuiltinModuleVerifierInfo? = nil
+
+    /// Helper method for determining if a given functionality is blocklisted for the active scope.
     func isBlocked<BlockListT: ProjectFailuresBlockList>(_ scope: MacroEvaluationScope, info: BlockListT?) -> Bool {
         guard let blocklistInfo = info else { return false }
         return blocklistInfo.isProjectListed(scope)
@@ -90,6 +101,10 @@ public struct DiscoveredClangToolSpecInfo: DiscoveredCommandLineToolSpecInfo {
     public func isCachingBlocked(_ scope: MacroEvaluationScope) -> Bool {
         return blocklists.isBlocked(scope, info: blocklists.caching)
      }
+
+    public func isBuiltinModuleVerifyBlocked(_ scope: MacroEvaluationScope) -> Bool {
+        return blocklists.isBlocked(scope, info: blocklists.builtinModuleVerify)
+    }
 }
 
 private let clangVersionRe = RegEx(patternLiteral: #""(?<llvm>[0-9]+(?:\.[0-9]+){0,}) \(clang-(?<clang>[0-9]+(?:\.[0-9]+){0,})\)(?: ((\[.+\])|(\(.+\)))+)?""#)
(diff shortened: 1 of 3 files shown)


cyndyishida:
@swift-ci please test

Sources/SWBCore/ToolInfo/ClangToolInfo.swift
```diff
@@ -10,9 +10,40 @@
 //
 //===----------------------------------------------------------------------===//
 
+public import SWBMacro
 public import SWBUtil
 import Foundation
 
+public struct ClangBlocklists : Sendable {
```
ian-twilightcoder, line 17: I don't know which ones to change but I really wish we were consistent about Blocklist vs BlockList capital L

Sources/SWBCore/ToolInfo/ClangToolInfo.swift
```diff
@@ -10,9 +10,40 @@
...
+
+    public struct CachingBlocklistInfo : ProjectFailuresBlockList, Codable, Sendable {
+        /// A blocklist of project names that do not support the `CLANG_ENABLE_COMPILE_CACHE` build setting.
+        let KnownFailures: [String]
+
+        enum CodingKeys: String, CodingKey {
```
ian-twilightcoder, line 23: Do we need coding keys for either of these?

Sources/SWBCore/ToolInfo/ClangToolInfo.swift
```diff
@@ -10,9 +10,40 @@
...
+        let KnownFailures: [String]
+        enum CodingKeys: String, CodingKey {
+            case KnownFailures
+        }
+    }
+
+    var builtinModuleVerify: BuiltinModuleVerifierInfo? = nil
+
+    /// Helper method for determining if a given functionality is blocklisted for the active scope.
+    func isBlocked<BlockListT: ProjectFailuresBlockList>(_ scope: MacroEvaluationScope, info: BlockListT?) -> Bool {
```
ian-twilightcoder, line 41: I wonder if we really need these helper methods since the Swift analogues don't use them

jakepetroules (approved):

Sources/SWBCore/ToolInfo/ClangToolInfo.swift
cyndyishida: The swift ones also handle blocking checks in different ways for each block list category.

Sources/SWBCore/ToolInfo/ClangToolInfo.swift
cyndyishida: Not sure & given the lack of certainty here, I'd prefer to leave previously working code alone.

ian-twilightcoder: I'm pretty sure we don't, at least a trivial test seems to work fine, and `SwiftBlocklists.CachingBlockList` doesn't have it so I think we don't need it anywhere?

Sources/SWBCore/ToolInfo/ClangToolInfo.swift
cyndyishida: I _think_ all the ones I introduced are spelled with an undercase `l`. While I agree that being consistent is good. I don't know that I'll spend the time to change it across the whole codebase.

Sources/SWBCore/ToolInfo/ClangToolInfo.swift
cyndyishida: It seems like all enum definitions of `CodingKeys` use `CodingKey`, this feels like another case of general cleanup if it's not necessary.

Sources/SWBCore/ToolInfo/ClangToolInfo.swift
cyndyishida: Strictly speaking though, I don't think we need the helper function I just think it's a nice abstraction to keep the details of the blocklist inside `ClangToolInfo`. And if one added more complexity in blocklist lookup's it could be done in this helper.

cyndyishida: Spoke with Ian offline, he's ok with merging this patch as is.

PR #785, opened 2025-09-09
Add additional flags for symbol graph generation options

cmcgee1024:
Add flags for pretty print and skip synthesized members to Swift Build so that
clients can request these options when generating symbol graphs for DocC
documentation.

SwiftPM requires control over these options when extracting symbol graphs
for targets, and allows clients to use these options through the command
plugins.


cmcgee1024:
@swift-ci please test

PR #787, opened 2025-09-09
Set PATH in the env when invoking dsymutil so it can find lipo in another toolchain

owenv:
For example, on macOS dsymutil in a Swift.org toolchain may need to find lipo in the Xcode toolchain

### DIFF
--- a/Sources/SWBCore/SpecImplementations/Tools/DsymutilTool.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/DsymutilTool.swift
@@ -11,7 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 public import SWBUtil
-import SWBMacro
+public import SWBMacro
 
 public final class DsymutilToolSpec : GenericCommandLineToolSpec, SpecIdentifierType, @unchecked Sendable {
     public static let identifier = "com.apple.tools.dsymutil"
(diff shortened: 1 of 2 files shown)


owenv:
@swift-ci test

Sources/SWBCore/SpecImplementations/Tools/DsymutilTool.swift
```diff
@@ -21,6 +21,13 @@ public final class DsymutilToolSpec : GenericCommandLineToolSpec, SpecIdentifier
         fatalError("unexpected direct invocation")
     }
 
+    public override func environmentFromSpec(_ cbc: CommandBuildContext, _ delegate: any DiagnosticProducingDelegate, lookup: ((MacroDeclaration) -> MacroExpression?)? = nil) -> [(String, String)] {
+        var env: [(String, String)] = super.environmentFromSpec(cbc, delegate, lookup: lookup)
+        // dsymutil may need to lookup lipo, which is not necessarily in the same toolchain.
+        env.append(("PATH", cbc.producer.executableSearchPaths.environmentRepresentation))
```
jakepetroules, line 27: I know this particular PR is about `lipo` but note that `PATH` is usually spelled `Path` on Windows -- this might be a good opportunity to switch out `String` for `EnvironmentKey` in a few more places, though I can understand if that's too far outside the scope of what you're aiming to fix right now.

PR #789, opened 2025-09-09
Stop passing -framework to libtool

owenv:
The comment on this code was incorrect, as libtool has always ignored the -framework flag, even where a framework contains a static archive


owenv:
@swift-ci test

owenv:
@swift-ci test

PR #790, opened 2025-09-09
Avoid building empty objects in tests

owenv:
Prevent emitting warnings when using an archiver which warns about empty object files


owenv:
@swift-ci test

mhrawdon (approved):

PR #773, opened 2025-09-04
Fix the build on iOS and add CI coverage

jakepetroules:

PR #791, opened 2025-09-10
[CI] GitHub Actions - Support macOS testing

shahmishal:


owenv:
Failures:
incrementalMetalLinkWithCodeSign, singleFileCompileMetal - failing because the Metal toolchain is not installed
cachedBuild - is seeing some unexpected cache misses

@shahmishal for the time being, I think we should probably only bring up the Xcode 26b6 job since we've only been maintaining compatibility with 16.2 for use in Swift CI

jakepetroules:
> Failures: incrementalMetalLinkWithCodeSign, singleFileCompileMetal - failing because the Metal toolchain is not installed cachedBuild - is seeing some unexpected cache misses
> 
> @shahmishal for the time being, I think we should probably only bring up the Xcode 26b6 job since we've only been maintaining compatibility with 16.2 for use in Swift CI

Note that there's **three**: 16.2, 16.3, and 26. And 16.3 is completely passing.

Are you suggesting we bring up _only_ 26, or only 26 + 16.3?

jakepetroules:
We can also just merge #773 and then look at adding 26 later.

shahmishal:
@swift-ci test

PR #792, opened 2025-09-10
Add minimum access level as a setting for the symbol graph extraction

cmcgee1024:
This setting wasn't present before, so this is added as a build setting enumeration with
a special "none" default value that performs the fallback that is based on the kind of the
module.

Remove the swift compiler feature checks for the previously added settings since the
feature is unrelated to them.


cmcgee1024:
@swift-ci please test

Sources/SWBCore/Settings/BuiltinMacros.swift
```diff
@@ -2757,6 +2759,18 @@ public enum LinkerDriverChoice: String, Equatable, Hashable, EnumerationMacroTyp
     case auto
 }
 
+public enum DoccMinimumAccessLevel: String, Equatable, Hashable, EnumerationMacroType {
```
owenv, line 2762: You may need to add a case to the switch in EnumBuildOptionType.declareMacro for this enum type too. It might be nice to have a test in DocumentationTaskConstructionTests that the flag is being passed through correctly

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci please test

owenv (approved):

PR #795, opened 2025-09-11
Update FilePathResolver so the product references can use a build setting

daveinglis:


daveinglis:
@swift-ci please test

owenv (approved):
This change should be safe and reasonable IMO

daveinglis:
@swift-ci please test macos

Sources/SWBCore/ProjectModel/FilePathResolver.swift
```diff
@@ -84,7 +84,7 @@ public final class FilePathResolver: Sendable
             }
 
             // FIXME: We are relying on the product reference name being constant here. This is currently true, given how our path resolver works, but it is possible to construct an Xcode project for which this doesn't work (Xcode doesn't, however, handle that situation very well). We should resolve this: <rdar://problem/29410050> Swift Build doesn't support product references with non-constant basenames
```
jakepetroules: I think you are resolving this issue and therefore need to remove the FIXME?

I'm also a little worried there may be extra non-obvious complexity here, but let's give it a try to start.

jakepetroules (changes requested):

Sources/SWBCore/ProjectModel/FilePathResolver.swift
daveinglis: fixed

daveinglis:
@swift-ci please test

jakepetroules (approved):

jakepetroules:
@daveinglis Just took another look at this and there appear to be a bunch of places in the target and linkage dependency resolvers where this is _not_ handled consistently with the changes you made here.

PR #796, opened 2025-09-13
Fix linker filelist parsing when discovering tests

owenv:
Take the format into account when parsing response files, and consistently track it when we read and write them in various places throughout the build system

The immediate motivation is to fix test discovery when the path to an object file contains whitespace


owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test

jakepetroules (approved):

PR #797, opened 2025-09-13
Add API to allow clients to manage build description lifecycle

owenv:
Add basic optional refcounting for build descriptions so that clients making repeated requests can choose to manually manage their lifecycle.

### DIFF
--- a/Sources/SWBBuildService/BuildOperationMessages.swift
+++ b/Sources/SWBBuildService/BuildOperationMessages.swift
@@ -165,7 +165,7 @@ final class ActiveBuild: ActiveBuildOperation {
     /// Whether this operation is intended only for creating and reporting the build description.
     let onlyCreatesBuildDescription: Bool
 
-    /// Whether this operation should retain the build description is uses.
+    /// Whether this operation should retain the build description it uses.
     let retainBuildDescription: Bool
 
     /// The current state of the build.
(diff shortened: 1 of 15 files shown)


owenv:
@swift-ci test

owenv:
@swift-ci test

jonathanpenn:
lgtm. Just to double check, this is ultimately only going to be a signal to the in-memory cache, right?

jonathanpenn (approved):

jonathanpenn (approved):

owenv:
@jonathanpenn this is a guarantee a build description will never be destroyed rather than a signal it shouldn't be. But it's opt-in, because it's designed primarily for usage in long-lived build servers

owenv:
@swift-ci test

owenv:
@swift-ci test

Sources/SWBBuildService/BuildOperationMessages.swift
```diff
@@ -165,6 +165,9 @@ final class ActiveBuild: ActiveBuildOperation {
     /// Whether this operation is intended only for creating and reporting the build description.
     let onlyCreatesBuildDescription: Bool
 
+    /// Whether this operation should retain the build description is uses.
```
jakepetroules:
```suggestion
    /// Whether this operation should retain the build description it uses.
```

owenv:
@swift-ci test

PR #798, opened 2025-09-15
Move a build setting from swiftPM to swift-build

daveinglis:
There was some code in swiftPM that would set
CLANG_COVERAGE_MAPPING_LINKER_ARGS = NO if the mach_o_type was "mh_object" due to some duplicate symbol problems, this change moves it down to the objfile product type spec where it belongs.


owenv (approved):

owenv:
@swift-ci test

daveinglis:
@swift-ci test

PR #799, opened 2025-09-15
Metrics: vend counters as strings to avoid making SWBProtocol part of the public interface for SwiftBuild

mirza-garibovic:


mirza-garibovic:
@swift-ci test

owenv (approved):

Sources/SwiftBuild/ConsoleCommands/SWBServiceConsoleBuildCommandProtocol.swift
```diff
@@ -169,10 +168,10 @@ public enum SwiftBuildMessage {
     }
 
     public struct BuildOperationMetrics {
-        let counters: [SWBProtocol.BuildOperationMetrics.Counter: Int]
+        let counters: [String: Int]
```
jakepetroules: Don't these fields need to be public, too? Otherwise you can't access them...

mirza-garibovic: Oops, that's a good point. Currently they're accessed via Codable conformance and JSON encoding. I'll fix it.

mirza-garibovic:
Going to wait for @akyrtzi review for this one.

jakepetroules (approved):

Sources/SwiftBuild/SWBuildMessage+Protocol.swift
```diff
@@ -48,7 +48,9 @@ extension SwiftBuildMessage {
     init(_ message: BuildOperationEnded) {
         let metrics: BuildOperationMetrics?
         if let messageMetrics = message.metrics, !messageMetrics.counters.isEmpty {
```
akyrtzi: This check will omit the metrics if `counters` is empty even if `taskCounters` is not. Maybe add a `BuildOperationMetrics.isEmpty` function that will check for both?

mirza-garibovic: Awesome, thanks for catching that! I [updated it](https://github.com/swiftlang/swift-build/compare/d544a5abf80de99f4214c26aa3b6f7c904d79d93..c8b3c74349eb2bff65b693bd9c657488c3f31d9a) but not as suggested, I didn't want nil to have ambiguous meaning (was it empty or nil for another reason), but I do propagate nil input (which is used for build cancellation and some types of build failures). Does that work?

akyrtzi: Sounds good!

akyrtzi (approved):

mirza-garibovic:
@swift-ci test

mirza-garibovic:
@swift-ci test

Sources/SwiftBuild/SWBuildMessage+Protocol.swift
mirza-garibovic: I went back and took your suggestion, because of `BuildCommandTests` tests which expect a specific json dump and no desire to revamp them at this point

PR #800, opened 2025-09-16
Expose SWIFT_LIBRARIES_ONLY setting for SwiftPM

owenv:
This will allow the PIF to explicitly set for every target whether we should pass -parse-as-library to swiftc


owenv:
@swift-ci test

PR #801, opened 2025-09-17
Add API to query information about artifacts produced by configured targets

owenv:
This is intended to support SwiftPM command plugin API for reporting artifacts which were built, so it's only applied to a subset of product types for now


owenv:
@swift-ci test

Sources/SwiftBuild/SWBConfiguredTargetInfo.swift
```diff
@@ -11,6 +11,32 @@
 //===----------------------------------------------------------------------===//
 
 import SWBProtocol
+import SWBUtil
+
+public struct SWBArtifactInfo {
```
jakepetroules: Where is the Sendable conformance? With the lifting of unsafeFlags limitations in 6.2 I wonder if we can turn on the setting to make explicit Sendable annotations a requirement now, for public types in the SwiftBuild module.

owenv:
@swift-ci test

PR #806, opened 2025-09-18
Skip tests requiring the metal toolchain on GH actions runners

owenv:
These runners don't have the optional toolchain download


owenv:
@swift-ci test

PR #807, opened 2025-09-18
Disable optimization to work around a compiler bug

Steelskin:
This disables optimization on one function to work around a Swift compiler bug in the LLVM 21 rebranch.

Bug: swiftlang/llvm-project#11377

### DIFF
--- a/Sources/SWBTaskExecution/BuildDescriptionManager.swift
+++ b/Sources/SWBTaskExecution/BuildDescriptionManager.swift
@@ -175,6 +175,9 @@ package final class BuildDescriptionManager: Sendable {
     /// NOTE: This is primarily accessible for performance testing purposes, actual clients should prefer to access via the cached methods.
     ///
     /// - Returns: A build description, or nil if cancelled.
+    // TODO: Optimizations are disabled to work around a compiler bug. Remove this attribute when the bug is fixed.
+    // See https://github.com/swiftlang/llvm-project/issues/11377 for details.
+    @_optimize(none)
     package static func constructBuildDescription(_ plan: BuildPlan, planRequest: BuildPlanRequest, signature: BuildDescriptionSignature, inDirectory path: Path, fs: any FSProxy, bypassActualTasks: Bool = false, planningDiagnostics: [ConfiguredTarget?: [Diagnostic]], delegate: any BuildDescriptionConstructionDelegate) async throws -> BuildDescription? {
         BuildDescriptionManager.descriptionsComputed.increment()
 


Steelskin:
@swift-ci please smoke test

compnerd:
@owenv we are investigating the LLVM side of the issue. PEI is at fault here, and changing that is going to take some time. This is to unblock the work for the rebranch.

CC: @bnbarham @AnthonyLatsis

bnbarham:
I've started a build in https://github.com/swiftlang/swift/pull/71188 to see how far we get. If this is the only issue then we still have time to keep investigating. Worst case we might be able to split things up and only apply `@_optimize(none)` in a much narrower case? Though having said that, this function actually isn't doing all that much work is it 🤔 (it's all mostly in the called function)?

owenv:
Could we restrict the attr to only target Windows? The bug  seems to imply this only affects the windows prologue/epilogue emission?

Steelskin:
> Could we restrict the attr to only target Windows? The bug seems to imply this only affects the windows prologue/epilogue emission?

I'll see about limiting it to Windows/arm64, since it is the only affected platform AFAIK.

AnthonyLatsis:
```swift
#if os(Windows)
@_optimize(none)
#endif
```
Should do the trick.

owenv:
ok, I'm ok with accepting this as a temporary change if we take @AnthonyLatsis 's suggestion

Steelskin:
> ```swift
> #if os(Windows)
> @_optimize(none)
> #endif
> ```
> 
> Should do the trick.

Thanks, done!

owenv (approved):

Steelskin:
@swift-ci please smoke test

owenv:
@swift-ci test

(we don't have smoke test jobs in this repo bc they don't build the whole stack)

jakepetroules (approved):

jakepetroules (approved):

AnthonyLatsis:
There appears to be another problematic function in SwiftPM: https://ci-external.swift.org/job/swift-PR-build-toolchain-windows-arm64/42/console

Steelskin:
@swift-ci please test linux platform

Steelskin:
@swift-ci please test macos platform

Steelskin:
It looks like there is still one unrelated failure here but I can't re-run the job.

Steelskin:
FYI, I got the full Windows arm64 toolchain building locally in conjunction with swiftlang/swift-package-manager#9178

owenv:
This is ok to merge, the failing job isn't fully brought up yet

PR #810, opened 2025-09-20
Take the host platform into account when writing clang response files

owenv:
This ensures the contents get properly escaped


owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test

Sources/SWBUtil/ProcessInfo.swift
```diff
@@ -167,6 +167,15 @@ public enum OperatingSystem: Hashable, Sendable {
             return .elf
         }
     }
+
+    public var defaultResponseFileFormat: ResponseFileFormat {
+        switch self {
```
jakepetroules, line 172: nit: use this?

```swift
self == .windows ? .windowsShellQuotedNewlineSeparated : .unixShellQuotedSpaceSeparated
```

Seems unlikely there will ever be difference except for Windows

daveinglis:
I tested this locally but the  `.windowsShellQuotedNewlineSeparated:` which uses `WindowsProcessArgumentsCodec()` does not seem to work for clang as it does not escape windows path separators, however using `LLVMStyleCommandCodec()` does seem to work, should this be the default for Windows?

owenv:
@daveinglis possibly, we may need to dig into the clang implementation a bit to see what it's expecting to receive

daveinglis (approved):

owenv:
Closing this one for now

PR #814, opened 2025-09-23
Fix the combination of DEAD_CODE_STRIPPING=YES and LINKER_DRIVER=swiftc

owenv:
Unconditionally use -Xlinker -dead_strip since it'll be handled correctly by either linker driver


owenv:
@swift-ci test

PR #815, opened 2025-09-23
Add build settings to control the emission of the Index Store

ahoppen:
Add build settings that allow fine-tuning how the index store is being emitted during index-while-building.


ahoppen:
@swift-ci Please test

PR #816, opened 2025-09-23
Add an option to make the bridging header internal

DougGregor:
Add `SWIFT_BRIDGING_HEADER_IS_INTERNAL` option, which defaults to NO. When it is YES, use the new `-internal-import-bridging-header` instead of `-import-objc-header`. The new version treats the contents of the bridging header as if they came in through an internal import.

Tracked by rdar://161172254.


DougGregor:
@swift-ci please test

PR #817, opened 2025-09-24
Correct cross-platform formatting of clang response files

owenv:
This is a more correct alternative to https://github.com/swiftlang/swift-build/pull/810

clang doesn't use Windows style response files unless you invoke it as clang-cl. Instead, use the existing LLVMStyleCommandCodec which should give correct behavior on any platform


owenv:
@swift-ci test

PR #818, opened 2025-09-24
Add a request to find a most appropriate configured target for indexi…

xedin:
…ng purposes

It's prompted by the fact that LSP has to operate on targets instead of configured targets at the moment because there is (currently) no way (or enough information) on LSP side to select a canonical target for a given document.

This change creates an API endpoint for `selectConfiguredTargetForIndex`. It makes it possible to select a single configured target given a target GUID and a run destination (+ other parameters).


xedin:
@swift-ci please test

PR #819, opened 2025-09-24
Disable SWIFT_MODULE_ONLY_ARCHS if we detect a SWIFT_PLATFORM_TARGET_PREFIX

owenv:
In these situations, we can't accurately predict what platform the user is actually targeting, so we shouldn't try to generate modules for extra archs

rdar://160601005


owenv:
@swift-ci test

jakepetroules (approved):

PR #821, opened 2025-09-26
[Explicit Module Builds] Add a setting to allow creating a reproducer for errors.

vsapsai:
The idea is the same as Clang flag `-gen-reproducer=error` but for explicit module builds. The name of the setting is a subject to change depending on how it ends up being used.

Not all module-related issues are crashes, so it is useful to allow creating and sharing reproducers for errors.

rdar://158780565


Sources/SWBTaskExecution/TaskActions/ClangCompileTaskAction.swift
```diff
@@ -319,7 +319,10 @@ public final class ClangCompileTaskAction: TaskAction, BuildValueValidatingTaskA
                         outputDelegate.emitOutput(ByteString(encodingAsUTF8: commandString) + "\n")
                     }
 
-                    if case .some(.failed) = lastResult, case .some(.exit(.uncaughtSignal, _)) = outputDelegate.result {
+                    let shouldGenerateReproducer = (lastResult == .failed) &&
+                        (explicitModulesPayload.shouldGenerateReproducerForErrors ||
+                         (outputDelegate.result?.isCrashed ?? false))
```
vsapsai, line 324: I've spent some time trying to improve this code, so let me share a few options I've discarded. If you prefer one of the mentioned options, let me know

```swift
let shouldGenerateReproducer: Bool
if lastResult == .failed {
    if explicitModulesPayload.shouldGenerateReproducerForErrors {
        shouldGenerateReproducer = true
    } else if case .exit(exitStatus: .uncaughtSignal, metrics: _) = outputDelegate.result {
        // In case of crash.
        shouldGenerateReproducer = true
    } else {
        shouldGenerateReproducer = false
    }
} else {
    shouldGenerateReproducer = false
}
```

Slight variation with if-expression
```swift
let shouldGenerateReproducer = if lastResult == .failed {
    if explicitModulesPayload.shouldGenerateReproducerForErrors {
        true
    } else if case .exit(exitStatus: .uncaughtSignal, metrics: _) = outputDelegate.result {
        // In case of crash.
        true
    } else {
        false
    }
} else {
    false
}
```

Try to add `case`
```swift
let shouldGenerateReproducer = if lastResult == .failed {
    switch (explicitModulesPayload.shouldGenerateReproducerForErrors, outputDelegate.result) {
    case (true, _): true
    case (_, .some(.exit(exitStatus: .uncaughtSignal, metrics: _))): true
    default: false
    }
} else {
    false
}
```

jakepetroules: No strong opinion from me, seems fine as it is.

jakepetroules:
@swift-ci test

vsapsai:
@swift-ci test

vsapsai:
As far as I can tell all the failed tests are in ClangCompilationCachingTests.swift that is unrelated to my change.

PR #824, opened 2025-09-29
Miscellaneous fixes for LINKER_DRIVER=auto

owenv:
- Fix -nostdlib handling in this mode
- Always use clang to link relocatable objects. swiftc doesn't support these very well right now. We should probably fix this, but linking one doesn't require much knowledge of the source language(s) so this should be ok for now


owenv:
@swift-ci test

Issue #610, opened 2025-06-24
Introduce a new product type for SwiftPM targets


owenv:
This will allow us to abstract away various platform differences, especially on Windows

PR #826, opened 2025-09-30
Create a new product type so the PIF remains platform agnostic

daveinglis:
Closes https://github.com/swiftlang/swift-build/issues/610


Sources/SWBApplePlatform/Specs/DarwinProductTypes.xcspec
```diff
@@ -474,4 +474,15 @@
             com.apple.package-type.mach-o-executable
         );
     },
+    {
+        _Domain = darwin;
+        Type = ProductType;
+        Identifier = org.swift.product-type.common.object.bundle;
```
daveinglis: I'm open to a better identifier if anyone would like to suggest one....

daveinglis:
@swift-ci test

Sources/SWBApplePlatform/Specs/DarwinProductTypes.xcspec
owenv: bundle would be good to avoid since it has a specific existing meeting. Maybe just org.swift.product-type.common.object.target?

owenv (approved):
LGTM once we figure out naming

Sources/SWBApplePlatform/Specs/DarwinProductTypes.xcspec
jakepetroules: I wonder if anything is relying on the prefix being com.apple.product-type, but I guess this is a good test to find out if that's the case.

Sources/SWBApplePlatform/Specs/DarwinProductTypes.xcspec
```diff
@@ -474,4 +474,15 @@
             com.apple.package-type.mach-o-executable
         );
     },
```
jakepetroules, line 476: If the objects have no differing properties across Windows/Darwin/Unix, should we just put one single spec in the global domain for simplicity rather than 3 separate ones?

jakepetroules: Over the years we deduplicated the same kind of thing across macOS and iOS to avoid similar redundancies.

owenv: The windows one inherits from a different product type, so that one is somewhat unique

jakepetroules: Ah, I missed that. That's a little unfortunate; I'd hoped to avoid platform-specific product type hierarchies as much as possible but we could always change this sort of thing later.

jakepetroules (approved):

Sources/SWBApplePlatform/Specs/DarwinProductTypes.xcspec
daveinglis: maybe I just drop bundle?  `target` seems a bit odd since its a product type identifier.
so some what about one of these: 
`org.swift.product-type.common.object` or maybe ~`org.swift.product-type.common.object.type`~?

daveinglis: forget that last one... type is already there...

daveinglis:
@swift-ci test

Sources/SWBApplePlatform/Specs/DarwinProductTypes.xcspec
daveinglis: ok... going with `org.swift.product-type.common.object`

Issue #827, opened 2025-09-30
SwiftBuild ProjectModel::BuildSettings.platformSpecificSettings needs to support single value settings


daveinglis:
SwiftBuild ProjectModel::BuildSettings.platformSpecificSettings only supports multi-value settings, in order to be able to set CLANG_ENABLE_MODULES using a platform specifier, single value settings need to be supported.

PR #828, opened 2025-09-30
Improve project model build settings API for platform-conditional settings

owenv:
The old API only allowed multi-value platform conditional settings and tended to inject lots of $(inherited). Add a replacement which more closely mirrors the non-conditional API

closes https://github.com/swiftlang/swift-build/issues/827


owenv:
@swift-ci test

PR #830, opened 2025-09-30
[Tests] NFC: Add `iOS` requirement to `selectConfiguredTargets` test

xedin:
The test checks whether it's possible to select an appropriate target given a run destination and uses both macOS and iOS SDKs.

If there is no `iOS` SDK installed it won't be possible to configure a test target for that platform and the test is going to fail.

Resolves: rdar://161676960


xedin:
@swift-ci please test

owenv (approved):

PR #831, opened 2025-10-01
Set TMPDIR when invoking linker drivers on Windows

owenv:
swiftc.exe falls over on Windows when used as a linker if it doesn't have TMPDIR set to a reasonable value. Point it to OBJROOT to unblock adoption of LINKER_DRIVER=auto in SwiftPM


owenv:
@swift-ci test

bkhouri (commented):
Can we add some automated tests to ensure the behaviour does not regress?

owenv:
It's tested at the SwiftPM layer, there's not much of interest to assert here

PR #837, opened 2025-10-02
Pass `-Xcc -std=c++2b` to Swift when `CLANG_CXX_LANGUAGE_STANDARD` is set to `c++2b`

egorzhdan:
If pbxproj contains this line:
```
CLANG_CXX_LANGUAGE_STANDARD = "c++2b";
```
then we should pass the corresponding flag to Swift compiler invocations that enable Swift/C++ interop.

Prior to this change, this worked fine for `c++23`, but not `c++2b`.

rdar://149474443


egorzhdan:
@swift-ci please test

owenv (approved):

egorzhdan:
The sole test failure on the CI doesn't seem to be caused by this PR.

PR #838, opened 2025-10-02
Introduce INSTALLLOC_DIRECTORY_CONTENTS to allow installloc to support installing ad-hoc bundle sub-contents

matthewseaman:
This change allows `installloc` to treat directories and package files in Resources and Copy Files build phases the same as if they were actual sub-bundles.

In practice, these usually can be thought of as sub-bundles and could contain their own lproj directories that should be installed into a localization root.

Putting behind a user-defined build setting for now until we can evaluate how this impacts a variety of project configurations.


matthewseaman:
@swift-ci test

owenv (approved):

matthewseaman:
@swift-ci please test macOS

matthewseaman:
@swift-ci test Linux

PR #840, opened 2025-10-02
CustomTasks should include environment in display strings

owenv:
Similar to shell scripts, this is often useful to display to users


owenv:
@swift-ci test

jakepetroules (approved):

PR #841, opened 2025-10-02
Don't fail when QNX 7 SDP is present on the system and Rosetta is not installed

### LINKED ISSUE #836: error: Could not compute dependency graph: Error Domain=NSPOSIXErrorDomain Code=86 "Bad CPU type in executable"
When `swift-build` computes the dependency graph for a build it seems to resolve the toolchains available on the system, which may include tools that are `x86_64` only. If Rosetta2 is not installed, this results in a hard-failure for every build in Xcode/xcodebuild on Apple Silicon system:

Image

Installing Rosetta2, and then watching process creation via [Process Monitor](https://objective-see.org/products/utilities.html#ProcessMonitor) during a build shows:

```json
{
  "event": "ES_EVENT_TYPE_NOTIFY_EXEC",
  "timestamp": "2025-10-02 09:39:14 +0000",
  "process": {
    "pid": 35254,
    "name": "qcc",
    "path": "/opt/qnx700/host/darwin/x86_64/usr/bin/qcc",
    "uid": 501,
    "architecture": "unknown",
    "arguments": [
      "/opt/qnx700/host/darwin/x86_64/usr/bin/qcc",
      "-dM",
      "E",
      "-x",
      "c",
      "-c",
      "/dev/null",
      "-o",
      "/dev/null"
    ],
    "ppid": 35253,
    "rpid": 34691,
    "ancestors": [
      34691,
      1
    ],
    "signing info (reported)": {
      "csFlags": 0,
      "platformBinary": 0,
      "signingID": "",
      "teamID": "",
      "cdHash": "0000000000000000000000000000000000000000"
    },
    "signing info (computed)": {
      "signatureStatus": -67062
    }
  }
}
```

With `SWBBuildService` as its parent process.

I don't know if there's any way to opt out of (specific) toolchain lookups? 

In any case, the logic [here for QNX](https://github.com/swiftlang/swift-build/blob/823f4f8c84bbf8e6a6752d4550fafaebfec43bb3/Sources/SWBQNXPlatform/QNXSDP.swift#L41), or possibly at a higher level, should probably handle the situation gracefully by treating it as the toolchain not found, possibly with an explicit warning, rather than the opaque "Bad CPU type in executable" with no info about _which_ executable failed to run 😄

jakepetroules:
Auxiliary platform lookup is intended to be failable; one of the three calls to cachedQNXSDPInstallations was missing a `try?`

Closes #836

### DIFF
--- a/Sources/SWBQNXPlatform/Plugin.swift
+++ b/Sources/SWBQNXPlatform/Plugin.swift
@@ -45,7 +45,7 @@ struct QNXEnvironmentExtension: EnvironmentExtension {
     let plugin: QNXPlugin
 
     func additionalEnvironmentVariables(context: any EnvironmentExtensionAdditionalEnvironmentVariablesContext) async throws -> [String : String] {
-        if let latest = try await plugin.cachedQNXSDPInstallations(host: context.hostOperatingSystem).first {
+        if let latest = try? await plugin.cachedQNXSDPInstallations(host: context.hostOperatingSystem).first {
             return .init(latest.environment)
         }
         return [:]


jakepetroules:
@swift-ci test

owenv (approved):

PR #844, opened 2025-10-03
Fix coverage mapping linker args when using swiftc linker driver

owenv:
Plumb the linker driver selection through to AdditionalLinkerArgs selection in specs, and adopt it for the coverage args


owenv:
@swift-ci test

jakepetroules (approved):

PR #845, opened 2025-10-04
Avoid duplicate OS rpaths when linking with swiftc and targeting older macOS

owenv:
The build system already inserts the /usr/lib/swift rpath if needed, so suppress the one the linker driver inserts to avoid warning about duplicates


owenv:
@swift-ci test

jakepetroules (approved):

PR #847, opened 2025-10-06
Retain old-syle response file escaping on macOS for compatibility

owenv:
Some projects on macOS currently manually shell escape the values of build settings which are translated to compiler args, which breaks the escaping change here, as they end up double escaped. Restore the old behavior on macOS for compatibility.

rdar://162035972


owenv:
@swift-ci test

neonichu (approved):

jakepetroules:
Wouldn't these projects have failed if response files were turned off entirely? Since arguments are used exactly as the macro system sees their resolved value.

PR #848, opened 2025-10-06
[Tests] NFC: Add iOS requirement to `selectConfiguredTargets` instead…

xedin:
… of `artifacts`

The test checks whether it's possible to select an appropriate target given a run destination and uses both macOS and iOS SDKs.

If there is no iOS SDK installed it won't be possible to configure a test target for that platform and the test is going to fail.

Resolves: rdar://162067445


xedin:
@swift-ci please test

jakepetroules (approved):

owenv (approved):

PR #850, opened 2025-10-07
Use canonical URLs for Info.plist documentation in specs

owenv:
Update docs to use canonical URLs instead of redirects


owenv:
@swift-ci test

jakepetroules (approved):

PR #852, opened 2025-10-07
Disable modular builds when compiling for >= c++20 modes

cyndyishida:
The clang compiler will implicitly enable the modules feature when in language mode c++20 or greater. This is for c++20 modules, which strictly differ from clang modules, but are conflated at times like when the compiler evaluates `__has_feature(modules)`.
Since the sdks cannot cleanly build in this mode, disable it by default but allow explicit overrides.

rdar://156935289

### DIFF
--- a/Sources/SWBUniversalPlatform/Specs/Clang.xcspec
+++ b/Sources/SWBUniversalPlatform/Specs/Clang.xcspec
@@ -365,7 +365,18 @@
                     "sourcecode.cpp.cpp",
                     "sourcecode.cpp.objcpp"
                 );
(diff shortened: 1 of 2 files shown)


cyndyishida:
@swift-ci please test

cyndyishida:
Are the tests that are required to pass the ones triggered by `swift-ci`? The currently failing tests don't seem related to my change, but I can look deeper if they're unexpected.

owenv:
This is the second time I've seen these this morning, I think it;s some kind of infra issue

PR #853, opened 2025-10-08
Fix building .S files on windows & resource bundles

daveinglis:
- building .S files on windows requires clang to have TMP/TEMP env passed (also move the linker change that added TMPDIR to the tool env to the spec file for consistency).
- fix the resource bundle error that when touching a file was passing a badly quoted path.


daveinglis:
@swift-ci test

owenv:
```
[2025-10-08T15:38:31.837Z] × Test touchTaskConstruction() recorded an issue at CommandLineSpecTests.swift:1723:34: Expectation failed: (actual → ["C:\\Windows\\system32\\cmd.exe", "/c", "copy", "/b", "C:\\tmp\\input", "+,,"]) == (expected → ["C:\\Windows\\system32\\cmd.exe", "/c", "copy /b \"C:\\tmp\\input\" +,,"])
```

Looks like the test just needs a small fixup

daveinglis:
@swift-ci test

Sources/SWBCore/SpecImplementations/Tools/TouchTool.swift
```diff
@@ -42,8 +42,7 @@ final class TouchToolSpec : CommandLineToolSpec, SpecIdentifierType, @unchecked
                 delegate.error("Can't determine path to cmd.exe because the ComSpec environment variable is not set")
                 return
             }
-            // FIXME: Need to properly quote the path here, or generally handle this better
-            commandLine = [commandShellPath, "/c", "copy /b \"\(input.absolutePath.str)\" +,,"]
+            commandLine = [commandShellPath, "/c", "copy", "/b", input.absolutePath.str, "+,,"]
```
jakepetroules, line 45: cmd.exe's help says:

> If /C or /K is specified, then the remainder of the command line after
> the switch is processed as a command line, where the following logic is
> used to process quote (") characters:
> 
>     1.  If all of the following conditions are met, then quote characters
>         on the command line are preserved:
> 
>         - no /S switch
>         - exactly two quote characters
>         - no special characters between the two quote characters,
>           where special is one of: &<>()@^|
>         - there are one or more whitespace characters between the
>           two quote characters
>         - the string between the two quote characters is the name
>           of an executable file.
> 
>     2.  Otherwise, old behavior is to see if the first character is
>         a quote character and if so, strip the leading character and
>         remove the last quote character on the command line, preserving
>         any text after the last quote character.

This is tricky to grok, but keep in mind that when invoking a command line on Windows, we (Foundation.NSTask / SwiftSubprocess) do manual quoting of the contents of the command line array to turn it into something that will "round trip" through CommandLineToArgvW. For example, note the difference:

```
C:\Users\jakepetroules>cmd /c echo "hello world"
"hello world"

C:\Users\jakepetroules>cmd /c "echo hello world"
hello world
```

I don't know if how cmd.exe decides to parse the remainder of its command line (as described above) is the same as CommandLineToArgvW.

It might be preferable to introduce a custom task action and update the timestamp via API rather than by invoking touch on the CLI, to ensure robustness with odd filenames. That could be a separate PR though.

daveinglis:
@swift-ci test

Sources/SWBCore/SpecImplementations/Tools/TouchTool.swift
daveinglis: ok, I think I will open and Issue suggesting that this task be converted to a task that uses API instead of command line utils. since it looks like paths on windows can still have issues.

daveinglis: https://github.com/swiftlang/swift-build/issues/857

PR #854, opened 2025-10-08
[TEST-ONLY] Fix CASPath in swiftCachingSwiftPM()

cachemeifyoucan:
Change CASPath used in `swiftCachingSwiftPM` to be in the temp directory like the other tests. Current path might be complained by certain OS configuration about permission issues.


cachemeifyoucan:
@swift-ci please test

PR #859, opened 2025-10-20
Include child diagnostics with source locations for missing header dependencies

bob-wilson:
The updated Clang "print-headers-direct-per-file" tracing feature includes source locations. Missing header diagnostics should include child diagnostics to show the locations of those includes.

rdar://162322751


bob-wilson:
I think this change is ready except that it requires removing the "fake" feature flag for "print-headers-direct-per-file" (rdar://161741746), so that we actually enable this only when Clang supports the updated tracing.

owenv (approved):

PR #863, opened 2025-10-20
fix required SDKs for validateModuleDependenciesClang

owenv:
This test has sources which require a macOS SDK


owenv:
@swift-ci test

neonichu (approved):

PR #865, opened 2025-10-22
Unify stub binary processing for XCTRunner/messages apps/watch apps

owenv:
Refactor the processing of stub binaries so messages and watch stubs use the same logic as XCTRunner

### DIFF
--- a/Sources/SWBTaskConstruction/TaskProducers/TaskProducer.swift
+++ b/Sources/SWBTaskConstruction/TaskProducers/TaskProducer.swift
@@ -818,6 +818,31 @@ public class TaskProducerContext: StaleFileRemovalContext, BuildFileResolution
         return false
     }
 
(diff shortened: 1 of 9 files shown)


owenv:
@swift-ci test

Sources/SWBTaskConstruction/TaskProducers/TaskProducer.swift
```diff
@@ -818,6 +818,31 @@ public class TaskProducerContext: StaleFileRemovalContext, BuildFileResolution
         return false
     }
 
+    public func availableMatchingArchitecturesInStubBinary(at stubBinary: Path, requestedArchs: [String]) -> [String] {
```
jakepetroules: nit: I'd make this async since getCachedMachOInfo should eventually be so as well due to file system access

owenv:
@swift-ci test

neonichu (approved):

PR #867, opened 2025-10-23
Fix static linking on windows

daveinglis:
- xcspec condition needs to test for toplevel product type.
- update objlib response file to use .o file in the Products dir instead of the Intermediates dir.

### DIFF
--- a/Sources/SWBTaskExecution/TaskActions/ObjectLibraryAssemblerTaskAction.swift
+++ b/Sources/SWBTaskExecution/TaskActions/ObjectLibraryAssemblerTaskAction.swift
@@ -39,7 +39,7 @@ public final class ObjectLibraryAssemblerTaskAction: TaskAction {
             _ = try await options.inputs.concurrentMap(maximumParallelism: 10) { input in
                 try executionDelegate.fs.copy(input, to: options.output.join(input.basename))
             }
-            let args = options.inputs.map { $0.strWithPosixSlashes }
+            let args = options.inputs.map { options.output.join($0.basename).strWithPosixSlashes }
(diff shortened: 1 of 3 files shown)


daveinglis:
@swift-ci test

jakepetroules (approved):

Sources/SWBWindowsPlatform/Specs/Windows.xcspec
```diff
@@ -109,11 +109,6 @@
         Domain = windows;
         Type = ProductType;
         Identifier = org.swift.product-type.common.object;
-        BasedOn = org.swift.product-type.library.object;
```
owenv, line 112: Is this left over from debugging or is it needed?

Sources/SWBTaskExecution/TaskActions/ObjectLibraryAssemblerTaskAction.swift
```diff
@@ -39,7 +39,7 @@ public final class ObjectLibraryAssemblerTaskAction: TaskAction {
             _ = try await options.inputs.concurrentMap(maximumParallelism: 10) { input in
                 try executionDelegate.fs.copy(input, to: options.output.join(input.basename))
             }
-            let args = options.inputs.map { $0.strWithPosixSlashes }
```
owenv, line 42: Thanks! wonder if we should add a test to check the response file contents is correct here

Sources/SWBWindowsPlatform/Specs/Windows.xcspec
daveinglis: So linking with static libraries seems to be much better than using the objlib artifacts, as you don’t get the linker warnings on duplicate symbols

daveinglis:
@swift-ci test

PR #870, opened 2025-10-27
Fix platform requirements of installloc directory tests

matthewseaman:
Changing tests `embeddedDirectoriesInResources` and `embeddedDirectoriesInCopyFiles` to only require macOS, and ensuring requirement annotations match.


matthewseaman:
@swift-ci please test

PR #874, opened 2025-10-29
Dump dependency information during the build

neonichu:
- Extend `BuildDependencyInfo` with module/header dependencies
- Extend `ValidateDependenciesTaskAction` to dump per target dependency information in the new format
- The new behavior is opt-in behind a `DUMP_DEPENDENCIES` build setting

### DIFF
--- a/Sources/SWBTaskExecution/TaskActions/ValidateDependenciesTaskAction.swift
+++ b/Sources/SWBTaskExecution/TaskActions/ValidateDependenciesTaskAction.swift
@@ -11,9 +11,11 @@
 //===----------------------------------------------------------------------===//
 
 import Foundation
+
 public import SWBCore
-import SWBUtil
+internal import SWBMacro
 internal import SWBProtocol
+import SWBUtil
 
 public final class ValidateDependenciesTaskAction: TaskAction {
     public override class var toolIdentifier: String {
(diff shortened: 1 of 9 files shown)


neonichu:
@swift-ci please test

Sources/SWBTaskExecution/TaskActions/ValidateDependenciesTaskAction.swift
```diff
@@ -141,4 +150,57 @@ public final class ValidateDependenciesTaskAction: TaskAction {
...
         return .succeeded
     }
+
+    private func dumpDependenciesIfNeeded(imports: [DependencyValidationInfo.Import], includes: [DependencyValidationInfo.Include], for target: ConfiguredTarget?, context: BuildRequestContext) throws {
+        guard let target else {
+            return
+        }
+
+        let settings = context.getCachedSettings(target.parameters, target: target.target)
+        guard settings.globalScope.evaluate(BuiltinMacros.DUMP_DEPENDENCIES) else {
```
owenv: Instead of recomputing settings during the build, it would be better to pass this as an arg to the task action on its command line

neonichu:
> When the new setting is not enabled, will this produce the same output format as before?

This adds a new key `dependencies` as a peer to the existing `inputs` one, but doesn't change the format of any existing keys otherwise. Versioning is probably a good idea either way, but my guess is that extra keys are typically ignored by parsers, so this change should be forward compatible.

I don't have a clear idea how the format should look like yet, so I was thinking to make additive changes like this at first and once we have a clearer picture, we could try to consolidate the existing structure if needed.

neonichu:
@swift-ci please test

neonichu:
@swift-ci please test

neonichu:
> Versioning is probably a good idea either way

Actually, we won't need a version here, yet, since we don't serialize any dependencies as part of the existing command. The new format is only utilized for the output generated during the build.

PR #875, opened 2025-10-29
Optimize matchesFilenamePattern

Catfish-Man:
A trace I took today shows matchesFilenamePattern spending a surprising amount of time doing some things that don't seem necessary. I haven't verified the speedup yet (figuring out how to get it to run the specific code path I sampled is proving a bit tricky), but this should be a noticeable improvement. I'll annotate the changes inline with what they're for.

Fixes rdar://164182287

### DIFF
--- a/Sources/SWBUtil/Path.swift
+++ b/Sources/SWBUtil/Path.swift
@@ -80,7 +80,7 @@ public struct Path: Serializable, Sendable {
     }
     @inline(__always) public static func firstPathSeparatorIndex(in str: some StringProtocol, separators: (some Collection<Character>)?) -> String.Index? {
         guard let separators else {
-            return str.utf8.firstIndex(where: { Path.isUTF8PathSeparator($0, separators) })
+            return str.utf8.firstIndex(where: { Path.isUTF8PathSeparator($0, separators: separators) })
         }
         return str.firstIndex(where: { separators.contains($0) })
     }


Sources/SWBUtil/fnmatch.swift
```diff
@@ -36,16 +36,6 @@ private enum RangeStatus {
     case error
 }
 
-private extension StringProtocol {
```
Catfish-Man, line 39: This was unused after my other changes

Sources/SWBUtil/fnmatch.swift
```diff
@@ -54,7 +44,7 @@ private extension StringProtocol {
 /// - returns: `true` if the pattern matches the input, `false` otherwise.
 ///
 /// - note: On Windows and when using the ``FnmatchOptions/pathname`` option, both separators (`/` and `\`) are recognized (see note on pattern parameter).
-public func fnmatch(pattern: some StringProtocol, input: some StringProtocol, options: FnmatchOptions = .default, pathSeparators: Set<Character> = Path.pathSeparators) throws
```
Catfish-Man, line 57: `some Collection` lets us pass in non-Set types efficiently. Having a whole separate allocation to store 1-2 items is wasteful.

Defaulting to nil (sorry about the weird spelling of nil, it couldn't infer the type properly otherwise) lets us use the new default behavior that avoids lookups entirely.

Sources/SWBUtil/fnmatch.swift
```diff
@@ -76,32 +66,32 @@ public func fnmatch(pattern: some StringProtocol, input: some StringProtocol, op
...
-            if options.contains(.pathname) && pathSeparators.contains(_sc) {
+            if options.contains(.pathname) && Path.isUTF8PathSeparator(_sc, separators: pathSeparators) {
                 if backtrack() {
                     return false
                 }
             }
             input = input.dropFirst()
         case "*":
-            var p = pattern.first
-            while pattern.first == "*" {
```
Catfish-Man, line 90: "*" is always a single grapheme, so using utf8 will get us the same results as characters

Sources/SWBUtil/fnmatch.swift
```diff
@@ -76,32 +66,32 @@ public func fnmatch(pattern: some StringProtocol, input: some StringProtocol, op
...
+            while pattern.utf8.first == UInt8(ascii: "*") {
                 // consume multiple '*' in pattern
                 pattern = pattern.dropFirst()
-                p = pattern.first
+                p = pattern.utf8.first
             }
             if p == nil {
                 if options.contains(.pathname) {
                     // make sure input does not have any more path separators
-                    return input.firstIndex(matching: pathSeparators) == nil ? true : false
```
Catfish-Man, line 98: I really wanted to delete the pathSeparators parameter entirely, but it's used for testing Windows on non-Windows hosts

Sources/SWBUtil/fnmatch.swift
```diff
@@ -110,14 +100,14 @@ public func fnmatch(pattern: some StringProtocol, input: some StringProtocol, op
             bt_pattern = pattern
             bt_input = input
         case "[":
-            guard let _sc = input.first else {
```
Catfish-Man, line 113: Reordering these checks lets us do a quick utf8 check without having to construct a Character (since those are String under the hood)

Sources/SWBUtil/Path.swift
```diff
@@ -69,18 +69,45 @@ public struct Path: Serializable, Sendable {
     /// The system path separator.
     #if os(Windows)
     public static let pathSeparator = Character("\\")
-    public static let pathSeparatorUTF8 = UInt8(ascii: "\\")
```
Catfish-Man, line 72: Using an inline computed var instead of a stored let saves a little memory and a little indirection. Is it actually a relevant amount? Not really.

Sources/SWBUtil/Path.swift
```diff
@@ -69,18 +69,45 @@ public struct Path: Serializable, Sendable {
...
-    public static let pathSeparatorsUTF8 = Set([UInt8(ascii: "\\"), UInt8(ascii: "/")])
+    @inline(__always) public static var pathSeparatorUTF8: UInt8 { UInt8(ascii: "\\") }
     public static let pathEnvironmentSeparator = Character(";")
-    public static let pathSeparators = Set("\\/")
```
Catfish-Man, line 75: Here we replace a full Set lookup of Strings with a few integer comparisons. The non-default separators path is probably slower, but there's no reason I can see to care about that.

Also removes a heap allocation and a lazy initialization.

Sources/SWBUtil/Path.swift
```diff
@@ -69,18 +69,45 @@ public struct Path: Serializable, Sendable {
...
+    @inline(__always) public static func isUTF8PathSeparator(_ char: UInt8) -> Bool {
+        guard let separators else {
+            return char == pathSeparatorUTF8 || char == UInt8(ascii: "/")
+        }
+        // This is a bit inefficient, but separators should always be nil outside of tests
+        return separators.contains(String(decoding: CollectionOfOne(char), as: UTF8.self))
+    }
+    public static func firstPathSeparatorIndex(in str: some StringProtocol, separators: (some Collection<Character>)?) -> String.Index? {
```
Catfish-Man: I pulled this out into its own function so that I could use firstIndex(where:) (which has to deal with a closure) on Windows, and index(of:) everywhere else, since only Windows has more than one value to check for

Sources/SWBUtil/Path.swift
```diff
@@ -69,18 +69,45 @@ public struct Path: Serializable, Sendable {
...
+    }
+    @inline(__always) public static func firstPathSeparatorIndex(in str: some StringProtocol, separators: (some Collection<Character>)?) -> String.Index? {
+        guard let separators else {
+            return str.utf8.index(of: pathSeparatorUTF8)
+        }
+        return str.firstIndex(where: { separators.contains($0) })
+    }
     #endif
 
+    @inline(__always) public static func isPathSeparator(_ char: Character, separators: (some Collection<Character>)?) -> Bool {
```
Catfish-Man, line 106: Again, replacing set lookups with integer comparisons

Sources/SWBUtil/Path.swift
```diff
@@ -717,9 +744,10 @@ public struct Path: Serializable, Sendable {
             var numComponents = 0
             var isInPathComponent = false
             var nextCharacterIsEscaped = false
-            for idx in pattern.indices {
+            for byte in pattern.utf8 {
```
Catfish-Man, line 747: Everything we're comparing this to is a single grapheme, so we can skip grapheme semantics here

Sources/SWBUtil/Path.swift
```diff
@@ -717,9 +744,10 @@ public struct Path: Serializable, Sendable {
...
                 // Skip over path separators, unless they're escaped.
-                if pattern[idx] == Path.pathSeparator {
+                //TODO: should this (and other similar uses) be Path.isUTF8PathSeparator(byte) instead for Windows?
```
Catfish-Man, line 749: I'd like feedback on this please. To my eye it looks like a spot where Windows would want us to check both path separators, but I wasn't sure, and the old code didn't do so.

Sources/SWBUtil/Path.swift
```diff
@@ -781,7 +810,7 @@ public struct Path: Serializable, Sendable {
         }
 
         // Create a string from the first index we found to the end of the path.
-        let trailingSubpath = String(self.str[first..<self.str.endIndex])
```
Catfish-Man, line 784: This just got converted back to a Substring immediately in the callee, no reason not to leave it as one to begin with.

Tests/SWBUtilTests/FnmatchTests.swift
```diff
@@ -247,7 +247,7 @@ import SWBUtil
 
     @Test(arguments: [true, false])
     func pathnameMatch(isWindows: Bool) throws {
-        let separators = isWindows ? Set("\\/") : Set([Character("/")])
```
Catfish-Man, line 250: Set("\\/") looks so much like a Set containing one two-character-long String that I got confused. The thing it's passed to is generic now, so let's save the next person the wasted time.

Catfish-Man:
@swift-ci please test

Sources/SWBUtil/Path.swift
jakepetroules: It's hard to say. Generally you wouldn't ever want to match \ without also looking for /, but I think \ is also used as an escape sequence within the pattern here, so I'm not sure...

owenv:
Not sure what's up with the Windows failure but the rest of this lgtm generally

Sources/SWBUtil/Path.swift
```diff
@@ -69,18 +69,45 @@ public struct Path: Serializable, Sendable {
...
+    @inline(__always) public static func isUTF8PathSeparator(_ char: UInt8, separators: (some Collection<Character>)? = ([Character]?).none) -> Bool {
+        guard let separators else {
+            return char == pathSeparatorUTF8 || char == UInt8(ascii: "/")
+        }
+        // This is a bit inefficient, but separators should always be nil outside of tests
+        return separators.contains(String(decoding: CollectionOfOne(char), as: UTF8.self))
+    }
+    @inline(__always) public static func firstPathSeparatorIndex(in str: some StringProtocol, separators: (some Collection<Character>)?) -> String.Index? {
+        guard let separators else {
+            return str.utf8.firstIndex(where: Path.isUTF8PathSeparator)
```
jakepetroules: This should be:


```suggestion
            return str.utf8.firstIndex(where: { Path.isUTF8PathSeparator($0, separators) })
```

jakepetroules:
> Not sure what's up with the Windows failure but the rest of this lgtm generally

https://github.com/swiftlang/swift-build/pull/875/files#r2478828056

Sources/SWBUtil/Path.swift
```diff
@@ -69,18 +69,45 @@ public struct Path: Serializable, Sendable {
...
+    @inline(__always) public static func isUTF8PathSeparator(_ char: UInt8, separators: (some Collection<Character>)? = ([Character]?).none) -> Bool {
+        guard let separators else {
+            return char == pathSeparatorUTF8 || char == UInt8(ascii: "/")
+        }
+        // This is a bit inefficient, but separators should always be nil outside of tests
+        return separators.contains(String(decoding: CollectionOfOne(char), as: UTF8.self))
+    }
+    @inline(__always) public static func firstPathSeparatorIndex(in str: some StringProtocol, separators: (some Collection<Character>)?) -> String.Index? {
+        guard let separators else {
+            return str.utf8.firstIndex(where: { Path.isUTF8PathSeparator($0, separators) })
```
jakepetroules: Whoops, missing argument label:

```suggestion
            return str.utf8.firstIndex(where: { Path.isUTF8PathSeparator($0, separators: separators) })
```

Catfish-Man: lol no worries, that catches me constantly when iterating on stuff directly in CI

neonichu (approved):

Catfish-Man:
@swift-ci please test

Catfish-Man:
Linux SDK build failure is "** ERROR: Signature verification failed", which seems unrelated?

Not entirely sure what the remaining macOS failure is, it just says Error: Process completed with exit code 1. after a bunch of passing tests.

jakepetroules:
A few of the CI jobs can be flaky from time to time. I re-ran them and they passed. Should be good to go now. Thanks for the fix!

Sources/SWBUtil/fnmatch.swift
jamieQ: > sorry about the weird spelling of nil, it couldn't infer the type properly otherwise

is that a regression or something? if i spell it like:

```swift
pathSeparators: (some Collection<Character>)? = nil
```

it seems like it works... at least locally & via [godbolt](https://swift.godbolt.org/z/GW6Mb84xM).

PR #880, opened 2025-11-01
Copy Files phase should automatically use the correct lproj directory even if applying build rules

matthewseaman:
`CopyFilesTaskProducer` now adds the lproj directory automatically when copying a file from a variant group.

It already did this in the normal case, but not when `APPLY_RULES_IN_COPY_FILES` was set.


matthewseaman:
@swift-ci please test

matthewseaman:
@swift-ci test

PR #882, opened 2025-11-03
Replace provisional tasks with task validity criteria

owenv:
validity criteria is a property of the task itself, and is immutable. This eliminates mutable state in ConstructedTask and allows us to annotate PlannedTask as Sendable. IMO it's also easier to understand the refactored code.


owenv:
@swift-ci test

Sources/SWBCore/PlannedTask.swift
```diff
@@ -258,11 +251,14 @@ public final class GateTask: PlannedTask {
...
-    public let alwaysExecuteTask = false
+    public var alwaysExecuteTask: Bool { false }
 
     public var priority: TaskPriority { .gate }
 
-    public let repairViaOwnershipAnalysis: Bool = false
+    public var repairViaOwnershipAnalysis: Bool { false }
+
+    /// Gate tasks never have validity criteria.
+    public var validityCriteria: (any TaskValidityCriteria)? { nil }
```
jakepetroules, line 261: nit: would it be better modeled as non-nil, with tasks able to vend an "unconditional" validity criteria?

owenv: I implemented it this way so we're not calling through to a protocol implementation with `return true` for every single task in the graph, but I haven't actually measured how expensive that would be

neonichu (approved):

PR #883, opened 2025-11-03
[SwiftPreview] Fix SwiftUI preview when using caching and response file

cachemeifyoucan:
Fix a bug that swift preview can fail when using compilation caching and command-line is too long (so response file is used).

rdar://162547707


cachemeifyoucan:
@swift-ci please test

owenv (approved):

Issue #885, opened 2025-11-03
Provide API to set number of parallel jobs


bkhouri:
As a consumer of SwiftBuild, I want to be able to programmatically inform SwiftBuild the number of jobs to spawn in parallel during the build process.

PR #888, opened 2025-11-03
Add SWBBuildRequestAPI to set llbuild scheduler width

owenv:
closes https://github.com/swiftlang/swift-build/issues/885


owenv:
@swift-ci test

PR #427, opened 2025-04-21
Simplify a few API calls in FSProxy

jakepetroules:
These now go through Foundation rather than Win32 APIs directly. This simplifies Unicode and long path handling in particular.

For realpath I left the POSIX implementation in place for non-Windows for now since it's not clear that has equivalent behavior. touch/setFileTimestamp now differ in that they no longer set atime, but virtually nothing uses atime.

PR #889, opened 2025-11-04
Fix realpath to do the right thing on all platforms

jakepetroules:
This regressed as part of #427, which attempted to simplify API calls in FSProxy to use Foundation rather than bespoke implementations going directly to Win32/POSIX API.

However, `standardizingPath` on Windows doesn't actually handle 8.3 filenames correctly, and so Swift Build doesn't know (for example) that C:\Users\JAKEPE~1 and C:\Users\jakepetroules are the same path.

Use the canonicalPathKey from URLResourceValues to canonicalize the path appropriately on both platforms, which under the hood uses GetFinalPathNameByHandleW on Windows and realpath on POSIX, matching the previous behavior.

With this change, the androidCommandLineTool test now passes on Windows, which was previously failing due to differences between paths in 8.3 form vs long form.

### DIFF
--- a/Sources/SWBUtil/FSProxy.swift
+++ b/Sources/SWBUtil/FSProxy.swift
@@ -757,16 +757,15 @@ class LocalFS: FSProxy, @unchecked Sendable {
     }
 
     func realpath(_ path: Path) throws -> Path {
-        #if os(Windows)
-        guard exists(path) else {
+        if path.isAbsolute && !exists(path) {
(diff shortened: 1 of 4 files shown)


jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

PR #890, opened 2025-11-04
Change how libraries are specified to the linker when using searched libs

### LINKED ISSUE #879: LdLinkerSpec::computeLibraryArgs needs to be fixed to remove the platform specifics on library prefixes and extensions
They way computeLibraryArgs written it results in libraries been specified with fullpaths resulting in not being able to link with any dll on windows (the dll it specified instead of the lib to the linker).

daveinglis:
- remove the platform specifics from computeLibraryArgs (we cannot assume
  that all libraries have a lib prefix and what there suffix is.) So we
  now use the FileType prefix and remove any suffix when using
  searchPathFlagsForLD, moving this into the LinkerSpec.LibrarySpecifier
  extension, this allows for proper searching of libraries, and linking
  of dynamic libraries (especially on Windows).

closes #879

### DIFF
--- a/Sources/SWBCore/SpecImplementations/LinkerSpec.swift
+++ b/Sources/SWBCore/SpecImplementations/LinkerSpec.swift
@@ -89,11 +89,12 @@ open class LinkerSpec : CommandLineToolSpec, @unchecked Sendable {
         /// The path to the privacy file, if one exists.
         public let privacyFile: Path?
 
-        public init(kind: Kind, path: Path, mode: Mode, useSearchPaths: Bool, swiftModulePaths: [String: Path], swiftModuleAdditionalLinkerArgResponseFilePaths: [String: Path], explicitDependencies: [Path] = [], topLevelItemPath: Path? = nil, dsymPath: Path? = nil, xcframeworkSourcePath: Path? = nil, privacyFile: Path? = nil) {
+        public let libPrefix: String?
+
+        public init(kind: Kind, path: Path, mode: Mode, useSearchPaths: Bool, swiftModulePaths: [String: Path], swiftModuleAdditionalLinkerArgResponseFilePaths: [String: Path], prefix: String? = nil, explicitDependencies: [Path] = [], topLevelItemPath: Path? = nil, dsymPath: Path? = nil, xcframeworkSourcePath: Path? = nil, privacyFile: Path? = nil) {
             self.kind = kind
             self.path = path
             self.mode = mode
-            self.useSearchPaths = useSearchPaths
             self.swiftModulePaths = swiftModulePaths
             self.swiftModuleAdditionalLinkerArgResponseFilePaths = swiftModuleAdditionalLinkerArgResponseFilePaths
             self.explicitDependencies = explicitDependencies
@@ -101,6 +102,10 @@ open class LinkerSpec : CommandLineToolSpec, @unchecked Sendable {
             self.dsymPath = dsymPath
             self.xcframeworkSourcePath = xcframeworkSourcePath
             self.privacyFile = privacyFile
+            self.libPrefix = prefix
+            // Only use search paths when no prefix is required or when the prefix matches
+            let hasValidPrefix = libPrefix.map { path.basename.hasPrefix($0) } ?? true
+            self.useSearchPaths = hasValidPrefix && useSearchPaths
         }
     }
 
--- a/Sources/SWBCore/SpecImplementations/Tools/LinkerTools.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/LinkerTools.swift
@@ -1312,41 +1312,12 @@ public final class LdLinkerSpec : GenericLinkerSpec, SpecIdentifierType, @unchec
     private static func computeLibraryArgs(_ libraries: [LibrarySpecifier], scope: MacroEvaluationScope) -> (args: [String], inputs: [Path]) {
         // Construct the library arguments.
         return libraries.compactMap { specifier -> (args: [String], inputs: [Path]) in
-            let basename = specifier.path.basename
-
-            // FIXME: This isn't a good system, we need to redesign how we talk to the linker w.r.t. search paths and our notion of paths.
             switch specifier.kind {
-            case .static:
-                if specifier.useSearchPaths, basename.hasPrefix("lib"), basename.hasSuffix(".a") {
-                    return (specifier.searchPathFlagsForLd(basename.withoutPrefix("lib").withoutSuffix(".a")), [])
-                }
-                return (specifier.absolutePathFlagsForLd(), [specifier.path])
-            case .dynamic:
-                let suffix = ".\(scope.evaluate(BuiltinMacros.DYNAMIC_LIBRARY_EXTENSION))"
-                if specifier.useSearchPaths, basename.hasPrefix("lib"), basename.hasSuffix(suffix) {
-                    return (specifier.searchPathFlagsForLd(basename.withoutPrefix("lib").withoutSuffix(suffix)), [])
-                }
-                return (specifier.absolutePathFlagsForLd(), [specifier.path])
-            case .textBased:
-                if specifier.useSearchPaths, basename.hasPrefix("lib"), basename.hasSuffix(".tbd") {
-                    // .merge and .reexport are not supported for text-based libraries.
-                    return (specifier.searchPathFlagsForLd(basename.withoutPrefix("lib").withoutSuffix(".tbd")), [])
-                }
-                return (specifier.absolutePathFlagsForLd(), [specifier.path])
-            case .framework:
-                let frameworkName = Path(basename).withoutSuffix
+            case .static, .dynamic, .textBased, .framework:
                 if specifier.useSearchPaths {
-                    return (specifier.searchPathFlagsForLd(frameworkName), [])
+                    return (specifier.searchPathFlagsForLd(), [])
                 }
-                let absPathArgs = specifier.absolutePathFlagsForLd()
-                let returnPath: Path
-                if let pathArg = absPathArgs.last, Path(pathArg).basename == frameworkName {
-                    returnPath = Path(pathArg)
-                }
-                else {
-                    returnPath = specifier.path
-                }
-                return (absPathArgs, [returnPath])
+                return (specifier.absolutePathFlagsForLd(), [specifier.path])
             case .object:
                 // Object files are added to linker inputs in the sources task producer.
                 return ([], [])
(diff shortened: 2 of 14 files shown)


daveinglis:
@swift-ci test

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test

Sources/SWBCore/SpecImplementations/Tools/LinkerTools.swift
```diff
@@ -1724,9 +1701,9 @@ public final class LibtoolLinkerSpec : GenericLinkerSpec, SpecIdentifierType, @u
                     delegate.warning("Product \(cbc.output.basename) cannot weak-link \(specifier.kind) \(basename)")
                 }
 
-                if specifier.useSearchPaths, basename.hasPrefix("lib"), basename.hasSuffix(".a") {
+                if specifier.useSearchPaths {
```
owenv, line 1738: Should we emit some kind of warning here if the library name doesn't have one of the prefixes we think will allow it to be found via search paths?

Sources/SWBCore/SpecImplementations/Tools/LinkerTools.swift
```diff
@@ -1559,35 +1530,41 @@ public final class LdLinkerSpec : GenericLinkerSpec, SpecIdentifierType, @unchec
 
 /// Extensions to `LinkerSpec.LibrarySpecifier` specific to the dynamic linker.
 fileprivate extension LinkerSpec.LibrarySpecifier {
-    func searchPathFlagsForLd(_ name: String) -> [String] {
+    func searchPathFlagsForLd() -> [String] {
+        let strippedName: String
```
owenv, line 1563: Same question here, should we warn if we get a filename format we don't recognize here / don't think will be found?

Sources/SWBCore/SpecImplementations/LinkerSpec.swift
```diff
@@ -101,6 +103,7 @@ open class LinkerSpec : CommandLineToolSpec, @unchecked Sendable {
             self.dsymPath = dsymPath
             self.xcframeworkSourcePath = xcframeworkSourcePath
             self.privacyFile = privacyFile
+            self.libPrefix = prefixes.first
```
owenv: Why do we only consider the first element here?

owenv:
overall this looks great, just have a few questions

owenv:
we should also run this through cross-repo testing before merging

Sources/SWBCore/SpecImplementations/LinkerSpec.swift
daveinglis: Ya, I was thinking about that and I just couldn't think of a case where there would ever be more that one so I forego the extract logic of handling multiple and just went with the lazy handle one method.  Do you feel its worth handling multiple?

Sources/SWBCore/SpecImplementations/Tools/LinkerTools.swift
daveinglis: Hmm, rethinking this.... I think we need to fallback to using absolute path linking when that happens, this would be the same as how it was before....

daveinglis: There are also tests that would have this warning (and fail due to it), but since we will fallback to an absolute path, not sure the warning it necessary now.

owenv: falling back w/ no warning makes sense to me

Sources/SWBCore/SpecImplementations/LinkerSpec.swift
owenv: Only handling one sounds ok to me, but I think we should then either only allow one in the spec or error out if we parse more than one

owenv (approved):
Looks like there is one more test failure to resolve, but this lgtm, thanks for driving it

✘ Test unitTestRunnerTarget() recorded an issue at UnitTestTaskConstructionTests.swift:388:37: Expectation failed: matched
↳ unexpected failure matching '/tmp/Test/aProject/build/Debug-linux/UnitTestTarget.so' against pattern or(SWBTestSupport.StringPattern.suffix("/libUnitTestTarget.so"), SWBTestSupport.StringPattern.suffix("\\UnitTestTarget.dll"))

daveinglis:
doh, put the build setting on the wrong target for that test...

Sources/SWBTaskConstruction/TaskProducers/BuildPhaseTaskProducers/SourcesTaskProducer.swift
```diff
@@ -364,14 +364,20 @@ package final class SourcesTaskProducer: FilesBasedBuildPhaseTaskProducerBase, F
...
+            // On Windows, always use search paths to ensure import libraries (.lib) are found correctly instead of DLLs.
             let useSearchPaths: Bool
             switch buildFile.buildableItem {
             case .reference:
                 useSearchPaths = true
             case .targetProduct(guid: let guid):
                 if let referenceTarget = self.context.workspaceContext.workspace.target(for: guid) {
                     let referenceProject = self.context.workspaceContext.workspace.project(for: referenceTarget)
-                    useSearchPaths = referenceProject === self.context.project
+                    // Always use search paths on Windows to ensure .lib files are found instead of .dll files
```
jakepetroules: How about replacing the .dll suffix with .lib instead? Ideally we want to use search paths less and less.

Sources/SWBTestSupport/RunDestinationTestSupport.swift
```diff
@@ -328,7 +328,10 @@ extension RunDestinationInfo {
         case .elf:
             environment.prependPath(key: "LD_LIBRARY_PATH", value: toolchain.path.join("usr/lib/swift/\(platform)").str)
         case .pe:
-            environment.prependPath(key: .path, value: core.developerPath.path.join("Runtimes").join(toolchain.version.description).join("usr/bin").str)
+            let currentPath = Environment.current[.path] ?? ""
```
jakepetroules: Why did this change to start reading from the current environment instead of the given one? This doesn't seem right. We should be aiming to reduce usage of globals throughout the codebase as much as possible.

Also, why was `throws` added? Nothing in here throws.

Sources/SWBTaskConstruction/TaskProducers/BuildPhaseTaskProducers/SourcesTaskProducer.swift
daveinglis: I can just that try...

Sources/SWBTestSupport/RunDestinationTestSupport.swift
daveinglis: we need path so that tests that run other tests can find the testing dlls on windows, will fix up the throws, must of got added at some point when debugging.

jakepetroules: If you really need to pick up the process environment then try making this change to the signature instead:

```swift
package func hostRuntimeEnvironment(_ core: Core, initialEnvironment: Environment = .current) -> Environment {
```

That way the function remains hermetic.

Sources/SWBTaskConstruction/TaskProducers/BuildPhaseTaskProducers/SourcesTaskProducer.swift
```diff
@@ -507,16 +507,25 @@ package final class SourcesTaskProducer: FilesBasedBuildPhaseTaskProducerBase, F
...
                     swiftModulePaths: swiftModulePaths,
                     swiftModuleAdditionalLinkerArgResponseFilePaths: swiftModuleAdditionalLinkerArgResponseFilePaths,
+                    prefix: fileType.prefix,
                     privacyFile: privacyFile
                 )
             } else if fileType.conformsTo(context.lookupFileType(identifier: "compiled.mach-o.dylib")!) {
+                let adjustedAbsolutePath: Path
+                // On Windows, ensure import libraries (.lib) are used instead of DLLs.
+                if context.sdkVariant?.llvmTargetTripleSys == "windows" {
+                        adjustedAbsolutePath = Path(absolutePath.withoutSuffix + ".lib")
```
jakepetroules: nit: this line got over-indented

Sources/SWBTaskConstruction/TaskProducers/BuildPhaseTaskProducers/SourcesTaskProducer.swift
```diff
@@ -507,16 +507,25 @@ package final class SourcesTaskProducer: FilesBasedBuildPhaseTaskProducerBase, F
                     useSearchPaths: useSearchPaths,
                     swiftModulePaths: swiftModulePaths,
                     swiftModuleAdditionalLinkerArgResponseFilePaths: swiftModuleAdditionalLinkerArgResponseFilePaths,
+                    prefix: fileType.prefix,
                     privacyFile: privacyFile
                 )
             } else if fileType.conformsTo(context.lookupFileType(identifier: "compiled.mach-o.dylib")!) {
+                let adjustedAbsolutePath: Path
+                // On Windows, ensure import libraries (.lib) are used instead of DLLs.
+                if context.sdkVariant?.llvmTargetTripleSys == "windows" {
```
jakepetroules: just to be extra cautious, also check that the existing suffix is (case insensitively) "dll"? I think if we for some reason had a .a, we wouldn't want to change the suffix.

owenv:
@swift-ci test

daveinglis:
> Thanks, this looks good.
> 
> I am curious though which specific path was in `Environment.current` that you needed the tests to pick up?

The paths to the testing DLLs under "%%LOCALAPPDATA%%\Programs\Swift\Platforms\0.0.0\Windows.platform\Developer\Library"

jakepetroules:
> > Thanks, this looks good.
> > I am curious though which specific path was in `Environment.current` that you needed the tests to pick up?
> 
> The paths to the testing DLLs under "%%LOCALAPPDATA%%\Programs\Swift\Platforms\0.0.0\Windows.platform\Developer\Library"

Ahhh, I see. Then actually can you put the environment back to `Environment()` and add this line inside hostRuntimeEnvironment?

```swift
if let path = core.platformRegistry.lookup(name: platform)?.platform?.path.join("Developer/Library") {
    environment.prependPath(key: .path, value: path.str)
}
```

I think that should do what you need. The disadvantage of the `.current` approach is that it doesn't guarantee that the testing DLLs are picked up from the same toolchain installation as the rest of the bits, which the above line of code I suggested, should.

EDIT: You'll also need to compute the subpath like Testing-6.2.0/usr/bin64 and XCTest-6.2.0/usr/bin64 based on the host architecture (see `Architecture.host`)

daveinglis:
@swift-ci test

jakepetroules:
@swift-ci test

jakepetroules:
Thanks, that looks good now!

BTW, if you end up needing to push again, it would be good to replace the `try!` with `try`. Otherwise can fix it up later.

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test

PR #892, opened 2025-11-05
Force linker to lld on amazon linux 2.

kcieplak:
There is a bug in the gold linker on amazon linux 2

See: https://sourceware.org/bugzilla/show_bug.cgi?id=23016

 It is triggered when linking relocatable objects built with  LLVM based compilers.

- Add gear to detect the os distribution and version
-  For amazon linux 2 force ALTERNATE_LINKER to lld
- Add tests mocking the various linux distributions

### DIFF
--- a/Sources/SWBUtil/ProcessInfo.swift
+++ b/Sources/SWBUtil/ProcessInfo.swift
@@ -265,7 +265,7 @@ public enum OperatingSystem: Hashable, Sendable {
                 return LinuxDistribution(kind: kind)
             }
         }
-        
+
         return nil
     }
 
@@ -419,3 +419,4 @@ extension FixedWidthInteger {
         return self != 0 ? self : other
     }
 }
+
--- a/Tests/SWBUtilTests/LinuxDistributionTests.swift
+++ b/Tests/SWBUtilTests/LinuxDistributionTests.swift
@@ -14,8 +14,7 @@ import Foundation
 import Testing
 import SWBTestSupport
 import SWBUtil
-
-@Suite
+@Suite(.skipHostOS(.windows))
 fileprivate struct LinuxDistributionTests {
 
     /// Test helper to create a mock filesystem with specific files


kcieplak:
Fixes https://github.com/swiftlang/swift-package-manager/issues/9249

kcieplak:
@swift-ci test

Tests/SWBUtilTests/LinuxDistributionTests.swift
```diff
@@ -0,0 +1,776 @@
...
+        for (filePath, content) in distributionFiles {
+            try fs.write(Path(filePath), contents: ByteString(encodingAsUTF8: content))
+        }
+
+        return try await operation(fs)
+    }
+
+    /// Test parsing various /etc/os-release formats for different distributions
+    @Test
+    func detectUbuntuFromOSRelease() async throws {
```
owenv, line 46: These tests are all duplicating the parsing code instead of testing the real implementation. If needed, I think we can just allow injecting a FileSystem into detectHostLinuxDistribution

Tests/SWBUtilTests/LinuxDistributionTests.swift
```diff
@@ -0,0 +1,776 @@
...
+
+        let ubuntuWithoutVersion = LinuxDistribution(kind: .ubuntu, version: nil)
+        #expect(ubuntuWithoutVersion.displayName == "Ubuntu")
+
+        let unknownWithVersion = LinuxDistribution(kind: .unknown, version: "1.0")
+        #expect(unknownWithVersion.displayName == "Unknown Linux 1.0")
+    }
+
+    @Test
+    func testLinuxDistributionHashableAndSendable() async throws {
```
owenv: These are synthesized impls, I don't think we need to test them

kcieplak: Changed to pass in the fs to the underlying `detectHostLinuxDistribution` function to exercise the actual parsing logic.

Sources/SWBUtil/ProcessInfo.swift
```diff
@@ -167,6 +227,110 @@ public enum OperatingSystem: Hashable, Sendable {
...
+
+    /// Detects the Linux distribution by examining system files with an injected filesystem
+    /// Start with the "generic" /etc/os-release then fallback
+    /// to various distribution named files.
+    public func detectHostLinuxDistribution(fs: any FSProxy) -> LinuxDistribution? {
+        // Try /etc/os-release first (standard)
+        let osReleasePath = Path("/etc/os-release")
+        if fs.exists(osReleasePath) {
+            if let osReleaseData = try? fs.read(osReleasePath),
+               let osRelease = String(data: Data(osReleaseData.bytes), encoding: .utf8) {
```
jakepetroules, line 243: nit: prefer the stdlib constructor instead of the Foundation one:

```suggestion
               let osRelease = String(decoding: Data(osReleaseData.bytes), as: UTF8.self) {
```

Tests/SWBUtilTests/LinuxDistributionTests.swift
```diff
@@ -0,0 +1,434 @@
...
+
+        try await withMockLinuxDistribution(osReleaseContent: osReleaseContent) { fs in
+            let operatingSystem = OperatingSystem.linux
+            let distribution = operatingSystem.detectHostLinuxDistribution(fs: fs)
+
+            // Unknown ID should map to nil
+            #expect(distribution == nil)
+        }
+    }
+}
```
jakepetroules: nit: files should end with newlines

jakepetroules (approved):
Nice, this looks great! Just some minor formatting feedback.

kcieplak:
@swift-ci test

Tests/SWBUtilTests/LinuxDistributionTests.swift
jakepetroules: Actually, can you please fix this before merging? #893 adds some validations

kcieplak:
@swift-ci test

PR #897, opened 2025-11-07
Merge dumped dependency info

neonichu:
Since #874, we are dumping per-target dependency information, this adds a new task which merges these for a given build request. The task is only created if there's at least one target in the graph that will dump dependency information, so should be a no-op for all builds right now.

Additionally, this declares the dependency info file as an output of `ValidateDependencies` which I neglected to do in my earlier PR.

### DIFF
--- a/Sources/SWBCore/SpecImplementations/Tools/BuildDependencyInfoSpec.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/BuildDependencyInfoSpec.swift
@@ -1,0 +1,40 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+private import Foundation
+public import SWBUtil
+import SWBMacro
+
+public final class BuildDependencyInfoSpec: CommandLineToolSpec, SpecImplementationType, @unchecked Sendable {
+    public static let identifier = "com.apple.tools.build-dependency-info"
+
+    public static func construct(registry: SpecRegistry, proxy: SpecProxy) -> Spec {
+        return Self.init(registry: registry)
+    }
+
+    public init(registry: SpecRegistry) {
+        let proxy = SpecProxy(identifier: Self.identifier, domain: "", path: Path(""), type: Self.self, classType: nil, basedOn: nil, data: ["ExecDescription": PropertyListItem("Merging build dependency info")], localizedStrings: nil)
+        super.init(createSpecParser(for: proxy, registry: registry), nil, isGeneric: false)
+    }
+
+    required init(_ parser: SpecParser, _ basedOnSpec: Spec?) {
+        super.init(parser, basedOnSpec, isGeneric: false)
+    }
+
+    override public func constructTasks(_ cbc: CommandBuildContext, _ delegate: any TaskGenerationDelegate) async {
+        fatalError("unexpected direct invocation")
+    }
+
+    public func createTasks(_ cbc: CommandBuildContext, _ delegate: any TaskGenerationDelegate, dumpDependencyPaths: [Path]) async {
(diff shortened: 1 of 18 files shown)


Sources/SWBTaskExecution/TaskActions/BuildDependencyInfoTaskAction.swift
```diff
@@ -0,0 +1,58 @@
...
+import SWBUtil
+
+public final class BuildDependencyInfoTaskAction: TaskAction {
+    public override class var toolIdentifier: String {
+        return "build-dependency-info"
+    }
+
+    public override func performTaskAction(_ task: any ExecutableTask, dynamicExecutionDelegate: any DynamicTaskExecutionDelegate, executionDelegate: any TaskExecutionDelegate, clientDelegate: any TaskExecutionClientDelegate, outputDelegate: any TaskOutputDelegate) async -> CommandResult {
+        let commandLine = Array(task.commandLineAsStrings)
+        guard commandLine.count >= 1, commandLine[0] == "builtin-build-dependency-info" else {
```
jakepetroules: For all new task actions, can you please use argument parser?

(see ObjectLibraryAssemblerTaskAction for example)

this also makes it easier to delineate arguments using flags rather than relying too much on positionals.

Sources/SWBTaskExecution/TaskActions/BuildDependencyInfoTaskAction.swift
```diff
@@ -0,0 +1,58 @@
...
+            var targets = [BuildDependencyInfo.TargetDependencyInfo]()
+            for dumpDependencyPath in commandLine.dropFirst() {
+                let dumpDependencyData = try Data(contentsOf: URL(fileURLWithPath: dumpDependencyPath))
+                let dumpDependencyInfo = try JSONDecoder().decode(BuildDependencyInfo.self, from: dumpDependencyData)
+                errors.append(contentsOf: dumpDependencyInfo.errors)
+                targets.append(contentsOf: dumpDependencyInfo.targets)
+            }
+
+            let dependencyInfo = BuildDependencyInfo(targets: targets, errors: errors)
+            let outputData = try JSONEncoder().encode(dependencyInfo)
```
jakepetroules: Please use `encoder.outputFormatting = [.prettyPrinted, .sortedKeys, .withoutEscapingSlashes]` to keep things readable/deterministic.

neonichu:
@swift-ci please test

neonichu:
Looks like the failure in `Static Linux Swift SDK Build (nightly-main - rhel-ubi9)` is a crash in SwiftPM?

Sources/SWBTaskConstruction/TaskProducers/WorkspaceTaskProducers/BuildDependencyInfoTaskProducer.swift
```diff
@@ -0,0 +1,62 @@
...
+        guard components.contains("build") else {
+            return []
+        }
+
+        let output = context.settings.globalScope.evaluate(BuiltinMacros.BUILD_DIR).join("BuildDependencyInfo.json")
+        let dumpDependencyPaths: [Path] = targetContexts.compactMap {
+            guard let target = $0.configuredTarget?.target as? SWBCore.StandardTarget else {
+                return nil
+            }
+            guard target.sourcesBuildPhase?.buildFiles.isEmpty == false else {
```
jakepetroules, line 38: Note there are some edge cases where empty targets still do some work (generation of the apple versioning .c file for example), I think there is code somewhere to handle that. If it doesn't hurt, you might just want to not skip based on this.

neonichu: You're right that we should handle this, but it seems like today the knowledge only exists inside `SourcesTaskProducer` and isn't readily accessible? Unless someone knows another way, I think I'd rather address this in a follow-up to keep a refactor of the logic in `SourcesTaskProducer` separate.

Sources/SWBCore/SpecImplementations/Tools/BuildDependencyInfoSpec.swift
```diff
@@ -0,0 +1,40 @@
...
+
+    required init(_ parser: SpecParser, _ basedOnSpec: Spec?) {
+        super.init(parser, basedOnSpec, isGeneric: false)
+    }
+
+    override public func constructTasks(_ cbc: CommandBuildContext, _ delegate: any TaskGenerationDelegate) async {
+        fatalError("unexpected direct invocation")
+    }
+
+    public func createTasks(_ cbc: CommandBuildContext, _ delegate: any TaskGenerationDelegate, dumpDependencyPaths: [Path]) async {
```
owenv, line 37: dumpDependencyPaths here could probably be cbc.inputs

PR #900, opened 2025-11-09
Add support for artifact bundles with static libraries

owenv:
- parsing for artifact bundle info.json metadata
- Propagation of search paths for libraries in artifact bundles via the global product plan. To support per-triple artifact bundle variants, this relies on a new __normalized_unversioned_triple macro condition
- linker library selection

### DIFF
--- a/Sources/SWBCore/Settings/Settings.swift
+++ b/Sources/SWBCore/Settings/Settings.swift
@@ -811,12 +811,12 @@ public final class Settings: PlatformBuildContext, Sendable {
     /// The information about the project model components from which these settings were constructed.
     public let constructionComponents: ConstructionComponents
 
-    public convenience init(workspaceContext: WorkspaceContext, buildRequestContext: BuildRequestContext, parameters: BuildParameters, project: Project, target: Target? = nil, purpose: SettingsPurpose = .build, provisioningTaskInputs: ProvisioningTaskInputs? = nil, impartedBuildProperties: [ImpartedBuildProperties]? = nil, includeExports: Bool = true, sdkRegistry: (any SDKRegistryLookup)? = nil) {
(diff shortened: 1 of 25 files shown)


owenv:
@swift-ci test

owenv:
@swift-ci test

Sources/SWBCore/Settings/Settings.swift
```diff
@@ -1576,6 +1580,52 @@ private class SettingsBuilder {
...
+                for artifact in artifactBundle.metadata.artifacts.values {
+                    switch artifact.type {
+                    case .staticLibrary:
+                        for variant in artifact.variants {
+                            guard let staticLibraryMetadata = variant.staticLibraryMetadata else {
+                                continue
+                            }
+                            var conditionSets: [MacroConditionSet?] = []
+                            if let triples = variant.supportedTriples {
+                                for triple in triples {
```
jakepetroules: Should the input triple here be normalized for comparison to the current triple?

Sources/SWBUtil/SwiftDriverTripleExtensions.swift
```diff
@@ -0,0 +1,28 @@
...
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftDriver
+
+/// Compares two triples for equality after normalization, ignoring OS versions.
+package func normalizedTriplesCompareDisregardingOSVersions(_ firstTripleString: String, _ secondTripleString: String) -> Bool {
```
jakepetroules, line 16: Does this handle Android triples correctly? They encode their version in the `environment` field rather than the `os` field like most triples do.

e.g. aarch64-unknown-linux-android28 vs arm64-apple-ios26.0-simulator

I think I had to handle them specially in SwiftPM's SwiftBuildSupport.

owenv: I added a test for this and it turned out it was already working

owenv:
@swift-ci test

owenv:
@swift-ci test macOS

PR #903, opened 2025-11-11
Do not pass index store options if indexing is disabled

ahoppen:
We were passing `-index-store-compress` to the compiler even though we don’t pass `-index-store-path` if `INDEX_STORE_ENABLE = YES` but `COMPILER_INDEX_STORE_ENABLE = NO`. This is because `CLANG_INDEX_STORE_ENABLE` still evaluated to `YES` and thus the `Condition` of `CLANG_INDEX_STORE_COMPRESS` also evaluated to `YES` but we didn’t pass `-index-store-path` because the condition of `CLANG_INDEX_STORE_ENABLE` evaluated to `NO`.

This was a misconception on my side when I added these build settings because I assumed that the `Condition` would influence the value of the build setting instead of just influencing `ComandLineArgs`.

I don’t like how we need to repeat the same condition in multiple build settings now but haven’t found a better solution.

rdar://164248532


jakepetroules:
@swift-ci test

PR #910, opened 2025-11-14
Cleanup targetArchitecture handling in macOS test run destination

owenv:
Resolve a longstanding FIXME related to the `.macOS` destination used in tests, and update the tests to consistently query targetArchitectures rather than hardcoding string values. This ensures that task construction tests use an arm64 destination on an arm64 host, which leads to less surprising behavior when authoring new tests


owenv:
@swift-ci test

Sources/SWBTestSupport/BuildOperationTester.swift
```diff
@@ -1575,7 +1575,7 @@ package final class BuildOperationTester {
         let operationParameters = buildRequest.parameters.replacing(activeRunDestination: runDestination, activeArchitecture: nil)
         let operationBuildRequest = buildRequest.with(parameters: operationParameters, buildTargets: [])
 
-        return try await checkBuild(runDestination: nil, buildRequest: buildRequest, operationBuildRequest: operationBuildRequest, persistent: persistent, sourceLocation: sourceLocation, body: body, performBuild: { try await $0.buildWithTimeout() })
```
jakepetroules, line 1578: Hmm, I don't remember if that was deliberate or not.

owenv: I undid this because I ended up not really touching the tests that use it

owenv:
@swift-ci test

owenv:
@swift-ci test

neonichu (approved):

PR #914, opened 2025-11-14
Correctly locate chown on all platforms

jakepetroules:
On Linux, chown is at /usr/bin/chown, while on BSDs it's at /usr/sbin/chown

It was also being incorrectly located based on the target platform, which is incorrect. Only the host platform matters for this lookup. This meant macOS to Android or WebAssembly cross compilation, FreeBSD to Linux cross compilation, or FreeBSD native compilation could have failed if the build graph contained any chown commands. Also use a search path so we don't have to hardcode knowledge of any specific platform for this.


jakepetroules:
@swift-ci test

owenv (approved):

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

PR #915, opened 2025-11-14
Ignore exit code when checking for system-level packages

jakepetroules:
FreeBSD's pkg returns 1 if a package isn't installed. Across all platforms, we really only care about parsing the tool's output, not whether it exited successfully or not. This allows tests to be properly skipped on FreeBSD when the requisite tool is not installed.


jakepetroules:
@swift-ci test

neonichu (approved):

PR #916, opened 2025-11-14
Don't skip coreLoadErrors on Linux

jakepetroules:
Whatever toolchain issue caused this to originally be skipped is no longer present.


jakepetroules:
@swift-ci test

owenv (approved):

PR #918, opened 2025-11-14
Stop passing -fobjc-arc to linker invocations

owenv:
We still need to pass this to the compiler when requested, but its primary purpose applied to linker invocations was to link arclite, which has been obsolete for a long time now. There's also no direct equivalent in the swiftc linker driver, so this fixes a link failure when that's used. -fobjc-link-runtime still gets passed for ObjC sources on platforms where it exists, so we don't lose that linkage.


owenv:
@swift-ci test

jakepetroules (approved):

plemarquand (approved):

owenv:
@swift-ci test

Issue #333, opened 2025-03-21
Build failure with SwiftPM package with spaces in directory name on Linux


cmcgee1024:
When building a SwiftPM package on a path with spaces in it there is a build failure on Linux.

Run this command from a swift-package manager Linux development environment:

```
swift run swift-build -v --build-system=swiftbuild --package-path=Fixtures/Miscellaneous/Spaces\ Fixture
```

It's expected that this should pass as it does for similar packages without spaces in the package path.

Instead, we get an error from clang:

```
info: Link Module Name 1_Module.o (aarch64)
    cd /swiftpm/Fixtures/Miscellaneous/Spaces\ Fixture
    /usr/bin/clang -target aarch64-unknown-linux-gnu -r --sysroot / -O0 -L/swiftpm/Fixtures/Miscellaneous/Spaces\ Fixture/.build/aarch64-unknown-linux-gnu/Intermediates.noindex/EagerLinkingTBDs/Debug-linux -L/swiftpm/Fixtures/Miscellaneous/Spaces\ Fixture/.build/aarch64-unknown-linux-gnu/Products/Debug-linux -L/usr/lib @/swiftpm/Fixtures/Miscellaneous/Spaces\ Fixture/.build/aarch64-unknown-linux-gnu/Intermediates.noindex/Spaces\ Fixture.build/Debug-linux/Module\ Name\ 1_Module.build/Objects-normal/aarch64/Module\ Name\ 1_Module.LinkFileList -nostdlib -rdynamic -o /swiftpm/Fixtures/Miscellaneous/Spaces\ Fixture/.build/aarch64-unknown-linux-gnu/Products/Debug-linux/Module\ Name\ 1_Module.o
error: unknown No such file or directory: '/swiftpm/Fixtures/Miscellaneous/Spaces' []
error: unknown No such file or directory: 'Fixture/.build/aarch64-unknown-linux-gnu/Intermediates.noindex/Spaces' []
error: unknown No such file or directory: 'Fixture.build/Debug-linux/Module' []
error: unknown No such file or directory: 'Name' []
error: unknown No such file or directory: '1_Module.build/Objects-normal/aarch64/Modules/Module_Name_1.o' []
error: unknown No such file or directory: '/swiftpm/Fixtures/Miscellaneous/Spaces' []
error: unknown No such file or directory: 'Fixture/.build/aarch64-unknown-linux-gnu/Intermediates.noindex/Spaces' []
info: Compile main.swift (aarch64)
[...]

PR #921, opened 2025-11-16
Correct linker file list escaping depending on the LINKER_DRIVER used

owenv:
LINKER_DRIVER=clang invocations need to use different formatting compared to LINKER_DRIVER=swiftc, and Libtool formatting should be independently configurable. Fixes issues reported in https://github.com/swiftlang/swift-build/issues/333


owenv:
@swift-ci test

owenv:
@swift-ci test

PR #923, opened 2025-11-17
[6.3] Do not pass index store options if indexing is disabled

ahoppen:
- **Explanation**: We were passing `-index-store-compress` to the compiler even though we don’t pass `-index-store-path` if `INDEX_STORE_ENABLE = YES` but `COMPILER_INDEX_STORE_ENABLE = NO`. This is because `CLANG_INDEX_STORE_ENABLE` still evaluated to `YES` and thus the `Condition` of `CLANG_INDEX_STORE_COMPRESS` also evaluated to `YES` but we didn’t pass `-index-store-path` because the condition of `CLANG_INDEX_STORE_ENABLE` evaluated to `NO`.
This was a misconception on my side when I added these build settings because I assumed that the `Condition` would influence the value of the build setting instead of just influencing `ComandLineArgs`.
- **Scope**: Builds that set one of the index-related options introduced in https://github.com/swiftlang/swift-build/pull/815
- **Issue**: rdar://164248532
- **Original PR**: https://github.com/swiftlang/swift-build/pull/903
- **Risk**: Low, the options affected by this have not been released yet
- **Testing**: Added unit tests
- **Reviewer**: @jakepetroules @neonichu

### DIFF
--- a/Sources/SWBApplePlatform/Specs/MetalCompiler.xcspec
+++ b/Sources/SWBApplePlatform/Specs/MetalCompiler.xcspec
@@ -239,7 +239,7 @@
                 Name = "METAL_INDEX_STORE_ONLY_PROJECT_FILES";
                 Type = Boolean;
                 DefaultValue = "$(INDEX_STORE_ONLY_PROJECT_FILES)";
-                Condition = "$(MTL_ENABLE_INDEX_STORE)";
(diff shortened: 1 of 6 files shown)


ahoppen:
@swift-ci Please test

owenv:
@ahoppen we're going to end up auto merging main into 6.3 for a bit so this should get included automatically once the first 6.3 toolchain gets released to unblock CI

ahoppen:
👍 Even better

PR #924, opened 2025-11-17
One time update of release/6.3 to desired branch point from main

owenv:
This is a one time update of release/6.3 to align with the desired branch point from main. From this point forward the expectation is that changes will land on release/6.3 first and then merge forward to main.


owenv:
@swift-ci test

rconnell9:
This may be unnecessary as the strategy may still be auto mergers from main -> 6.3

PR #925, opened 2025-11-17
Flip direction of the 6.3 merge workflow

owenv:


owenv:
@swift-ci test

.github/workflows/automerge.yml
```diff
@@ -15,8 +15,8 @@ jobs:
     name: Create PR to merge release into main branch
```
jakepetroules: This title, the title on the first line, and the code comment near the top, need to be updated to reflect the change in directionality.

jakepetroules (changes requested):

owenv:
@swift-ci test

PR #943, opened 2025-11-26
Verbosity fix

johnbute:
Fixes https://github.com/swiftlang/swift-package-manager/issues/9299

Solving --very-verbose flag causing rebuilds. Just filtered the "-v" flag when computing hashes and encoding to avoid rebuilding project.

### DIFF
--- a/Sources/SWBCore/SpecImplementations/CommandLineToolSpec.swift
+++ b/Sources/SWBCore/SpecImplementations/CommandLineToolSpec.swift
@@ -400,7 +400,7 @@ open class CommandLineToolSpec : PropertyDomainSpec, SpecType, TaskTypeDescripti
     }
 
     public func commandLineForSignature(for task: any ExecutableTask) -> [ByteString]? {
-        nil
+        return nil
     }
 
     static func parseCommandLineTemplate(_ parser: SpecParser, _ components: [String]) -> [CommandLineTemplateArg] {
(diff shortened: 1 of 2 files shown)


Sources/SWBCore/BuildParameters.swift
```diff
@@ -237,6 +237,15 @@ public struct BuildParameters: Hashable, SerializableCodable, Sendable {
         hasher.combine(toolchainOverride)
         return hasher.finalize()
     }
+
```
owenv: Instead of removing the settings override from the hash here, I think we should add the verbose flag to the output agnostic flags list for clang, swift, and ld, which is a narrower and safer change and ensures if -v. 

commandLineForSignature() in SwiftCompilerSpec is an example of how this works today for other flags

johnbute: Implemented your suggestion!

johnbute:
@swift-ci please test

Sources/SWBCore/SpecImplementations/CommandLineToolSpec.swift
```diff
@@ -399,8 +399,32 @@ open class CommandLineToolSpec : PropertyDomainSpec, SpecType, TaskTypeDescripti
         return false
     }
 
+    static let outputAgnosticCompilerArguments = Set<ByteString>([
```
owenv: I'm not sure we should apply this to every command line tool spec, because for some tools we may want to keep -v in the signature. Could we apply this to only the linker and compilers instead, even though that may require restating the list of flags each can ignore?

johnbute: I have done testing and it seems that we only need to add this to the linker.

jakepetroules (approved):

jakepetroules:
@swift-ci test

johnbute:
@swift-ci please test linux

johnbute:
@swift-ci please test macOS

PR #947, opened 2025-12-02
Pin GitHub actions workflows to a version

bkhouri:
Pinning has a few benefits:
- Workflows are more repeatable for a given a commit ID. We know exactly what steps were performed.
- Updating the workflows version goes though the PR validation process, so we know whether we want to apply a change to the entire repository.
- We avoid potential issues with breaking changes in the upstream workflows.


bkhouri:
@swift-ci test

PR #948, opened 2025-12-02
Only drop fobjc-arc from link lines when linking with swiftc

owenv:
Dropping it from clang link lines isn't strictly needed, and is a minor compatibility break.

rdar://165635777


owenv:
@swift-ci test

PR #953, opened 2025-12-04
Don't rely on SystemPrefix to set IOS_UNZIPPERED_TWIN_PREFIX_PATH

ian-twilightcoder:
Mac Catalyst shouldn't have a SystemPrefix set in its SDK variant, so don't rely on that to set IOS_UNZIPPERED_TWIN_PREFIX_PATH, hard code it instead. Set MACOS_UNZIPPERED_TWIN_PREFIX_PATH too for completeness.

rdar://165798693


Sources/SWBCore/Settings/Settings.swift
```diff
@@ -2585,9 +2585,10 @@ private class SettingsBuilder: ProjectMatchLookup {
             sdkTable.pushContentsOf(defaultSettingsTable)
         }
 
-        // Set IOS_UNZIPPERED_TWIN_PREFIX_PATH to the Mac Catalyst variant's prefix path, even for the macOS variant.
-        if let macCatalystVariant = sdk.variant(for: MacCatalystInfo.sdkVariantName) {
-            sdkTable.push(BuiltinMacros.IOS_UNZIPPERED_TWIN_PREFIX_PATH, literal: macCatalystVariant.systemPrefix)
+        // Set twin prefix paths in macOS for Mac Catalyst.
+        if sdk.canonicalNameComponents?.basename == "macosx" {
```
jakepetroules: I would do this in addPlatformSettings based on `if platform.name == "macosx"`

ian-twilightcoder:
@swift-ci test

Issue #956, opened 2025-12-04
Enable Swift types exposed in Objective-C APIs to be visible to the Swift consumer


timofeys1:
Hey folks. This is a bit more specific feature request for a set of issues:
* https://github.com/swiftlang/swift-package-manager/issues/9735
* https://github.com/swiftlang/swift-package-manager/issues/4362
* https://forums.swift.org/t/importing-swift-types-exposed-in-an-objective-c-api-back-into-swift/83492

When building SwiftPM targets in Xcode, it is challenging to enable the Swift consumer to see Swift types from transitive Swift modules when they are exposed through an Objective-C API, which makes the API itself uncallable.

My specific use case: we in Kotlin are building a SwiftPM interop for our native backend. When we emit a header file with types that we imported from SwiftPM, such APIs become uncallable in the final application.

I've built the reproduction for this issue [here](https://github.com/abdulowork/SwiftPMObjCTransitiveExport), and it's identical to the one described in the forum thread above.

Would it be possible to pass the `-Xcc -fmodule-map-file` to the consuming Swift compilations from transitive SwiftPM Swift targets to enable this use case?

timofeys1:
Just as a showcase, I tried implementing a fix for this in Swift Build (#957), and it seems to work, but there are likely other subtleties that need to be handled in a proper fix. I checked that it helps by running:
```
XCBBUILDSERVICE_PATH=/path/to/swift-build/.build/arm64-apple-macosx/debug/SWBBuildServiceBundle xcodebuild clean build -scheme SwiftPMObjCTransitiveExport -destination 'generic/platform=iOS Simulator' ARCHS=arm64
```
on the reproduction above

PR #957, opened 2025-12-04
Showcase a fix for exposing Swift types in an Objecive-C API

timofeys1:
This is a PoC fix for the issue described in #956

### DIFF
--- a/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
@@ -1598,6 +1598,19 @@ public final class SwiftCompilerSpec : CompilerSpec, SpecIdentifierType, SwiftDi
                 }
             }
 
+            if let target = cbc.producer.configuredTarget {
+                let depScopes = cbc.producer.targetSwiftDependencyScopes(for: target, arch: cbc.scope.evaluate(BuiltinMacros.CURRENT_ARCH), variant:  cbc.scope.evaluate(BuiltinMacros.CURRENT_VARIANT))
+                for scope in depScopes {
+                    let moduleName = scope.evaluate(BuiltinMacros.PRODUCT_MODULE_NAME)
+                    // From CoreBuildSystem.xcspec: $(OBJROOT)/GeneratedModuleMaps$(EFFECTIVE_PLATFORM_NAME)
+                    let objroot = scope.evaluate(BuiltinMacros.OBJROOT).str
+                    let moduleMapDir = "\(objroot)/GeneratedModuleMaps\(scope.evaluate(BuiltinMacros.EFFECTIVE_PLATFORM_NAME))"
+                    // Doesn't work for SwiftPM targets
+                    // let moduleMapDir = scope.evaluate(BuiltinMacros.GENERATED_MODULEMAP_DIR)
+                    args.append(contentsOf: ["-Xcc", "-fmodule-map-file=\(moduleMapDir)/\(moduleName).modulemap"])
+                }
+            }
+
             // Add -F for the effective framework search paths.
             let frameworkSearchPaths = GCCCompatibleCompilerSpecSupport.frameworkSearchPathArguments(cbc.producer, cbc.scope, asSeparateArguments: true)
             args += frameworkSearchPaths.searchPathArguments(for: self, scope: cbc.scope)


Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
```diff
@@ -1598,6 +1598,19 @@ public final class SwiftCompilerSpec : CompilerSpec, SpecIdentifierType, SwiftDi
                 }
             }
 
+            if let target = cbc.producer.configuredTarget {
```
owenv, line 1601: Thanks for the pr! Not every target will have a generated modulemap, and not every transitive dependency will build a module compatible with the current compilation, so I don't think we should attempt to inject module map flags here. 

Generally speaking, the intent is that the PIF generated by swift-package-manager is supposed to include module map paths in the "imparted build properties" as needed for SwiftPM targets generating module maps. Then, swift-build is responsible for ensuring those imparted build properties are imposed on transitive dependents which build for the same platform. Based on your bug report it's possible either SwiftPM is failing to add the right path to the PIF, or the imparted properties exist but aren't propagated to dependents correctly (I haven't taken a detailed look yet). The former we'd want to fix in SwiftPM, the latter we'd likely want to fix in Swift Build at dependency resolution time instead of when we're constructing the Swift compilation task

timofeys1: I looked at the output of `swift package dump-pif` and at similar information emitted with `defaults write com.apple.dt.XCBuild EnableBuildDebugging -bool YES`. I see that `ObjcTarget` has the following in `impartedBuildProperties`:
```
"guid": "PACKAGE-TARGET:ObjcTarget",
...
"OTHER_SWIFT_FLAGS": [
  "-Xcc",
  "-fmodule-map-file=$(GENERATED_MODULEMAP_DIR)/ObjcTarget.modulemap",
```
And the SwiftTarget doesn't have the imparted `OTHER_SWIFT_FLAGS`, so the transitive Swift consumer doesn't see the generated `SwiftTarget.modulemap`.

That still leaves me with a couple of questions: do I understand correctly that, for general Swift -> Swift compilation dependencies, SwiftPM relies on Swift Build to produce all outputs in `$(BUILT_PRODUCTS_DIR)` and have them picked up by a shared `swiftc -I $(BUILT_PRODUCTS_DIR)` search path?

If my understanding is correct, I see another possible fix for this issue being Swift Build creating `$(BUILT_PRODUCTS_DIR)/SwiftTarget` directory with the `module.modulemap` and the `SwiftTarget-Swift.h`. Then, from my understanding, both `clang -fmodules` and `swiftc` can pick the necessary Clang module, and wouldn't need to rely on the imparted build settings at all.

Would such a fix be acceptable, and would it need to be implemented in Swift Build?

timofeys1: I think the benefit of such a solution might be that if there is no `Swift Producer -> ObjC Consumer` dependency, then the build system doesn't need to ask the compiler to generate the `-Swift.h` file, and it doesn't need to create the modulemap, but if such an edge does occur, then in the case of `Swift Producer -> ObjC -> Swift Consumer` the `Swift Consumer` will still be able to see everything due to emitted module and the header in the shared search path.

Or is this something SwiftPM must decide on ahead of time based on its knowledge about the dependency graph?

timofeys1: I browsed some more through the PIF files, and I see the `MODULEMAP_FILE_CONTENTS`, `MODULEMAP_PATH`, `SWIFT_OBJC_INTERFACE_HEADER_DIR` and `SWIFT_OBJC_INTERFACE_HEADER_NAME` build settings there, which I assume get encoded into PIF from [here](https://github.com/swiftlang/swift-package-manager/blob/6c2755da6fb4de1fd71b6c5fda5a6ab422d58e96/Sources/SwiftBuildSupport/PackagePIFProjectBuilder%2BModules.swift#L358-L359). So perhaps for the fix with the shared search path, this would still require a change in SwiftPM and not in Swift Build

owenv: In general, I think the right thing to do here would be to have Swift targets impart Swift clang importer flags to pick up the module map covering the generated header. However, exposing an underlying clang module for Swift modules like this for the first time may have some compatibility implications, so it might make sense to take that through swift-evolution

timofeys1: Ok, thank you for the explanation. I will then recreate this issue in the SwiftPM repository and consider what next steps are necessary to organize this discussion

PR #960, opened 2025-12-05
Introduce BUILD_ONLY_KNOWN_LOCALIZATIONS build setting

Tantalum73:
This PR introduces a build setting `BUILD_ONLY_KNOWN_LOCALIZATIONS`. Only languages explicitly added to the project's "Localizations" will be built when set.
The new build setting allows developers to have work-in-progress translations in their project, which don't get built.

In the old world with only `.strings` files, it was possible to have partial translations in files that were on disk, but not referenced in the Xcode project. When localization was complete, they were simply added as a reference, and then they started contributing to the build.  
Since String Catalogs contain all localizations in a single file, it was not possible to _not build_ content for a given language, making this workflow more complicated to achieve.

This build setting fixes that. When set, only languages in the list of "Localizations" in the project settings are built – including String Catalog content.

Now, for the time that a new language is brought up that isn't fully translated yet, developers enable `BUILD_ONLY_KNOWN_LOCALIZATIONS`, and remove that language from the list of Localizations. The work on this language can continue, and when it's ready to ship they add it  back to the list, or disable `BUILD_ONLY_KNOWN_LOCALIZATIONS`.

### DIFF
--- a/Sources/SWBProtocol/ProjectModel/Project.swift
+++ b/Sources/SWBProtocol/ProjectModel/Project.swift
@@ -63,7 +63,7 @@ extension Project: Serializable {
     }
 
     public init(from deserializer: any Deserializer) throws {
-        try deserializer.beginAggregate(12)
+        let count = try deserializer.beginAggregate(11...12)
         self.guid = try deserializer.deserialize()
         self.isPackage = try deserializer.deserialize()
         self.xcodeprojPath = try deserializer.deserialize()
@@ -73,7 +73,11 @@ extension Project: Serializable {
         self.buildConfigurations = try deserializer.deserialize()
         self.defaultConfigurationName = try deserializer.deserialize()
         self.developmentRegion = try deserializer.deserialize()
-        self.knownLocalizations = try deserializer.deserialize()
+        if count >= 12 {
+            self.knownLocalizations = try deserializer.deserialize()
+        } else {
+            self.knownLocalizations = nil
+        }
         self.classPrefix = try deserializer.deserialize()
         self.appPreferencesBuildSettings = try deserializer.deserialize()
     }


Sources/SWBApplePlatform/XCStringsCompiler.swift
```diff
@@ -189,7 +189,35 @@ public final class XCStringsCompilerSpec: GenericCompilerSpec, SpecIdentifierTyp
...
+                guard !isRunningInstallloc else {
+                    // No need to intercept anything for installloc, its
+                    // language specification should always take precedence.
+                    return nil
+                }
+                if cbc.scope.evaluate(BuiltinMacros.BUILD_ONLY_KNOWN_LOCALIZATIONS), var knownLocalizations = cbc.producer.project?.knownLocalizations {
+
+                    knownLocalizations.removeAll(where: { $0 == "Base" })
+                    if !knownLocalizations.isEmpty {
+                        delegate.note("XCStrings will compile languages for known regions: \(knownLocalizations.joined(separator: ", "))")
```
matthewseaman, line 208: Just heads up that messages like this are going to show up for every xcstrings file, although in this particular case the specific file is not provided. Not saying it needs to be provided, but may appear as duplicate messages. Also for IB Compiler, etc.

Sources/SWBCore/FileToBuild.swift
```diff
@@ -179,4 +179,61 @@ public extension RegionVariable {
...
+            return true
+        }
+
+        if knownLocalizations.contains(regionVariantName) {
+            // This is a known locale, allow it to build:
+            return true
+        }
+
+        // This region is not supported, so it shouldn't build.
+        delegate?.note("Skipping .lproj directory '\(regionVariantName).lproj' because '\(regionVariantName)' is not in project's known localizations (BUILD_ONLY_KNOWN_LOCALIZATIONS is enabled)")
```
matthewseaman: Might want to look into specifying file paths for diagnostics like this. I don't mean in the message, but rather as part of the diag instance.

Sources/SWBCore/FileToBuild.swift
```diff
@@ -179,4 +179,61 @@ public extension RegionVariable {
...
+            // We can't read the supported localizations, allow it to build:
+            return true
+        }
+
+        guard let regionVariantName else {
+            // Unable to get a region name, allow it to built:
+            return true
+        }
+
+        if regionVariantName == "mul" {
```
matthewseaman, line 230: What if it is `Base`?

matthewseaman:
@swift-ci test

Sources/SWBCore/FileToBuild.swift
Tantalum73: Sure! By doing this, a double click on the log message will automatically open the source file, neat!

https://github.com/swiftlang/swift-build/pull/960/commits/9764b31be1d21109465f52ebbacb319092a7999d

Sources/SWBApplePlatform/XCStringsCompiler.swift
Tantalum73: With the changes in https://github.com/swiftlang/swift-build/pull/960/commits/9764b31be1d21109465f52ebbacb319092a7999d where a source location added to the message, they might look like duplicates at first, but double clicking the entry will open the corresponding source file.
Thanks for pointing that out!

Sources/SWBCore/FileToBuild.swift
Tantalum73: `Base` will be included in the list of known languages if using Base internationalization. I think no special rules need to apply to Base here.

matthewseaman: Oh I'm surprised that Base is included in knownLocalizations, but that works I think

matthewseaman:
@swift-ci test

Sources/SWBProtocol/ProjectModel/Project.swift
```diff
@@ -22,10 +22,11 @@ public struct Project: Sendable {
     public let buildConfigurations: [BuildConfiguration]
     public let defaultConfigurationName: String
     public let developmentRegion: String?
+    public let knownLocalizations: [String]?
     public let classPrefix: String
     public let appPreferencesBuildSettings: [BuildConfiguration.MacroBindingSource]
 
-    public init(guid: String, isPackage: Bool, xcodeprojPath: Path, sourceRoot: Path, targetSignatures: [String], groupTree: FileGroup, buildConfigurations: [BuildConfiguration], defaultConfigurationName: String, developmentRegion: String?, classPrefix: String, appPreferencesBuildSettings: [BuildConfiguration.MacroBindingSource] = []) {
```
owenv, line 28: I think this is ABI and will need to add a new initializer instead of extending the original

Sources/SWBProtocol/ProjectModel/Project.swift
```diff
@@ -54,13 +56,14 @@ extension Project: Serializable {
...
             serializer.serialize(defaultConfigurationName)
             serializer.serialize(developmentRegion)
+            serializer.serialize(knownLocalizations)
             serializer.serialize(classPrefix)
             serializer.serialize(appPreferencesBuildSettings)
         }
     }
 
     public init(from deserializer: any Deserializer) throws {
-        try deserializer.beginAggregate(11)
```
owenv, line 63: For compatibility this likely needs to remain able to decode messages without the new field

Sources/SWBProtocol/ProjectModel/Project.swift
jakepetroules: The SWBProtocol library is not part of the public API+ABI, so this should be fine.

Sources/SWBProtocol/ProjectModel/Project.swift
jakepetroules: Yep. Use `try deserializer.beginAggregate(11...12)` and then use the returned count value to decide whether to deserialize knownLocalizations

Tantalum73: Oh that is a very good point! And thanks for the actionable comment!

https://github.com/swiftlang/swift-build/pull/960/commits/12fe102a0f5a50f2df2d5823d4a92b81adaacd08

matthewseaman:
@swift-ci test

Tantalum73:
@swift-ci  test

PR #962, opened 2025-12-08
Compare file types instead of extensions when identifying artifact bundles

owenv:
Unpacked artifact bundles won't always have the artifactbundle extensions, so we need to look at the actual file type of the reference, which SwiftPM is responsible for encoding


owenv:
@swift-ci test

PR #965, opened 2025-12-09
Normalize discovered inputs to copy tasks via source symlink

owenv:
Otherwise, llbuild may form directory structure nodes which trigger asserts during rule scanning.

Closes https://github.com/swiftlang/swift-package-manager/issues/9464


owenv:
@swift-ci test

owenv:
@swift-ci test

neonichu (approved):

PR #966, opened 2025-12-09
Swift SDK for Wasm using the run destination

cmcgee1024:
The client can provide a Swift SDK manifest path along with a triple in the run destination to synthesize a Swift Build SDK. This serves as a fallback for cases where Swift Build cannot match an SDK for it.

Note that in this first implementation the platform is hard coded to WebAssembly, and so only SDK's that align with Wasm, such as static linking of compiler runtime pieces, and some Unix semantics will work.

### DIFF
--- a/Tests/SWBTaskConstructionTests/TaskConstructionTests.swift
+++ b/Tests/SWBTaskConstructionTests/TaskConstructionTests.swift
@@ -3513,6 +3513,156 @@ fileprivate struct TaskConstructionTests: CoreBasedTests {
         }
     }
 
+    @Test
+    func swiftSDKRunDestination() async throws {
+        let clangCompilerPath = try await self.clangCompilerPath
+        let swiftCompilerPath = try await self.swiftCompilerPath
+        let swiftVersion = try await self.swiftVersion
+        let targetName = "CoreFoo"
+        let testProject = try await TestProject(
+            "aProject",
+            groupTree: TestGroup(
+                "SomeFiles", path: "Sources",
+                children: [
+                    TestFile("SourceFile.m"),
+                    TestFile("SwiftFile.swift"),
+                ]),
+            targets: [
+                TestStandardTarget(
+                    targetName,
+                    type: .framework,
+                    buildConfigurations: [
+                        TestBuildConfiguration("Debug",
+                                               buildSettings: [
+                                                "GENERATE_INFOPLIST_FILE": "YES",
+                                                "PRODUCT_NAME": "$(TARGET_NAME)",
+                                                "SDKROOT": "macosx",
+                                                "CLANG_ENABLE_MODULES": "YES",
+                                                "SWIFT_EXEC": swiftCompilerPath.str,
+                                                "SWIFT_VERSION": swiftVersion,
+                                                "TAPI_EXEC": tapiToolPath.str,
+                                                "CC": clangCompilerPath.str,
+                                                "CLANG_EXPLICIT_MODULES_LIBCLANG_PATH": libClangPath.str,
+                                                "CLANG_WARN_BLOCK_CAPTURE_AUTORELEASING": "",
+                                                "CLANG_WARN_COMMA": "",
+                                                "CLANG_WARN_FLOAT_CONVERSION": "",
+                                                "CLANG_WARN_NON_LITERAL_NULL_CONVERSION": "",
+                                                "CLANG_WARN_OBJC_LITERAL_CONVERSION": "",
+                                                "CLANG_WARN_STRICT_PROTOTYPES": "",
+                                                "CLANG_WARN_IMPLICIT_FALLTHROUGH": "",
+                                                "CLANG_USE_RESPONSE_FILE": "NO",
+                                               ]),
+                    ],
+                    buildPhases: [
+                        TestSourcesBuildPhase([
+                            TestBuildFile("SourceFile.m"),
+                            TestBuildFile("SwiftFile.swift"),
+                        ]),
+                    ]),
+                TestStandardTarget(
+                    "FrameworkWithModule",
+                    type: .framework,
+                    buildConfigurations: [
+                        TestBuildConfiguration("Debug",
+                                               buildSettings: [
+                                                "GENERATE_INFOPLIST_FILE": "YES",
+                                                "PRODUCT_NAME": "$(TARGET_NAME)",
+                                                "SDKROOT": "macosx",
+                                                "CLANG_ENABLE_MODULES": "YES",
+                                                "DEFINES_MODULE": "YES",
+                                                "SWIFT_EXEC": swiftCompilerPath.str,
+                                                "SWIFT_VERSION": swiftVersion,
+                                                "TAPI_EXEC": tapiToolPath.str,
+                                                "CC": clangCompilerPath.str,
+                                                "CLANG_EXPLICIT_MODULES_LIBCLANG_PATH": libClangPath.str,
+                                                "CLANG_WARN_BLOCK_CAPTURE_AUTORELEASING": "",
+                                                "CLANG_WARN_COMMA": "",
+                                                "CLANG_WARN_FLOAT_CONVERSION": "",
+                                                "CLANG_WARN_NON_LITERAL_NULL_CONVERSION": "",
+                                                "CLANG_WARN_OBJC_LITERAL_CONVERSION": "",
+                                                "CLANG_WARN_STRICT_PROTOTYPES": "",
+                                                "CLANG_WARN_IMPLICIT_FALLTHROUGH": "",
+                                               ]),
+                    ],
+                    buildPhases: [
+                        TestSourcesBuildPhase([
+                            TestBuildFile("SourceFile.m"),
+                            TestBuildFile("SwiftFile.swift"),
+                        ]),
+                    ])
+            ])
+        let core = try await getCore()
+        let tester = try TaskConstructionTester(core, testProject)
+        let SRCROOT = tester.workspace.projects[0].sourceRoot.str
+
+        // Swift SDK contents
+        let sdkManifestContents = """
+        {
+            "schemaVersion" : "4.0",
+            "targetTriples" : {
+                "wasm32-unknown-wasip1" : {
+                    "sdkRootPath" : "WASI.sdk",
+                    "swiftResourcesPath" : "swift.xctoolchain/usr/lib/swift_static",
+                    "swiftStaticResourcesPath" : "swift.xctoolchain/usr/lib/swift_static",
+                    "toolsetPaths" : [
+                        "toolset.json"
+                    ]
+                }
+            }
+        }
+        """
+        let sdkManifestDir = try localFS.createTemporaryDirectory(parent: localFS.realpath(Path.root.join("tmp")))
+        let sdkManifestPath = sdkManifestDir.join("swift-sdk.json")
+        try await localFS.writeFileContents(sdkManifestDir.join("swift-sdk.json"), waitForNewTimestamp: false, body: { $0.write(sdkManifestContents) })
+        try await localFS.writeFileContents(sdkManifestDir.join("toolset.json"), waitForNewTimestamp: false, body: { stream in
+            stream.write("""
+            {
+                "rootPath" : "swift.xctoolchain/usr/bin",
+                "schemaVersion" : "1.0",
+                "swiftCompiler" : {
+                    "extraCLIOptions" : [
+                        "-static-stdlib"
+                    ]
+                }
+            }
+            """)
+        })
+
+        let sysroot = sdkManifestDir.join("WASI.sdk")
+        let sdkroot = sdkManifestDir.join("WASI.sdk")
+
+        // Check a build with both targets; this should use the VFS since one defines a module.
+        let derivedData = Path("\(SRCROOT)/DerivedData")
+        let arena = ArenaInfo.buildArena(derivedDataRoot: derivedData, enableIndexDataStore: true)
+        let parameters = BuildParameters(configuration: "Debug", activeRunDestination: RunDestinationInfo(buildTarget: .swiftSDK(sdkManifestPath: sdkManifestPath.str, triple: "wasm32-unknown-wasip1"), targetArchitecture: "wasm32", supportedArchitectures: ["wasm32"], disableOnlyActiveArch: false), arena: arena)
+        let request1 = BuildRequest(parameters: parameters, buildTargets: tester.workspace.projects[0].targets.map({ BuildRequest.BuildTargetInfo(parameters: parameters, target: $0) }), continueBuildingAfterErrors: true, useParallelTargets: true, useImplicitDependencies: true, useDryRun: false)
+        await tester.checkBuild(parameters, runDestination: nil, buildRequest: request1) { results in
+            results.checkTarget(targetName) { target in
+                // There should be one CompileC task.
+                results.checkTask(.matchTarget(target), .matchRuleType("CompileC")) { task in
+                    task.checkCommandLineContains([
+                        [clangCompilerPath.str],
+                        ["-target", "wasm32-unknown-wasip1"],
+                        ["--sysroot", sysroot.str],
+                    ].reduce([], +))
+                }
+
+                // There should be one SwiftDriver task.
+                results.checkTask(.matchTarget(target), .matchRuleType("SwiftDriver Compilation")) { task in
+                    task.checkCommandLineContains([
+                        ["-resource-dir", sdkManifestDir.join("swift.xctoolchain").join("usr").join("lib").join("swift_static").str],
+                        ["-static-stdlib"],
+                        ["-sdk", sdkroot.str],
+                        ["-target", "wasm32-unknown-wasip1"],
+                    ].reduce([], +))
+                }
+            }
+
+            // Check there are no diagnostics.
+            results.checkNoDiagnostics()
+        }
+    }
+
     @Test(.requireSDKs(.macOS))
     func moduleCompilerOptions() async throws {
         let clangCompilerPath = try await self.clangCompilerPath


Sources/SWBCore/SDKRegistry.swift
```diff
@@ -1111,8 +1112,103 @@ public final class SDKRegistry: SDKRegistryLookup, CustomStringConvertible, Send
...
+                        ].merging(defaultProperties, uniquingKeysWith: { _, new in new })),
+                        "CustomProperties": .plDict([
+                            "LIBRARY_SEARCH_PATHS": .plArray(librarySearchPaths),
+                            "HEADER_SEARCH_PATHS": .plArray(headerSearchPaths),
+                            "OTHER_SWIFT_FLAGS": .plArray(["$(inherited)"] + extraSwiftCompilerSettings.map( {.plString($0)} )),
+                            "SWIFTC_RESOURCE_DIR": .plString(swiftResourceDir.str), // Resource dir for linking Swift
+                            "SWIFT_RESOURCE_DIR": .plString(swiftResourceDir.str), // Resource dir for compiling Swift
+                            "CLANG_RESOURCE_DIR": .plString(clangResourceDir.str), // Resource dir for linking C/C++/Obj-C
+                            "SDKROOT": .plString(sysroot.str),
+                            "OTHER_LDFLAGS": .plArray(["$(OTHER_SWIFT_FLAGS)"]), // The extra swift compiler settings in JSON are intended to go to the linker driver too
```
owenv: If we need to include swift flags here, we should use extraSwiftCompilerSettings instead of `$(OTHER_SWIFT_FLAGS)`, because `$(OTHER_SWIFT_FLAGS)` may include other project/command line flags which are only intended for the compiler

Sources/SWBProtocol/MessageSupport.swift
```diff
@@ -325,14 +325,18 @@ public struct RunDestinationInfo: SerializableCodable, Hashable, Sendable {
     public var supportedArchitectures: OrderedSet<String>
     public var disableOnlyActiveArch: Bool
     public var hostTargetedPlatform: String?
+    public var sdkManiftestPath: String?
```
owenv:
```suggestion
    public var sdkManifestPath: String?
```

Sources/SwiftBuild/SWBBuildParameters.swift
```diff
@@ -43,6 +43,8 @@ public struct SWBRunDestinationInfo: Codable, Sendable {
     public var supportedArchitectures: [String]
     public var disableOnlyActiveArch: Bool
     public var hostTargetedPlatform: String?
+    public var sdkManifestPath: String?
```
jakepetroules: We need to determine what happens when multiple of these are specified. What happens if you specify sdkManifestPath _and_ sdk _and_ sdkVariant? Can you provide some examples of what the values would be in such cases?

Similarly, what happens if targetArchitecture and triple are both specified? What are some examples of inputs?

Sources/SWBCore/Settings/Settings.swift
```diff
@@ -3504,7 +3504,7 @@ private class SettingsBuilder: ProjectMatchLookup {
         // If the target supports specialization and it already has an SDK, then we need to use that instead of attempting to override the SDK with the run destination information. This is very important in scenarios where the destination is Mac Catalyst, but the target is building for iphoneos. The target will be re-configured for macosx/iosmac.
         do {
             let scope = createScope(sdkToUse: nil)
-            let sdk = try sdkRegistry.lookup(scope.evaluate(BuiltinMacros.SDKROOT).str, activeRunDestination: nil)
+            let sdk = try sdkRegistry.lookup(scope.evaluate(BuiltinMacros.SDKROOT).str, activeRunDestination: parameters.activeRunDestination)
```
jakepetroules: Why was this change necessary?

Sources/SWBCore/SDKRegistry.swift
```diff
@@ -1036,8 +1112,103 @@ public final class SDKRegistry: SDKRegistryLookup, CustomStringConvertible, Send
...
         }
 
-        // Xcode has special logic so that if there's no match here, and we'e *not* looking for a suffixed SDK, but we have an suffixed SDK which would otherwise match, then we use that. c.f. <rdar://problem/11414721> But we haven't needed that logic yet in Swift Build, so maybe we never will.
+        // Let's check the active run destination to see if there's an SDK path that we should be using
+        if matchedSDK == nil,
+            let sdkManifestPath = activeRunDestination?.sdkManiftestPath,
+            let triple = activeRunDestination?.triple,
+            case let llvmTriple = try LLVMTriple(triple) {
+
+            // TODO choose the platform based on the triple, and fallback to a default one in the default case
```
jakepetroules: Why are you basing the platform on the triple?

The run destination already has a `platform` field that can be used to look up the platform.

Sources/SWBCore/SDKRegistry.swift
```diff
@@ -1036,8 +1112,103 @@ public final class SDKRegistry: SDKRegistryLookup, CustomStringConvertible, Send
...
+                            "HEADER_SEARCH_PATHS": .plArray(headerSearchPaths),
+                            "OTHER_SWIFT_FLAGS": .plArray(["$(inherited)"] + extraSwiftCompilerSettings.map( {.plString($0)} )),
+                            "SWIFTC_RESOURCE_DIR": .plString(swiftResourceDir.str), // Resource dir for linking Swift
+                            "SWIFT_RESOURCE_DIR": .plString(swiftResourceDir.str), // Resource dir for compiling Swift
+                            "CLANG_RESOURCE_DIR": .plString(clangResourceDir.str), // Resource dir for linking C/C++/Obj-C
+                            "SDKROOT": .plString(sysroot.str),
+                            "OTHER_LDFLAGS": .plArray(["$(OTHER_SWIFT_FLAGS)"]), // The extra swift compiler settings in JSON are intended to go to the linker driver too
+                        ]),
+                        "SupportedTargets": .plDict([
+                            "webassembly": .plDict([
```
jakepetroules: You've already looked up a platform, so this should be `platform.name`.

Sources/SWBCore/SDKRegistry.swift
```diff
@@ -1036,8 +1112,103 @@ public final class SDKRegistry: SDKRegistryLookup, CustomStringConvertible, Send
...
+
+            // TODO choose the platform based on the triple, and fallback to a default one in the default case
+            guard let platform = delegate.platformRegistry?.lookup(name: "webassembly") else {
+                return nil
+            }
+
+            let host = hostOperatingSystem
+
+            // Don't allow re-registering the same SDK
+            if let existing = sdksByCanonicalName[sdkManifestPath] {
```
jakepetroules: There are multiple maps based on the type of key is being used to look up the SDK -- this one is for canonical name. This one should be placed into the `sdksByPath` map instead, and furthermore I think all of this logic you're adding should go into the lookup() overload (there's 3) that deals in terms of paths rather than this one, which deals in terms of canonical names.

Sources/SWBCore/Settings/Settings.swift
cmcgee1024: @owenv pointed this one out where the active run destination wasn't being consistently passed down to the lookup.

Sources/SWBCore/SDKRegistry.swift
cmcgee1024: SwiftPM doesn't operate on platforms. It only has the triple, so it seems that the heuristics to pick a platform based on the triple falls to Swift Build when the triple is specified without a platform.

Sources/SwiftBuild/SWBBuildParameters.swift
cmcgee1024: I'm not very familiar with this API. I think that it would be preferable to make some of these options be mutually exclusive and others grouped together. In particular, I think that the SDK manifest path and triple should be mutually required and that should be exclusive to the SDK (variant) and targetArchitecture options.

Would it be a breaking change to introduce an enum with product and sums?

Sources/SWBCore/SDKRegistry.swift
cmcgee1024: Thanks, I've moved things over to the other map since these are paths. Although the paths are to Swift SDK manifest JSON files, and not the root directory of Apple platform SDK's these should never interfere with each other.

Sources/SWBCore/Settings/Settings.swift
cmcgee1024: I've restored the existing arguments because it appears that this is a special case and the lookup method has been restored to its existing behaviour now that there's a new synthesizedSDK() method.

Sources/SWBCore/SDKRegistry.swift
jakepetroules: Today SwiftPM has both the triple and the platform as it takes on the responsibility of mapping triples to platforms when it's constructing the Swift Build build request (see here: https://github.com/swiftlang/swift-package-manager/blob/63047d9b669fa231ed39073e252421da97d3eea8/Sources/SwiftBuildSupport/SwiftBuildSystem.swift#L839), because it's a required property of the API. But there is no reason it has to be done that way, that was just to test things out for bringup.

So if we want the SwiftPM layer to not be aware of platforms at all, then we need to change the shape of the Swift Build API to be something different, for example, perhaps something like:

```swift
enum BuildTarget {
    case appleSDK(platform: String, sdk: String, sdkVariant: String?, targetArchitecture: String, supportedArchitectures: [String])
    case swiftSDK(path: String, triple: String)
}
```

In terms of mapping triples to platforms, maybe that can be an additional method on `PlatformInfoExtension`, and the individual platform plugins can "claim" their triples, with everything else falling back to the "none" (Bare Metal) platform.

Sources/SwiftBuild/SWBBuildParameters.swift
cmcgee1024: I've refactored the API to introduce these two cases in the run destination as a build target.

Sources/SWBCore/SDKRegistry.swift
cmcgee1024: I've implemented an API along these lines to better distinguish between the two cases while preserving backward compatibility.

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci please test

Sources/SwiftBuild/SWBBuildParameters.swift
```diff
@@ -36,13 +36,47 @@ public struct SWBBuildParameters: Codable, Sendable {
 
 /// Refer to `SWBProtocol.RunDestinationInfo`
 public struct SWBRunDestinationInfo: Codable, Sendable {
+    public enum SWBBuildTarget: Codable, Sendable {
```
jakepetroules: The general idea here makes sense. However, so far we've generally avoided using enums in the public API, in part because it's a nightmare to evolve, for things which might gain new parameters over time.

Instead, I would follow the pattern I established with `SWBBuildCommand`, where instead of an enum, you make `SWBBuildTarget` a struct, and make the enum cases into static functions. This allows you to provide additional overloads in order to extend the API with more arguments over time in a more compatible fashsion.

Then inside `SWBBuildCommand` you have a _private_ enum which models the actual data, but is not exposed through the API. That one can then change however it likes without impacting ABI.

Let's also mark the loose properties as deprecated, so that we can start to transition clients to use the new ones. Instead of making the `SWBBuildTarget` operate in terms of the loose properties, make the loose properties operate in terms of the `SWBBuildTarget`. That will prevent you from having unsilenceable warnings from a non-deprecated API relying on deprecated API.

Sources/SwiftBuild/SWBBuildParameters.swift
```diff
@@ -36,13 +36,47 @@ public struct SWBBuildParameters: Codable, Sendable {
...
+                self.sdk = sdk
+                self.sdkVariant = sdkVariant
+                self.targetArchitecture = targetArchitecture
+                self.supportedArchitectures = supportedArchitectures
+            case .swiftSDK:
+                self._internalBuildTarget = newValue
+                self.platform = ""
+                self.sdk = ""
+                self.sdkVariant = nil
+                self.targetArchitecture = ""
```
jakepetroules: I think targetArchitecture and supportedArchitectures should stay on the `SWBRunDestinationInfo` level, since they aren't logically part of platform establishment.

(This would also make more sense for future enhancements, like supporting multiple target triples in a single build request - I'd like to do that for Apple platforms, and it would make sense for Android and probably Windows, too.)

Sources/SWBUniversalPlatform/Specs/Ld.xcspec
jakepetroules, line 163: I think this one needs to be `$(SYSROOT:default=$(SDKROOT))`

Sources/SWBWebAssemblyPlatform/Specs/WasmLd.xcspec
```diff
@@ -129,6 +137,14 @@
                 Type = Path;
                 DefaultValue = "";
             },
+            {
+                Name = SWIFTC_SDKROOT_LINKER_INPUT;
+                Type = Path;
+                DefaultValue = "$(SYSROOT:default=$(SDKROOT))";
```
jakepetroules: This doesn't seem right -- based on discussions around how the various platform-targeting flags would work, the consensus was:

- When targeting platforms where Swift SDK and sysroot are the same (Apple platforms), use `-sdk` only
- When targeting Windows, use `-sdk` only since the sysroot concept doesn't exist
- For all others, use both `-sdk` and `-sysroot`

Wasm falls into bullet 3, so I think the `sdk` flag needs to be based on `$(SDKROOT)`, while the `$(SYSROOT)` var would be used to pass `-sysroot`. That means you should be able to remove this override and use the one from the base Ld.xcspec.

Sources/SWBProtocol/MessageSupport.swift
```diff
@@ -318,13 +318,16 @@ public struct BuildParametersMessagePayload: SerializableCodable, Equatable, Sen
 }
 
 public struct RunDestinationInfo: SerializableCodable, Hashable, Sendable {
```
jakepetroules, line 320: This is not part of public API and there is no need to maintain backwards compatibility, so this should eliminate the dedicated properties and go straight to having a BuildTarget enum with the two cases (swiftSDK for swiftSdkPath + triple, and appleSDK for platform + sdk + sdkVariant).

Sources/SWBCore/SDKRegistry.swift
```diff
@@ -1078,6 +1156,96 @@ public final class SDKRegistry: SDKRegistryLookup, CustomStringConvertible, Send
...
+
+        if let swiftSDK = try SwiftSDK(identifier: sdkManifestPath, version: "1.0.0", path: Path(sdkManifestPath), fs: localFS) {
+            let defaultProperties: [String: PropertyListItem] = [
+                "SDK_STAT_CACHE_ENABLE": "NO",
+
+                "GENERATE_TEXT_BASED_STUBS": "NO",
+                "GENERATE_INTERMEDIATE_TEXT_BASED_STUBS": "NO",
+
+                // TODO check how this impacts operation on Windows
+                "CHOWN": "/usr/bin/chown",
```
jakepetroules, line 1177: No such concept of CHOWN, so shouldn't matter. `addSpecificChangePermissionTasks` bails out early for Windows targets.

Sources/SWBCore/Settings/Settings.swift
```diff
@@ -3639,9 +3675,23 @@ private class SettingsBuilder: ProjectMatchLookup {
...
-        guard let destinationPlatform: Platform = self.core.platformRegistry.lookup(name: runDestination.platform) else {
-            self.errors.append("unable to resolve run destination platform: '\(runDestination.platform)'")
-            return
+
+        let destinationPlatform: Platform
+
+        switch runDestination.buildTarget {
+        case .swiftSDK:
+            // TODO use the triple to decide the platform, or use a fallback
+            guard let platform: Platform = self.core.platformRegistry.lookup(name: "webassembly") else {
```
jakepetroules: Let's fix this hardcoded webassembly reference before landing the initial change. Paraphrasing what I mentioned in another comment, a good way to do this would be to additional method on `PlatformInfoExtension` like `func platformName(triple: LLVMTriple) -> String?`.

Then, in each of the implementations:
- AndroidPlatformExtension: return "android" if triple.os == "linux" and triple.environment has prefix "android"
- GenericUnixPlatformInfoExtension: return triple.os if (triple.os is "linux" and (triple.environment has prefix "gnu" or == "musl")), "freebsd", or "openbsd"
- QNXPlatformExtension: return "qnx" if triple.os == "nto"
- WebAssemblyPlatformExtension: return "webassembly" if triple.os has prefix "wasi"
- WindowsPlatformExtension: return "windows" if triple.os == "windows

Here in Settings.swift, you enumerate the extension point conformances, add them to a Set, and if you have exactly one match, use it, else emit an error.

If you get no matches at all, fall back to the "none" platform.

--

Also make sure to factor this logic into a dedicated function; I see the hardcoded webassembly lookup is currently in 3 different places.

Sources/SWBWebAssemblyPlatform/Specs/WasmLd.xcspec
cmcgee1024: The UnixLd.xcspec is getting in the way between Ld.xcspec and this WasmLd.xcspec. If I remove this then it will prevent it from adding the required "-sdk" argument to the linker driver.

jakepetroules: The point is that the `DefaultValue` here is not correct -- `SYSROOT` should not influence the `-sdk` flag. Only `SDKROOT` should. Meaning you need this change:

```suggestion
                DefaultValue = "$(SDKROOT)";
```

and which consequently makes this equivalent to the base definition.

Sources/SWBProtocol/MessageSupport.swift
cmcgee1024: Thanks, this revealed several spots where there needed to be consideration for other kinds of build targets.

Sources/SwiftBuild/SWBBuildParameters.swift
cmcgee1024: I've reworked the API following these guidelines and the example provided.

Sources/SwiftBuild/SWBBuildParameters.swift
cmcgee1024: I've moved these out of the appleSDK structure into the run destination info.

Sources/SWBWebAssemblyPlatform/Specs/WasmLd.xcspec
cmcgee1024: Thanks, I've adjusted the WasmLD.xcspec to consider only the SDKROOT, and not the SYSROOT.

Sources/SWBCore/Settings/Settings.swift
cmcgee1024: I've created an initial platformName API as part of the PlatformInfoExtension mechanism and use that to determine platform from triple. For now, it only handles wasm32 and linux, directing those to the WebAssembly platform because it works for both types of Swift SDK's.

Sources/SwiftBuild/SWBBuildParameters.swift
```diff
@@ -36,29 +36,145 @@ public struct SWBBuildParameters: Codable, Sendable {
...
-    public var sdk: String
-    public var sdkVariant: String?
+    public struct SWBBuildTarget: Codable, Sendable {
+        fileprivate var _internalBuildTarget: InternalBuildTarget
+
+        public static func appleSDK(platform: String, sdk: String, sdkVariant: String?) -> SWBBuildTarget {
+            SWBBuildTarget(_internalBuildTarget: .appleSDK(platform: platform, sdk: sdk, sdkVariant: sdkVariant))
+        }
+
+        public static func swiftSDK(sdkManifestPath: String, triple: String) -> SWBBuildTarget {
```
jakepetroules, line 46: One thing is realized is that technically we have a third type of SDK: toolchain SDKs (currently, for Windows and Android), which are installed as part of the toolchain on Windows. In the fullness of time perhaps those can just become Swift SDKs, but right now in Swift Build they're modeled as "appleSDK", which is a bit confusing.

I'm not suggesting a third case be added, but I think for now it would be best to rename `appleSDK` to `toolchainSDK` in order to better reflect that it's also used for the Windows and Android SDKs embedded in the toolchain on Windows hosts, as well as the ability to target the host platform on other platforms' toolchains like Linux and FreeBSD.

Sources/SWBTestSupport/RunDestinationTestSupport.swift
```diff
@@ -299,20 +300,28 @@ extension RunDestinationInfo {
...
     }
 
     package func imageFormat(_ core: Core) -> ImageFormat {
-        switch platform {
-        case "webassembly":
+        switch buildTarget {
+        case let .appleSDK(platform: platform, _, _):
+            switch platform {
+            case "webassembly":
+                fatalError("not implemented")
```
jakepetroules, line 315: there should be a new enum case added for wasm, this could be done in a follow-on though

Sources/SWBTestSupport/RunDestinationTestSupport.swift
```diff
@@ -326,7 +335,12 @@ extension RunDestinationInfo {
         }
         switch imageFormat(core) {
         case .elf:
-            environment.prependPath(key: "LD_LIBRARY_PATH", value: toolchain.path.join("usr/lib/swift/\(platform)").str)
+            if case let .appleSDK(platform: platform, _, _) = buildTarget {
+                environment.prependPath(key: "LD_LIBRARY_PATH", value: toolchain.path.join("usr/lib/swift/\(platform)").str)
+            } else {
+                // Fall back to the OS provided Swift runtime
```
jakepetroules: If it's a Swift SDK you still need to specify this env var. You can parse the triple using LLVMTriple and use the 'os' component as the platform.

Sources/SWBWebAssemblyPlatform/Plugin.swift
```diff
@@ -45,6 +45,15 @@ struct WebAssemblyPlatformExtension: PlatformInfoExtension {
             ])
         ]
     }
+
+    func platformName(triple: LLVMTriple) -> String? {
+        // FIXME this shouldn't be recommending webassembly for linux triples
+        if triple.arch == "wasm32" || triple.system == "linux" {
```
jakepetroules: GenericUnixPlatformInfoExtension should supply the Linux condition, not WebAssemblyPlatformExtension.

Also, [I suggested](https://github.com/swiftlang/swift-build/pull/966#discussion_r2615924274) basing the condition on `triple.os.hasPrefix("wasi")` for a reason -- there is also a wasm64 arch and it's one less thing to update later when that comes online, so we shouldn't make it conditional on the architecture.

Sources/SwiftBuild/SWBBuildServiceSession.swift
```diff
@@ -838,7 +838,13 @@ extension SWBBuildParameters {
 
 fileprivate extension RunDestinationInfo {
     init(_ x: SWBRunDestinationInfo) {
-        self.init(platform: x.platform, sdk: x.sdk, sdkVariant: x.sdkVariant, targetArchitecture: x.targetArchitecture, supportedArchitectures: OrderedSet(x.supportedArchitectures), disableOnlyActiveArch: x.disableOnlyActiveArch, hostTargetedPlatform: x.hostTargetedPlatform)
+        if let appleSDK = x.buildTarget.asAppleSDK() {
+            self.init(buildTarget: .appleSDK(platform: appleSDK.platform, sdk: appleSDK.sdk, sdkVariant: appleSDK.sdkVariant), targetArchitecture: x.targetArchitecture, supportedArchitectures: OrderedSet(x.supportedArchitectures), disableOnlyActiveArch: x.disableOnlyActiveArch, hostTargetedPlatform: x.hostTargetedPlatform)
+        } else if let swiftSDK = x.buildTarget.asSwiftSDK() {
+           self.init(buildTarget: .swiftSDK(sdkManifestPath: swiftSDK.sdkManifestPath, triple: swiftSDK.triple), targetArchitecture: x.targetArchitecture, supportedArchitectures: OrderedSet(x.supportedArchitectures), disableOnlyActiveArch: x.disableOnlyActiveArch, hostTargetedPlatform: x.hostTargetedPlatform)
+        } else {
+            fatalError("not implemented")
```
jakepetroules: You should be able to switch over the internal enum rather than use an if-else chain with an unreachable else case.

Sources/SWBCore/Settings/Settings.swift
```diff
@@ -1798,20 +1800,46 @@ private class SettingsBuilder: ProjectMatchLookup {
...
+                sdk = try project.map {
+                    switch parameters.activeRunDestination?.buildTarget {
+                    case let .swiftSDK(sdkManifestPath: sdkManifestPath, triple: triple):
+                        let llvmTriple = try LLVMTriple(triple)
+
+                        var platformName: String? = nil
+                        for platformExtension in core.pluginManager.extensions(of: PlatformInfoExtensionPoint.self) {
+                            platformName = platformExtension.platformName(triple: llvmTriple)
+                            if platformName != nil {
+                                break
```
jakepetroules: Instead of using a for loop and a `break`, we should collect all the possible answers into a Set<String>, and then issue an error if there are multiple matches. Like so:

```swift
let platformNames = Set(core.pluginManager.extensions(of: PlatformInfoExtensionPoint.self).compactMap { $0.platformName(triple: llvmTriple) })
guard let platformName = platformNames.only else {
    errors.append("conflicting platforms for \(triple): \(platformNames)")
}
```

We never want this to nondeterministically or unpredictably give multiple answers in the unlikely case of a conflict.

Sources/SWBCore/Settings/Settings.swift
```diff
@@ -1798,20 +1800,46 @@ private class SettingsBuilder: ProjectMatchLookup {
...
+               case let .appleSDK(platform: _, sdk: activeSDK, sdkVariant: _) = parameters.activeRunDestination?.buildTarget {
                 sdkroot = activeSDK
             }
 
             do {
-                sdk = try project.map { try sdkRegistry.lookup(nameOrPath: sdkroot, basePath: $0.sourceRoot, activeRunDestination: parameters.activeRunDestination) } ?? nil
+                sdk = try project.map {
+                    switch parameters.activeRunDestination?.buildTarget {
+                    case let .swiftSDK(sdkManifestPath: sdkManifestPath, triple: triple):
+                        let llvmTriple = try LLVMTriple(triple)
```
jakepetroules: Please factor this logic into a function like: `func platform(core: Core, triple: String, errors: inout [String]) -> String?` or similar, it's too much logic to be duplicated 3 times.

Sources/SWBWebAssemblyPlatform/Specs/WasmLd.xcspec
jakepetroules: You should still remove the `SWIFTC_SDKROOT_LINKER_INPUT` definition from WasmLd.xcspec though, since it's now identical to the base definition in Ld.xcspec and is therefore a no-op.

Sources/SWBCore/Settings/Settings.swift
jakepetroules: Yep, that structure's looking good. That said, please consider adding the other values based on what I suggested above. Though if you want to leave them for another PR and focus this one on Wasm, that's fine too.

Sources/SwiftBuild/SWBBuildParameters.swift
cmcgee1024: Do you expect that these toolchainSDK would retain the platform, sdk and sdk variant as the defining variables for them, or is this tuple specific to Apple toolchains?

Sources/SWBTestSupport/RunDestinationTestSupport.swift
cmcgee1024: I haven't really changed this case at all. It was there in the original and so far in my testing the image format hasn't been needed when using any of the Swift SDK's.

Sources/SWBWebAssemblyPlatform/Specs/WasmLd.xcspec
cmcgee1024: There's still the interim UnixLd.xcspec that interferes, so this is needed to override that until it is fixed.

Sources/SWBCore/SDKRegistry.swift
```diff
@@ -1153,6 +1156,95 @@ public final class SDKRegistry: SDKRegistryLookup, CustomStringConvertible, Send
...
     }
 
+    public func synthesizedSDK(platform: Platform, sdkManifestPath: String, triple: String) throws -> SDK? {
+        // Let's check the active run destination to see if there's an SDK path that we should be using
+        let llvmTriple = try LLVMTriple(triple)
+
+        let host = hostOperatingSystem
+
+        // Don't allow re-registering the same SDK
+        if let existing = sdksByPath[Path(sdkManifestPath)] {
```
owenv, line 1166: Should the key here include the triple in addition to the path?

Sources/SwiftBuild/SWBBuildParameters.swift
owenv: yeah I think they'll probably continue to be modeled that way

owenv (commented):
This is looking good to me, I think it's pretty much good as a first step to land once we get some basic tests in place. Left one minor comment/question

Sources/SwiftBuild/SWBBuildParameters.swift
jakepetroules: Yes, the modeling should be the same, it's just about making the name more accurate.

Sources/SWBWebAssemblyPlatform/Specs/WasmLd.xcspec
jakepetroules: Ahh, I missed that. That is certainly a bit confusing that the Ld.xcspec setting is passing `-sdk` while the same setting in UnixLd.xcspec is passing `-sysroot`. Something there seems not quite right; I think we need to have two settings in UnixLd.xcspec, one passing `-sdk` and one passing `-sysroot`, since that model applies to all Unix-like platforms (except Apple).

Sources/SWBCore/SDKRegistry.swift
cmcgee1024: The SDK manifests can support multiple triples, so I don't think it makes sense to add it to the path key.

jakepetroules: I think it shouldn't: Swift SDKs can support multiple triples, and I'd like the ability for the frontend to eventually be able to specify multiple at once, like you can today with `--build-system xcode`.

Also, Swift SDKs describe their own supported triples.

Sources/SWBProtocol/MessageSupport.swift
```diff
@@ -318,23 +318,82 @@ public struct BuildParametersMessagePayload: SerializableCodable, Equatable, Sen
...
+        // These are the old coding keys that were previously associated with toolchain SDK's
+        case platform
+        case sdk
+        case sdkVariant
+    }
+
+    public init(from decoder: any Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        do {
```
jakepetroules: Instead of do-catch, conditional decoding would be the right way to do this:

```swift
if let buildTarget = try container.decodeIfPresent(BuildTarget.self, forKey: .buildTarget) {
    self.buildTarget = buildTarget
} else {
    ...
}
```

Sources/SWBProtocol/MessageSupport.swift
```diff
@@ -318,23 +318,82 @@ public struct BuildParametersMessagePayload: SerializableCodable, Equatable, Sen
...
+            self.buildTarget = try container.decode(BuildTarget.self, forKey: .buildTarget)
+        }
+
+        self.targetArchitecture = try container.decode(String.self, forKey: .targetArchitecture)
+        self.supportedArchitectures = try container.decode(OrderedSet.self, forKey: .supportedArchitectures)
+        self.disableOnlyActiveArch = try container.decode(Bool.self, forKey: .disableOnlyActiveArch)
+
+        // Some versions of the message don't have this field
+        do {
+            self.hostTargetedPlatform = try container.decode(String?.self, forKey: .hostTargetedPlatform)
```
jakepetroules: correct/idiomatic way to decode optionals is decodeIfPresent:

```suggestion
            self.hostTargetedPlatform = try container.decodeIfPresent(String.self, forKey: .hostTargetedPlatform)
```

(Also, just to be clear RE "Some versions of the message don't have this field" -- this isn't an Optional solely for backwards compatibility, it's an Optional on purpose because this property is supposed to be nil for most requests)

Sources/SwiftBuild/SWBBuildParameters.swift
```diff
@@ -36,21 +36,131 @@ public struct SWBBuildParameters: Codable, Sendable {
...
+        // These are the old coding keys that were previously associated with toolchain SDK's
+        case platform
+        case sdk
+        case sdkVariant
+    }
+
+    public init(from decoder: any Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        do {
```
jakepetroules: Same comments as in MessageSupport.swift re decodeIfPresent instead of do-catch, and decodeIfPresent vs decode(String?)

Sources/SWBProtocol/MessageSupport.swift
```diff
@@ -318,23 +318,82 @@ public struct BuildParametersMessagePayload: SerializableCodable, Equatable, Sen
...
+    }
+
+    public init(from decoder: any Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        do {
+            // Handle the message payload from earlier versions that didn't have the buildTarget enumeration
+            let platform = try container.decode(String.self, forKey: .platform)
+            let sdk: String = try container.decode(String.self, forKey: .sdk)
+            let sdkVariant: String? = try container.decode(String?.self, forKey: .sdkVariant)
```
jakepetroules, line 358:
```suggestion
            let sdkVariant: String? = try container.decodeIfPresent(String.self, forKey: .sdkVariant)
```

Sources/SwiftBuild/SWBBuildParameters.swift
```diff
@@ -36,21 +36,131 @@ public struct SWBBuildParameters: Codable, Sendable {
...
+
+        public static func toolchainSDK(platform: String, sdk: String, sdkVariant: String?) -> SWBBuildTarget {
+            SWBBuildTarget(_internalBuildTarget: .toolchainSDK(platform: platform, sdk: sdk, sdkVariant: sdkVariant))
+        }
+
+        public static func swiftSDK(sdkManifestPath: String, triple: String) -> SWBBuildTarget {
+            SWBBuildTarget(_internalBuildTarget: .swiftSDK(sdkManifestPath: sdkManifestPath, triple: triple))
+        }
+
+        public func asToolchainSDK() -> (platform: String, sdk: String, sdkVariant: String?)? {
```
jakepetroules: nit: this and asSwiftSDK aren't used anymore and could be removed, unless we expect to need to call them.

Sources/SWBCore/Settings/Settings.swift
```diff
@@ -1265,6 +1265,35 @@ private class SettingsBuilder: ProjectMatchLookup {
         }
     }
 
+    enum FindPlatformError: Error { case message(String) }
+
+    func findPlatform(for triple: String, core: Core) -> Result<Platform,FindPlatformError> {
+        let llvmTriple = try? LLVMTriple(triple)
+
+        guard let llvmTriple else {
```
jakepetroules: I'd do:

```swift
let llvmTriple: LLVMTriple
do {
    llvmTriple = try LLVMTriple(triple)
} catch {
    return .failure(.message("\(error)"))
}
```

The error that LLVMTriple itself throws, contains a little bit richer information which can tell you a bit more about _why_ a triple might be invalid, and will allow that to propagate to the user as LLVMTripleError evolves over time.

Sources/SwiftBuild/SWBBuildParameters.swift
jakepetroules: nit: this and asSwiftSDK aren't used anymore and could be removed, unless we expect to need to call them.

Sources/SWBCore/Settings/Settings.swift
```diff
@@ -1265,6 +1265,35 @@ private class SettingsBuilder: ProjectMatchLookup {
...
+        }
+
+        var platformNames = OrderedSet<String>()
+        for platformExtension in core.pluginManager.extensions(of: PlatformInfoExtensionPoint.self) {
+            if let platformName = platformExtension.platformName(triple: llvmTriple) {
+                platformNames.append(platformName)
+            }
+        }
+
+        guard let platformName = platformNames.first, platformNames.count == 1 else {
```
jakepetroules: `only` is the idiomatic way to do this, which handles checking for exactly one value and returning it.

```suggestion
        guard let platformName = platformNames.only else {
```

Sources/SWBCore/Settings/Settings.swift
```diff
@@ -1265,6 +1265,35 @@ private class SettingsBuilder: ProjectMatchLookup {
...
+
+    func findPlatform(for triple: String, core: Core) -> Result<Platform,FindPlatformError> {
+        let llvmTriple = try? LLVMTriple(triple)
+
+        guard let llvmTriple else {
+            return .failure(.message("invalid triple \(triple)"))
+        }
+
+        var platformNames = OrderedSet<String>()
+        for platformExtension in core.pluginManager.extensions(of: PlatformInfoExtensionPoint.self) {
```
jakepetroules: nit: this is fine from a correctness perspective, but I suggested the `compactMap` way before because that maintains immutability (`let`) and the ordering here is never relevant, so you don't really need the OrderedSet. In terms of keeping the error message deterministic you could just call `.sorted()` on the Set when constructing the string.

Sources/SWBGenericUnixPlatform/Plugin.swift
```diff
@@ -92,6 +92,15 @@ struct GenericUnixPlatformInfoExtension: PlatformInfoExtension {
             ])
         }
     }
+
+    func platformName(triple: LLVMTriple) -> String? {
+        // FIXME this shouldn't be the webassembly platform here
+        if triple.system == "linux" {
+            return "webassembly"
```
jakepetroules: blocking: Linux needs to return the Linux platform, not WebAssembly.

```suggestion
            return "linux"
```

If you want to focus this PR just on WebAssembly, and not handle Linux at the moment, that's fine too, and you can delete this implementation of `func platformName` -- but whatever the case we must not create an incorrect association between Linux triples and the WebAssembly platform.

Sources/SWBGenericUnixPlatform/Plugin.swift
```diff
@@ -92,6 +92,15 @@ struct GenericUnixPlatformInfoExtension: PlatformInfoExtension {
             ])
         }
     }
+
+    func platformName(triple: LLVMTriple) -> String? {
+        // FIXME this shouldn't be the webassembly platform here
+        if triple.system == "linux" {
```
jakepetroules: blocking: You need to check that the triple.environment has a "gnu" prefix or is "musl", since this should not match Android triples.

```suggestion
        if triple.system == "linux" && (triple.environment.hasPrefix("gnu") || triple.environment == "musl") {
```

Sources/SWBCore/Settings/Settings.swift
jakepetroules: I'll close this thread for now - I'm fine if you want to leave the remaining platforms for a later PR, but we do need to make sure the conditions and values are correct for the platforms that you do add this in PR.

Sources/SWBTestSupport/RunDestinationTestSupport.swift
jakepetroules: Fine to do this in a later PR.

Sources/SWBGenericUnixPlatform/Plugin.swift
cmcgee1024: Linux platform appears to be broken at the moment. It doesn't work with the static Linux SDK, but the WebAssembly one works fine in the smoke testing.

Ok, I'll pull this out until the Linux platform bugs can be resolved in a later PR.

jakepetroules:
The only remaining feedback I have so far are the two unresolved comments related to SYSROOT/SDKROOT/-sdk/-sysroot.

Sources/SWBWebAssemblyPlatform/Specs/WasmLd.xcspec
cmcgee1024: I'm hoping to focus this PR on the WASM portion. Refactoring Ld.xcspec and UnixLd.xcspec to better support the distinction between sysroot and sdk could perhaps be done in a separate PR. I am concerned that it will require a bigger change and more testing unrelated to support WASM with Swift SDK's.

Sources/SWBUniversalPlatform/Specs/Ld.xcspec
cmcgee1024: I'm hoping that this problem with the existing Ld.xcspec can be resolved in a separate PR that breaks the build settings out appropriately.

jakepetroules: OK, I suppose that's fine.

Sources/SWBWebAssemblyPlatform/Specs/WasmLd.xcspec
jakepetroules: OK, I suppose that's fine.

jakepetroules (approved):

cmcgee1024:
@swift-ci please test

PR #967, opened 2025-12-09
Fix PIF API which dropped resourceRule field

owenv:


owenv:
@swift-ci test

owenv:
@swift-ci test macOS

owenv:
@swift-ci test macOS

Sources/SwiftBuild/ProjectModel/BuildFile.swift
jakepetroules, line 46: I think all of the default values for instance variables should be removed; they tend to hide this problem. Let the constructors handle the default parameters.

jakepetroules (approved):

PR #970, opened 2025-12-10
Adjust SDK import test criteria

owenv:
Narrow the scope of the expected linked libraries to only the ones explicitly specified by the user's configuration.

rdar://164982001


owenv:
@swift-ci test

PR #973, opened 2025-12-10
Move many Apple platforms build rules under the darwin domain

owenv:
This ensures we don't attempt to apply them when targeting a non-Apple platform.

Closes https://github.com/swiftlang/swift-package-manager/issues/9423

### DIFF
--- a/Sources/SWBTaskConstruction/TaskProducers/TaskProducer.swift
+++ b/Sources/SWBTaskConstruction/TaskProducers/TaskProducer.swift
@@ -218,8 +218,8 @@ public class TaskProducerContext: StaleFileRemovalContext, BuildFileResolution
         }
     }
 
-    public let appShortcutStringsMetadataCompilerSpec: AppShortcutStringsMetadataCompilerSpec
-    let appleScriptCompilerSpec: CommandLineToolSpec
+    public let appShortcutStringsMetadataCompilerSpec: AppShortcutStringsMetadataCompilerSpec?
+    let appleScriptCompilerSpec: CommandLineToolSpec?
     let buildDependencyInfoSpec: BuildDependencyInfoSpec
     public let clangSpec: ClangCompilerSpec
     public let clangAssemblerSpec: ClangCompilerSpec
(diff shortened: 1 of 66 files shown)


owenv:
@swift-ci test

owenv:
@swift-ci test

Sources/SWBTaskConstruction/TaskProducers/TaskProducer.swift
```diff
@@ -343,8 +343,8 @@ public class TaskProducerContext: StaleFileRemovalContext, BuildFileResolution
         //
         // FIXME: These should really be bound even earlier, like in the spec cache. Or at least, we should throw here and just produce a dep graph error if any are missing.
         let domain = settings.platform?.name ?? ""
-        self.appShortcutStringsMetadataCompilerSpec = workspaceContext.core.specRegistry.getSpec("com.apple.compilers.appshortcutstringsmetadata", domain: domain) as! AppShortcutStringsMetadataCompilerSpec
-        self.appleScriptCompilerSpec = workspaceContext.core.specRegistry.getSpec("com.apple.compilers.osacompile", domain: domain) as! CommandLineToolSpec
+        self.appShortcutStringsMetadataCompilerSpec = workspaceContext.core.specRegistry.getSpec("com.apple.compilers.appshortcutstringsmetadata", domain: domain) as? AppShortcutStringsMetadataCompilerSpec
+        self.appleScriptCompilerSpec = workspaceContext.core.specRegistry.getSpec("com.apple.compilers.osacompile", domain: domain) as? CommandLineToolSpec
```
jakepetroules, line 347: Are you sure `as? CommandLineToolSpec` is correct? Should it be `as CommandLineToolSpec?`? I forgot how the generics are set up here, but just worth double checking.

e.g. if this was an instance of OtherKindOfSpec I'd expect this to fail rather than return nil.

owenv: I ended up just reworking getSpec to take an explicit type parameter so we're not relying on coercions and casting behavior here

owenv:
@swift-ci test

PR #974, opened 2025-12-11
Mock out time itself in the elapsed timer and rate limiter tests

jakepetroules:
These occasionally fail nondeterministically; remove the property which allows them to do so (timing).

This also fixes some "off by one" errors that relying on a mock clock revealed -- there was an implicit assumption that sleep() would take a bit more time than requested, and environments with extreme timing sensitivity sometimes failed to do so.


jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

PR #982, opened 2025-12-16
Ensure TARGET_TEMP_DIR values do not collide when package PIF targets differ only in case

owenv:
It turns out that SwiftPM will allow a package to have targets/products whose names only differ in case (and this is somewhat common in real packages), which could cause TARGET_TEMP_DIR collisions and build failures on case-insensitive filesystems like APFS on macOS. Introduce a TARGET_NAME_DISCRIMINANT which is used to differentiate these only when the collision exists.

Notably this isn't enough to solve the issue of colliding products in every case, but it's a necessary prerequisite.

Closes https://github.com/MobileNativeFoundation/XCLogParser.git


owenv:
@swift-ci test

owenv:
@swift-ci test

daveinglis:
FYI, I believe this made our paths longer which now are causing some test failures in swiftPM...

daveinglis:
Note, https://github.com/swiftlang/swift-package-manager/pull/9268 should fix it (using withKnownIssue) so hopefully we can get this in soon

PR #983, opened 2025-12-16
Exclude C/C++-only targets from resource bundle accessor generation

owenv:
Match the behavior of SwiftPM's build system, since only ObjC/ObjC++ compiles can parse the resource accessor header, and C/C++ targets may be compiling on a platform without an ObjC compiler/ObjC-compatible Foundation (Linux, Windows, etc.)


owenv:
@swift-ci test

PR #984, opened 2025-12-16
Provide LOCALIZED_STRING_CODE_COMMENTS build setting definition

matthewseaman:
Adding cousin setting to `LOCALIZED_STRING_MACRO_NAMES` and `LOCALIZED_STRING_SWIFTUI_SUPPORT` to specify whether extracting strings from code comments is desirable. By default it is not.

Like the similar settings, clients of SwiftBuild can use this to inform String Catalog sync and other localization behaviors.

The use case for setting this to YES is for projects that previously relied on this documented feature of `genstrings` in order to specify strings in a source file that would otherwise be dynamically constructed in code.


matthewseaman:
@swift-ci test

matthewseaman:
@swift-ci test macOS

PR #987, opened 2025-12-17
SWBBuildServer: send target change notification when build description regenerates

owenv:
This ensures clients can re-request the target list, since it will have been empty before the build description is computed


owenv:
@swift-ci test

PR #990, opened 2025-12-18
Log instead of ignoring failures in preview info requests (rdar://165722419)

jonathanpenn:
We'd like to rethink the preview info request types and payloads over time. But in the short term we need to get information about failed preview info requests in previews diagnostics we gather.

### DIFF
--- a/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
@@ -21,6 +21,11 @@ public import enum SWBProtocol.BuildAction
 import Foundation
 public import SWBMacro
 
+#if canImport(Darwin) && canImport(os.log)
+// For Previews logging
+import os.log
+#endif
+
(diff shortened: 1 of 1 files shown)


Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
```diff
@@ -3959,3 +3964,14 @@ extension SwiftDiscoveredCommandLineToolSpecInfo {
         return try await discoveredSwiftCompilerInfo(producer, delegate, at: toolPath, blocklistsPathOverride: userSpecifiedBlocklists)
     }
 }
+
+#if canImport(Darwin) && canImport(os.log)
+let previewsLogger = Logger(subsystem: "com.apple.dt.Previews", category: "swift-build")
```
jakepetroules, line 3969: How about subsystem=`org.swiftlang.swift-build`, category=`preview-info`?

jonathanpenn: We need to match the existing subsystem that is used by our public logging profile to selectively un-redact when we need deeper diagnostics.

jakepetroules (approved):

jonathanpenn:
@swift-ci test

PR #991, opened 2025-12-18
SWBUtil should not link SwiftDriver

owenv:
SWBUtil gets pulled in by the top level SwiftBuild library, so we want to keep it's link dependencies minimal


owenv:
@swift-ci test

owenv:
@swift-ci test

PR #992, opened 2025-12-19
Fix stack overflow in macro namespace lookups

pepicrft:
When building projects with many targets using the swiftbuild backend, the process crashes with a stack overflow. The teco project, which has 522 targets, is an example of this.

The `lookupMacroDeclaration` and `lookupConditionParameter` methods traverse parent namespaces recursively. With hundreds of nested namespaces, this blows the call stack.

This PR converts those recursive lookups to iterative ones using a while loop, which can handle any depth.

Fixes swiftlang/swift-package-manager#9270


jakepetroules:
@swift-ci test

owenv:
@swift-ci test

PR #998, opened 2025-12-26
Miscellaneous fixes for Wasm platform toward SwiftPM integration

kateinoigakukun:
- SwiftSDK: Support specifying metadata JSON path as an artifact variant
- SwiftSDK: Fix default Swift SDKs directory to respect `XDG_CONFIG_HOME` on non-macOS
- Wasm: Remove obsolete linker flags hacks now that swiftc linker driver is supported
- Pass `-resource-dir` to linker driver
- Wasm: Override SWIFTC_SDKROOT_LINKER_INPUT to use `-sdk` over `-sysroot` for now
- SWBCore: Impose `SDKROOT` only if SDK is available

Close https://github.com/swiftlang/swift-package-manager/issues/9364

### DIFF
--- a/Sources/SWBCore/DependencyResolution.swift
+++ b/Sources/SWBCore/DependencyResolution.swift
@@ -206,7 +206,17 @@ struct SpecializationParameters: Hashable, CustomStringConvertible {
     func imposed(on parameters: BuildParameters, workspaceContext: WorkspaceContext) -> BuildParameters {
         var overrides = parameters.overrides
         if let sdkRoot {
-            overrides["SDKROOT"] = sdkRoot
+            // The sdkCanonicalName from a Platform is not guaranteed to be available
+            let defaultSDKAvailable: Bool
+            do {
+                let foundSDK = try workspaceContext.core.sdkRegistry.lookup(sdkRoot, activeRunDestination: parameters.activeRunDestination)
+                defaultSDKAvailable = foundSDK != nil
+            } catch {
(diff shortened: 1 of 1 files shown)


Sources/SWBCore/DependencyResolution.swift
```diff
@@ -206,7 +206,17 @@ struct SpecializationParameters: Hashable, CustomStringConvertible {
...
+            let defaultSDKAvailable: Bool
+            do {
+                let foundSDK = try workspaceContext.core.sdkRegistry.lookup(sdkRoot, activeRunDestination: parameters.activeRunDestination)
+                defaultSDKAvailable = foundSDK != nil
+            } catch {
+                defaultSDKAvailable = false
+            }
+            if defaultSDKAvailable {
+                overrides["SDKROOT"] = sdkRoot
+            }
```
kateinoigakukun, line 219: I'm not 100% sure if this is the right approach. Do we really need to override `SDKROOT` here in the first place? The overridden `SDKROOT` was always `platform.sdkCanonicalName`, which is always the same as `platform.name` (`"webassembly"`), but `"webassembly"` SDK is unavailable.

I'm a bit lost here because I can't access the git history before it got open-sourced.

jakepetroules:
@cmcgee1024 already has some changes up in #966 which will be restructuring this code substantially. You may want to coordinate on swiftlang/swift-driver#2053 though.

PR #1007, opened 2026-01-07
Do not hold lock in Registry when creating potentially expensive value

rconnell9:
Do not hold lock in Registry when creating potentially expensive value.

### DIFF
--- a/Sources/SWBUtil/Registry.swift
+++ b/Sources/SWBUtil/Registry.swift
@@ -39,14 +39,19 @@ public final class Registry<K: Hashable & Sendable, V: Sendable>: KeyValueStorag
     ///
     /// NOTE:  Currently there is only one lock for the whole registry.  If contention becomes a problem we could also have a separate lock for each value, and only hold the registry-wide lock while probing for (and potentially creating a slot for) the key.  The per-value lock would then be held while checking for (and potentially creating) the value.  This is possible because the key is known without invoking the creator block.
     public func getOrInsert(_ key: K, _ creator: () throws -> V) rethrows -> V {
-        try dict.withLock {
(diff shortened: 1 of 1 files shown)


rconnell9:
@swift-ci test

Sources/SWBUtil/Registry.swift
```diff
@@ -39,14 +39,19 @@ public final class Registry<K: Hashable & Sendable, V: Sendable>: KeyValueStorag
     ///
     /// NOTE:  Currently there is only one lock for the whole registry.  If contention becomes a problem we could also have a separate lock for each value, and only hold the registry-wide lock while probing for (and potentially creating a slot for) the key.  The per-value lock would then be held while checking for (and potentially creating) the value.  This is possible because the key is known without invoking the creator block.
     public func getOrInsert(_ key: K, _ creator: () throws -> V) rethrows -> V {
```
jakepetroules, line 41: The changes in this PR would contradict the documented behavior of the class:

>  Unlike a Dictionary, which has independent lookup and insertion operations, a registry has an atomic lookup-or-insert operation that takes a constructor block, which is called only if the key isn't already in the registry.  **The block is guaranteed to be called only once, even if multiple threads lookup-or-insert the same key at the same time** (this allows it to have side effects without needing additional checking).

I think if we're going to make this change, we should parameterize the behavior so we can opt in only the use sites where we know it's safe to do so.

e.g. some `policy` parameter of the following:

```swift
enum ValueCreationPolicy {
    case blocking
    case nonBlocking
}
```

rconnell9: I'm not convinced this is a worthwhile change overall.  It looks like just additional redundant work.   I need to dig in further to see if this expensive bit can be pulled out

PR #1008, opened 2026-01-07
Create a common static archive product-type for swiftPM to use

daveinglis:
The product-type that swiftPM uses for static archive and executable source modules are currently the same, which is set to the common object type, these need to be different on windows as we can't link test dlls with static archives currently, which is what the common object type maps to for windows. We need to create a new static archive type that can be mapped differently on windows vs other platforms. Then swiftPM can use the 2 different product types for a static archive and executable source module and these 2 types can be mapped them differently on each platform..

### DIFF
--- a/Sources/SWBApplePlatform/Specs/DarwinProductTypes.xcspec
+++ b/Sources/SWBApplePlatform/Specs/DarwinProductTypes.xcspec
@@ -474,15 +474,4 @@
             com.apple.package-type.mach-o-executable
         );
     },
-    {
-        _Domain = darwin;
-        Type = ProductType;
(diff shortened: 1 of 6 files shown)


daveinglis:
@swift-ci test

daveinglis:
@swift-ci test

daveinglis:
> Can we define the "regular" product type in SWBUniversalPlatform and then override it only in the Windows platform so we maintain a reasonable behavior on bare metal/wasm/etc?

Hmm, should I have done the same when I added org.swift.product-type.common.object ?

owenv:
@swift-ci test

daveinglis:
@swift-ci test

PR #1009, opened 2026-01-07
Drop '-sdk /' from swiftc args when using the fallback sdk on generic unix platforms

owenv:
This fixes a couple miscellaneous issues:
- All headers/modules under / are no longer treated as system content, which is especially important because this usually includes the build directory itself
- Fixes issues with duplicate includes when C++ interop is enabled

### DIFF
--- a/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
@@ -2240,7 +2240,7 @@ public final class SwiftCompilerSpec : CompilerSpec, SpecIdentifierType, SwiftDi
             var environment: [(String, String)] = environmentFromSpec(cbc, delegate)
             environment.append(("DEVELOPER_DIR", cbc.scope.evaluate(BuiltinMacros.DEVELOPER_DIR).str))
             let sdkroot = cbc.scope.evaluate(BuiltinMacros.SDKROOT)
-            if !sdkroot.isEmpty {
+            if !sdkroot.isEmpty && cbc.scope.evaluate(BuiltinMacros.SWIFTC_PASS_SDKROOT) {
                 environment.append(("SDKROOT", sdkroot.str))
             }
             let toolchains = cbc.scope.evaluateAsString(BuiltinMacros.TOOLCHAINS)
--- a/Tests/SWBBuildSystemTests/SwiftDriverTests.swift
+++ b/Tests/SWBBuildSystemTests/SwiftDriverTests.swift
@@ -23,6 +23,8 @@ import SWBCore
 import SWBMacro
 import SWBTaskExecution
 
+import Foundation
+
 @Suite
 fileprivate struct SwiftDriverTests: CoreBasedTests {
     @Test(.requireSDKs(.host))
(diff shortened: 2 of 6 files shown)


owenv:
@swift-ci test

owenv:
@swift-ci test

Tests/SWBBuildSystemTests/SwiftDriverTests.swift
```diff
@@ -146,22 +148,26 @@ fileprivate struct SwiftDriverTests: CoreBasedTests {
                         compileTask.checkCommandLineMatches([.anySequence, "-primary-file", .anySequence, .equal(SRCROOT.join("Sources/\(fileName).swift").str), .anySequence, "-o", .suffix("\(fileName).o")])
                         let env = compileTask.environment.bindingsDictionary
                         #expect(env["DEVELOPER_DIR"] != nil)
-                        #expect(env["SDKROOT"] != nil)
+                        if  try ProcessInfo.processInfo.hostOperatingSystem() != .linux {
```
jakepetroules: Also != freebsd and openbsd?

Tests/SWBBuildSystemTests/SwiftDriverTests.swift
```diff
@@ -146,22 +148,26 @@ fileprivate struct SwiftDriverTests: CoreBasedTests {
...
+                try results.checkTask(.matchTargetName("TargetA"), .matchRulePattern([.or("SwiftMergeModule", "SwiftEmitModule"), "normal", .any, .or("Merging module TargetA", "Emitting module for TargetA")])) { task in
                     if task.ruleIdentifier == "SwiftMergeModule" {
                         task.checkCommandLineMatches([.anySequence, "-merge-modules", .anySequence, .suffix("file1~partial.swiftmodule"), .suffix("file2~partial.swiftmodule"), .anySequence, "-o", .suffix("TargetA.swiftmodule")])
                     } else {
                         task.checkCommandLineMatches([.anySequence, "-emit-module", .anySequence, "-o", .suffix("TargetA.swiftmodule")])
                     }
                     let env = task.environment.bindingsDictionary
                     #expect(env["DEVELOPER_DIR"] != nil)
-                    #expect(env["SDKROOT"] != nil)
+                    if try ProcessInfo.processInfo.hostOperatingSystem() != .linux {
```
jakepetroules: Also != freebsd and openbsd?

Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
```diff
@@ -2240,7 +2240,7 @@ public final class SwiftCompilerSpec : CompilerSpec, SpecIdentifierType, SwiftDi
             var environment: [(String, String)] = environmentFromSpec(cbc, delegate)
             environment.append(("DEVELOPER_DIR", cbc.scope.evaluate(BuiltinMacros.DEVELOPER_DIR).str))
             let sdkroot = cbc.scope.evaluate(BuiltinMacros.SDKROOT)
-            if !sdkroot.isEmpty {
+            if !sdkroot.isEmpty && cbc.scope.evaluate(BuiltinMacros.SWIFTC_PASS_SDKROOT) {
```
jakepetroules, line 2243: do we really need a setting for this, or should we just do `!sdkroot.isEmpty && sdkroot != .root`? Not sure if there's legit cases where we'll want to pass `-sdk /`.

owenv: IMO it's better to be explicit here, it's easier to understand why the argument is suppressed if it's via a setting that was explicitly set on the fallback sdk

owenv:
@swift-ci test

jakepetroules (approved):

owenv:
@swift-ci test macOS

PR #1010, opened 2026-01-07
Skip a slow-running test in unoptimized builds

owenv:
This test is difficult to speed up in unoptimized builds, so I'm skipping it for now


owenv:
@swift-ci test

mhrawdon (approved):

PR #1011, opened 2026-01-07
Ensure a Swift test entrypoint doesn't suppress package resource bundle accessors for ObjC

owenv:
A C/ObjC test target can end up with multiple source languages if generating a test entry point, so this just drops the single language assumption since it's good future-proofing. Resolves https://github.com/swiftlang/swift-package-manager/issues/9453


owenv:
@swift-ci test

PR #1012, opened 2026-01-08
Revert https://github.com/swiftlang/swift-build/pull/1006

owenv:
This may have introduced a test failure in resource dir computation

PR #1013, opened 2026-01-08
Fill out platformName extension point for other platforms

jakepetroules:
Inert until used, but gets it into place. Android, Linux, and FreeBSD in particular are relevant to begin testing with since all of those are Swift SDK based or have Swift SDK variations.

Based on my comments in https://github.com/swiftlang/swift-build/pull/966#discussion_r2615924274.


jakepetroules:
Note this can be landed independently of #1012.

jakepetroules:
@swift-ci test

owenv (approved):

PR #1014, opened 2026-01-08
Reapply: Swift SDK for Wasm using the run destination

owenv:
Reapply the original commits, plus one fix to ensure we don't pass -resource-dir to the linker when the settings' value(s) are empty


owenv:
@swift-ci test

owenv:
@swift-ci test macOS

PR #1016, opened 2026-01-09
Configure Clang Importer search paths when running the API digester

owenv:
When using a version of the digester which supports -Xcc, pass it search paths, preprocessor definitions, module maps, etc. This ensures that it can correctly load clang module dependencies of the module being processed.


owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test

PR #1019, opened 2026-01-12
Add dependency graph analysis to apply -static more correctly to swift compilations targeting Windows

owenv:
Walk the dependency graph to determine whether the code corresponding to a module will be linked into a dll, and apply -static (or don't) to the compilation accordingly.

Not addressed here: If e.g. an object gets linked statically and included in a dll, we're not duplicating its compilation yet, it just doesn't get -static.

### DIFF
--- a/Sources/SWBCore/LinkageDependencyResolver.swift
+++ b/Sources/SWBCore/LinkageDependencyResolver.swift
@@ -126,13 +126,9 @@ actor LinkageDependencyResolver {
             }
         }
 
-        var allTargets = OrderedSet<ConfiguredTarget>()
-        for (target, dependencies) in dependenciesPerTarget {
-            allTargets.insert(target, at: 0)
-            for dependency in dependencies {
-                allTargets.insert(dependency.target, at: 0)
-            }
-        }
+        let allTargets = OrderedSet(topologicalSort(Array(dependenciesPerTarget.keys), successors: { vertex in
+            return dependenciesPerTarget[vertex]?.map { dependency in dependency.target } ?? []
+        }))
 
         return (allTargets, dependenciesPerTarget)
     }
--- a/Sources/SWBTaskConstruction/ProductPlanning/ProductPlan.swift
+++ b/Sources/SWBTaskConstruction/ProductPlanning/ProductPlan.swift
@@ -173,6 +173,10 @@ package final class GlobalProductPlan: GlobalTargetInfoProvider
     /// A map of `MH_BUNDLE` targets to any clients of that target.
     let clientsOfBundlesByTarget: [ConfiguredTarget:[ConfiguredTarget]]
 
+    /// The set of targets which should compile any swiftmodules they emit assuming the corresponding binary will be statically linked.
+    /// Currently, this information is only used when targeting Windows.
+    let targetsWhichShouldCompileSwiftModulesForStaticLinking: Set<ConfiguredTarget>
+
     private static let dynamicMachOTypes = ["mh_execute", "mh_dylib", "mh_bundle"]
 
     // Checks that we have either been passed a configuration override for packages or we are building Debug/Release.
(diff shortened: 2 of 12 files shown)


Sources/SWBCore/LinkageDependencyResolver.swift
```diff
@@ -126,13 +126,9 @@ actor LinkageDependencyResolver {
             }
         }
 
-        var allTargets = OrderedSet<ConfiguredTarget>()
```
owenv, line 129: This was documented as being topologically sorted but wasn't, so there's some possibility of behavior changes here.

owenv:
@swift-ci test

jakepetroules:
> Not addressed here: If e.g. an object gets linked statically and included in a dll, we're not duplicating its compilation yet, it just doesn't get -static.

In other words, this PR fixes all of the cases where `-static` was missing, but we will still have cases where `-static` is incorrectly present (in effect). Right?

Sources/SWBTaskConstruction/ProductPlanning/ProductPlan.swift
```diff
@@ -456,6 +461,46 @@ package final class GlobalProductPlan: GlobalTargetInfoProvider
         }
     }
 
+    // TODO: We should consider an additional pass which duplicates targets which are both incorporated in a shared library and used statically.
+    private static func computeTargetsWhichShouldCompileSwiftModulesForStaticLinking(buildRequest: BuildRequest, buildGraph: TargetBuildGraph, getLinkageGraph: @Sendable () async throws -> TargetLinkageGraph, provisioningInputs: [ConfiguredTarget: ProvisioningTaskInputs], buildRequestContext: BuildRequestContext, delegate: any GlobalProductPlanDelegate) async -> Set<ConfiguredTarget> {
+        // For the time being, we only perform this analysis when targeting Windows, where it's necessary for correctness. In the future, this may be useful to enable elsewhere to enable optimizations.
+        guard buildGraph.allTargets.contains(where: { configuredTarget in
+            buildRequestContext.getCachedSettings(configuredTarget.parameters, target: configuredTarget.target, provisioningTaskInputs: provisioningInputs[configuredTarget]).platform?.name == "windows"
```
jakepetroules, line 468: This "should" apply everywhere but `-static` is a no-op on all platforms except Windows, right? Are there others to which it could conceivably apply in the future? cc @compnerd

Tests/SWBUtilTests/GraphAlgorithmsTests.swift
```diff
@@ -77,6 +77,63 @@ import SWBUtil
...
+        let diamond: [Int: [Int]] = [
+            1: [2, 3],
+            2: [4],
+            3: [4],
+            4: []
+        ]
+        let diamondResult = SWBUtilTests.topologicalSort([1, 2, 3, 4], diamond)
+        #expect(diamondResult.first == 1)
+        #expect(diamondResult.last == 4)
+        #expect(diamondResult.firstIndex(of: 1)! < diamondResult.firstIndex(of: 2)!)
```
jakepetroules: nit: would be preferable to use `#require` instead of force unwrap

Sources/SWBTaskConstruction/ProductPlanning/ProductPlan.swift
owenv: This could conceivably apply to any platform, it's restricted to windows for now to reduce risk.

owenv:
> In other words, this PR fixes all of the cases where -static was missing, but we will still have cases where -static is incorrectly present (in effect). Right?

The opposite. This will not apply -static if a module's code is linked into both a DLL and static library. What we should do in that situation is compile the code twice, once with -static and once without.

owenv:
@swift-ci test

jakepetroules:
> > In other words, this PR fixes all of the cases where -static was missing, but we will still have cases where -static is incorrectly present (in effect). Right?
> 
> The opposite. This will not apply -static if a module's code is linked into both a DLL and static library.

👍

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test

PR #1022, opened 2026-01-13
Ensure we don't skip linking a dylib with no sources composed of static archives

owenv:
SwiftPM triggers this condition when building a DLL when targeting Windows


owenv:
cc @daveinglis

owenv:
@swift-ci test

PR #1024, opened 2026-01-14
Add new _LD_MACHINE_ARCH settings for windows dll linking

daveinglis:
- since the dlls link with .libs instead of .o we need to tell the linker the machine type as it can't auto discover with no .o files
- also don't pass -rdynamic on windows, it just produces a warning that its an unused argument
- update the common.object product type to map to object library so that when building tests with an executable dependency symbols can be properly resolved.

### DIFF
--- a/Sources/SWBWindowsPlatform/Specs/Windows.xcspec
+++ b/Sources/SWBWindowsPlatform/Specs/Windows.xcspec
@@ -111,7 +111,7 @@
         Domain = windows;
         Type = ProductType;
         Identifier = org.swift.product-type.common.object;
-        BasedOn = com.apple.product-type.library.static;
+        BasedOn = org.swift.product-type.library.object;
     },
 
     {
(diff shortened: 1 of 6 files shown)


daveinglis:
@swift-ci test

Sources/SWBWindowsPlatform/Specs/Windows.xcspec
```diff
@@ -111,7 +111,7 @@
         Domain = windows;
         Type = ProductType;
         Identifier = org.swift.product-type.common.object;
-        BasedOn = com.apple.product-type.library.static;
```
owenv, line 114: Is this left over from testing or is it changing because we're adjusting the product type on the SwiftPM side of things?

daveinglis: This should fix the undefined symbol failures when building some tests

daveinglis: Added to the description.

daveinglis:
@swift-ci test

PR #1026, opened 2026-01-15
Include “wholearchive” when linking DLLs

daveinglis:
This is a quick fix for building and exporting symbols for DLLs on Windows. However, it should be revised to apply this to each library required, as we may don't need it for all transitive dependencies.


daveinglis:
@swift-ci test

daveinglis:
@swift-ci test

PR #1027, opened 2026-01-16
Introduce formal build setting for skipping the bundle hook in mergeable libraries

matthewseaman:
Introducing a new linker build setting for mergeable libraries: `SKIP_MERGEABLE_LIBRARY_BUNDLE_HOOK`.

This setting translates to the `-no_merged_libraries_hook` linker flag.

It also prevents code generation of the `__BundleLookupHelper` Swift class and explicitly sets `SWIFT_MODULE_RESOURCE_BUNDLE_UNAVAILABLE` in Swift so that `#bundle` would return an error when used within the library.


matthewseaman:
@swift-ci please test

owenv (approved):

PR #1033, opened 2026-01-21
Fix some new issues with linking DLL

daveinglis:
Seems using WHOLEARCHIVE to make DLLs does not work in most cases switch to using object library type for swift modules which seems to work better


daveinglis:
@swift-ci test

PR #1035, opened 2026-01-22
Explicitly specialize fnmatch implementation

owenv:
This is a hot path for settings construction, so explicitly specializing yields a modest perf benefit to build planning


owenv:
@swift-ci test

jakepetroules (approved):

PR #1036, opened 2026-01-22
MODULE_DEPENDENCIES should not participate in implicit dependency resolution

owenv:
This inadvertently serialized settings construction as it was the first place we constructed these settings objects. This information is currently not used, so remove the computation


owenv:
@swift-ci test

PR #1037, opened 2026-01-22
Convert Plist dictionary keys to utf8 when deserializing

owenv:
This avoids a perf hit later on, as these strings tend to be accessed very often but never mutated as part of settings construction


owenv:
@swift-ci test

PR #1038, opened 2026-01-22
computeDependencyGraph operation should not block other messages

owenv:
Before this change, if both a background index preparation and foreground build tried to compute the dependency graph around the same time, the former could block the latter, causing the build to appear to stall. Update this API so it doesn't block the session's channel

### DIFF
--- a/Sources/SWBBuildService/DependencyGraphMessages.swift
+++ b/Sources/SWBBuildService/DependencyGraphMessages.swift
@@ -87,6 +87,7 @@ struct ComputeDependencyClosureMsg: MessageHandler {
     }
 }
 
+// Maintained for message format compatibility only.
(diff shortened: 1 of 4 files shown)


owenv:
@swift-ci test

jakepetroules (approved):
All it's doing is removing the lock, right?

owenv:
@jakepetroules I ended up not using withInfoOperation here so there was no lock to remove. The main diff is using handleDelegateMessageOnChannel to kick off the task asynchronously and respond back to the client later without blocking other messages. So this does allow more than one dependency graph computation to run in parallel

owenv:
@swift-ci test

owenv:
@swift-ci test

PR #1039, opened 2026-01-22
Eliminate accidental quadratic behavior in commandLineForSignature(for:)

owenv:
Task.commandLine is O(n), and we should probably make it a method to hint at the fact that it is. For now, fix up a specific issue which could cause build operations to be slow to start in projects with extremely long command lines


owenv:
@swift-ci test

PR #1040, opened 2026-01-22
Disable autolink extract and other linker arguments for Wasm SDK

cmcgee1024:
Autolink extract and other library linker line propagation fails on the
mh_object resource type for the Swift standard libraries when pre-linking
using clang because clang doesn't have a resource dir that includes the
libraries. When pre-linking object files into a larger object file
doesn't make use of the libraries at this stage.

Disable autolink extract when using Wasm doesn't require this feature. Block
the adding of the libraries and library search paths for mh_object resource
types.

Linkers for non-Apple platforms don't recognize certain linker parameters
such as deterministic mode and dependency info files. Clear these
parameters out in the generic unix platform.

Update the SwiftSDK struct to reflect the scheme of the JSON files that
are used in the popular SDKs.

### DIFF
--- a/Sources/SWBTestSupport/RunDestinationTestSupport.swift
+++ b/Sources/SWBTestSupport/RunDestinationTestSupport.swift
@@ -334,7 +334,7 @@ extension RunDestinationInfo {
             return environment
         }
         switch imageFormat(core) {
-        case .elf, .wasm:
+        case .elf:
             switch buildTarget {
             case let .toolchainSDK(platform, _, _):
                 environment.prependPath(key: "LD_LIBRARY_PATH", value: toolchain.path.join("usr/lib/swift/\(platform)").str)
@@ -363,7 +363,7 @@ extension RunDestinationInfo {
                 }
             }
             environment.prependPath(key: .path, value: core.developerPath.path.join("Runtimes").join(toolchain.version.description).join("usr/bin").str)
-        case .macho:
+        case .macho, .wasm:
             // Fall back to the OS provided Swift runtime
             break
         }


Sources/SWBCore/SDKRegistry.swift
```diff
@@ -1132,7 +1132,7 @@ public final class SDKRegistry: SDKRegistryLookup, CustomStringConvertible, Send
                     "Type": .plString("SDK"),
                     "Version": .plString(swiftSDK.version),
                     "CanonicalName": .plString(swiftSDK.identifier),
-                    "Aliases": [],
+                    "Aliases": ["Swift SDK"],
```
jakepetroules: Let's not use Aliases for this purpose, it will potentially cause conflicts

Sources/SWBCore/Settings/Settings.swift
```diff
@@ -2707,7 +2707,7 @@ private class SettingsBuilder: ProjectMatchLookup {
             }
 
             sdkTable.push(BuiltinMacros.DYNAMIC_LIBRARY_EXTENSION, literal: imageFormat.dynamicLibraryExtension)
-            sdkTable.push(BuiltinMacros.PLATFORM_REQUIRES_SWIFT_AUTOLINK_EXTRACT, literal: imageFormat.requiresSwiftAutolinkExtract)
+            sdkTable.push(BuiltinMacros.PLATFORM_REQUIRES_SWIFT_AUTOLINK_EXTRACT, literal: imageFormat.requiresSwiftAutolinkExtract && !sdk.aliases.contains("Swift SDK"))
```
jakepetroules: This is not correct, as there are Swift SDKs that target ELF-based platforms (like Linux), and autolink-extract should NOT be disabled there.

Instead, ImageFormat needs a new case for wasm, since it has its own object format that isn't Mach-O, PE/COFF, or ELF. You can then add a new case to the if-else above (`if llvmTargetTripleSys.hasPrefix("wasi")`), then use the wasm object format in that case.

Note that this is what SwiftPM is already doing in its low-level `Triple.ObjectFormat` type. Apparently there's also an xcoff format which is AIX-specific.

cmcgee1024: Disabling the auto-link extract is working in some ELF contexts, such as the Static Linux SDK. How can we distinguish between that case and the ones where auto-link extract is truly required?

In the mean time, I'll add an image format for wasm and get that case working again until we have an answer for Linux cases that don't require auto-link extract.

jakepetroules: We don't have a great answer for how to distinguish Static Linux from "regular" Linux today.

Right now you can use the environment field of the triple to differentiate -- Static Linux is musl, while regular desktop Linux is gnu*. But that's only a proxy - the Static Linux SDK _happens to use_ Musl because it supports static linking well. It could just as well use another libc, or Musl could be used in a non-static context. But we need better ways to distinguish different behaviors in the long term.

Sources/SWBTestSupport/RunDestinationTestSupport.swift
```diff
@@ -334,7 +334,7 @@ extension RunDestinationInfo {
             return environment
         }
         switch imageFormat(core) {
-        case .elf:
+        case .elf, .wasm:
```
jakepetroules: The wasm case here should go with the macho case; wasm runs in a sandboxed environment so there's no LD_LIBRARY_PATH.

Sources/SWBUtil/ProcessInfo.swift
```diff
@@ -337,12 +337,13 @@ public enum ImageFormat {
...
     case elf
     case pe
+    case wasm
 }
 
 extension ImageFormat {
     public var executableExtension: String {
         switch self {
-        case .macho, .elf:
+        case .macho, .elf, .wasm:
```
jakepetroules: "wasm" is the standard file extension for (binary) Wasm executables.

Sources/SWBUtil/ProcessInfo.swift
```diff
@@ -361,6 +362,8 @@ extension ImageFormat {
             return "so"
         case .pe:
             return "dll"
+        case .wasm:
+            return ""
```
jakepetroules: not clear to me what the story is for dynamic linking in wasm these days, but I'd start with "wasm" for now; I bet many things won't like an empty suffix for dylibs.

Sources/SWBUtil/ProcessInfo.swift
```diff
@@ -386,7 +391,7 @@ extension ImageFormat {
 
     public var usesRpaths: Bool {
         switch self {
-            case .macho, .elf:
+            case .macho, .elf, .wasm:
```
jakepetroules: wasm does not have any sort of rpath concept as far as I'm aware

Sources/SWBUtil/ProcessInfo.swift
cmcgee1024: It's my understanding that dynamic linking isn't something that wasm would ever support.

jakepetroules: Based on a brief search I think some Wasm environments (like Emscripten) may support it? Maybe that's non-standard, I'm not sure. See this article for example: https://www.nutrient.io/blog/dynamic-linking-in-webassembly/ 🤷‍♂️

cmcgee1024:
@swift-ci please test

Sources/SWBGenericUnixPlatform/Specs/UnixLd.xcspec
```diff
@@ -166,6 +166,16 @@
                 DefaultValue = YES;
                 Condition = "NO";
             },
+            {
```
jakepetroules, line 169: I don't understand why this would be needed here; both of these are already disabled in WasmLd.xcspec, can you take a look?

cmcgee1024:
@swift-ci please test

Sources/SWBGenericUnixPlatform/Specs/UnixLd.xcspec
cmcgee1024: It fixes some problems I encountered with a previous version of this PR with the static Linux SDK that are probably relevant to other Linux platform cases too. These features of the Apple linker wouldn't be relevant to other Unix linkers.

PR #1041, opened 2026-01-22
Revert "Add dependency graph analysis to apply -static more correctly…

owenv:
… to swift compilations targeting Windows"

This reverts commit 907ac21d943149f08a3832fc54ac55b322e6b4bf.

We decided to handle this differently, reverting this for now while we work out some design decisions


owenv:
cc @daveinglis

owenv:
@swift-ci test

mhrawdon (approved):

PR #1043, opened 2026-01-23
libtool version regex should be more lenient on Apple platforms

owenv:
Make this check a bit more permissive to we don't fallback to an error if the format doesn't exactly match our expectations


owenv:
@swift-ci test

PR #1044, opened 2026-01-23
[Swift] By default, don't pass `-enforce-exclusivity=checked` to Swift

DougGregor:
The Swift compiler enables exclusivity checking using its own defaulting scheme. Rather than swift-build overriding that default, pass nothing by default and the appropriate "checked" or "unchecked" only when explicitly specified. This lets us stage in some changes around the defaulting for Embedded Swift dynamic exclusivity checking.

Fixes rdar://168724934.


DougGregor:
@swift-ci please test

PR #1050, opened 2026-01-27
Drop use of frontend parsable output

owenv:
Frontend parsable output was a hack used to produce structured per-file subtask logs for batch mode compiles using the integrated driver. Introduce a new output parser which constructs this structure manually instead, using the serialized diagnostics. This isn't a huge performance win, but it avoids some expensive JSON parsing in compiles of large modules and lets us phase out the compiler hacks supporting it.

### DIFF
--- a/Sources/SWBCore/SwiftOutputParsing.swift
+++ b/Sources/SWBCore/SwiftOutputParsing.swift
@@ -43,7 +43,7 @@ public final class SwiftCompilerOutputParser: TaskOutputParser {
         case nil:
             exitStatus = .buildSystemCanceledTask
         }
-        for entry in task.type.serializedDiagnosticsInfo(task, localFS) {
+        for entry in task.type.serializedDiagnosticsInfo(task, workspaceContext.fs) {
             if let sourceFilePath = entry.sourceFilePath {
                 // FIXME: find a better way to get at these
                 let variant = task.ruleInfo[1]


owenv:
@swift-ci test

owenv:
@swift-ci test

Sources/SWBCore/SwiftOutputParsing.swift
```diff
@@ -0,0 +1,554 @@
...
+        case .exit(let status, _)?:
+            exitStatus = status
+        case .failedSetup?:
+            exitStatus = .buildSystemCanceledTask
+        case .skipped?:
+            exitStatus = .exit(0)
+        case nil:
+            exitStatus = .buildSystemCanceledTask
+        }
+        for entry in task.type.serializedDiagnosticsInfo(task, localFS) {
```
jakepetroules:
```suggestion
        for entry in task.type.serializedDiagnosticsInfo(task, workspaceContext.fs) {
```

Sources/SWBCore/TaskGeneration.swift
```diff
@@ -935,6 +935,30 @@ extension ConditionallyStartable {
     }
 }
 
+/// Information about serialized diagnostics for a source file.
+public struct SerializedDiagnosticInfo: Sendable, Serializable {
```
jakepetroules, line 939: suggestion: SerializableCodable could allow omitting the serialization implementations.

owenv: This bit is very perf-sensitive so I'd like to err on the side of avoiding JSON serialization for now

owenv:
@swift-ci test

PR #1051, opened 2026-01-27
[Caching] Make sure .cas-config is deterministic

cachemeifyoucan:
Use sortedKey when generating JSON .cas-config file. This ensures the file generated is deterministic. This fixes a non-determinsitic failure when building using a CAS that requires a JSON configuration that has more than 1 key.

rdar://168957458

### DIFF
--- a/Sources/SWBTaskConstruction/TaskProducers/WorkspaceTaskProducers/CASConfigFileTaskProducer.swift
+++ b/Sources/SWBTaskConstruction/TaskProducers/WorkspaceTaskProducers/CASConfigFileTaskProducer.swift
@@ -27,8 +27,13 @@ final class CASConfigFileTaskProducer: StandardTaskProducer, TaskProducer {
     func generateTasks() async -> [any SWBCore.PlannedTask] {
         var tasks = [any PlannedTask]()
         do {
-            let casConfigFiles = try Dictionary(try await targetContexts.concurrentMap(maximumParallelism: 100) { (targetContext: TaskProducerContext) async throws -> (Path, ByteString)? in
+            let casConfigFiles = try Dictionary(grouping: await targetContexts.concurrentMap(maximumParallelism: 100) { (targetContext: TaskProducerContext) async throws -> (Path, ByteString)? in
                 let scope = targetContext.settings.globalScope
+
+                // Aggregated targets doesn't need to build anything.
+                if targetContext.configuredTarget?.target.type == .aggregate {
+                    return nil
+                }
 
                 // If compilation caching is not on, then there is no file to write.
                 // The condition here is more relax than the actual check in the compile task generation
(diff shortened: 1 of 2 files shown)


cachemeifyoucan:
@swift-ci please test

Sources/SWBTaskConstruction/TaskProducers/WorkspaceTaskProducers/CASConfigFileTaskProducer.swift
```diff
@@ -47,7 +47,9 @@ final class CASConfigFileTaskProducer: StandardTaskProducer, TaskProducer {
                     let CASPath: String
                     let PluginPath: String?
                 }
-                let content = try JSONEncoder().encode(CASConfig(CASPath: casOpts.casPath.str, PluginPath: casOpts.pluginPath?.str))
+                let encoder = JSONEncoder()
+                encoder.outputFormatting = .sortedKeys
```
jakepetroules: Can you please add .withoutEscapingSlashes as well? This data contains path strings, and the default behavior of escaping / is really obnoxious to look at.

cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:
@swift-ci please test

Sources/SWBTaskConstruction/TaskProducers/WorkspaceTaskProducers/CASConfigFileTaskProducer.swift
```diff
@@ -47,12 +47,14 @@ final class CASConfigFileTaskProducer: StandardTaskProducer, TaskProducer {
...
-                let content = try JSONEncoder().encode(CASConfig(CASPath: casOpts.casPath.str, PluginPath: casOpts.pluginPath?.str))
+                let encoder = JSONEncoder()
+                encoder.outputFormatting = [.sortedKeys, .withoutEscapingSlashes]
+                let content = try encoder.encode(CASConfig(CASPath: casOpts.casPath.str, PluginPath: casOpts.pluginPath?.str))
                 let path = scope.evaluate(BuiltinMacros.TARGET_TEMP_DIR).join(".cas-config")
                 return (path, ByteString(content))
             }.compactMap { $0 }, uniquingKeysWith: { first, second in
                 guard first == second else {
-                    throw StubError.error("Unexpected difference in CAS config file.\nPath: \(first.asString)\nContent:\(second.asString)")
+                    throw StubError.error("Unexpected difference in CAS config file.\n\"\(first.asString)\"\nvs\n\"\(second.asString)\"")
```
akyrtzi: Is it possible to also print the target path that got associated with different contents?

akyrtzi:
Should we exclude "aggregate" targets from the calculation? Aggregate targets just have a group of dependencies, they don't build something themselves. Excluding them would be like this:

```
                if targetContext.configuredTarget?.target.type == .aggregate {
                    return nil
                }
```

Sources/SWBTaskConstruction/TaskProducers/WorkspaceTaskProducers/CASConfigFileTaskProducer.swift
cachemeifyoucan: Not easily in this function. Need to restructure to see if I can emit info in there.

cachemeifyoucan:
@swift-ci please test

Sources/SWBTaskConstruction/TaskProducers/WorkspaceTaskProducers/CASConfigFileTaskProducer.swift
```diff
@@ -47,19 +52,28 @@ final class CASConfigFileTaskProducer: StandardTaskProducer, TaskProducer {
...
+            }.compactMap { $0 }) { $0.0 }
 
-            for (configFilePath, configFileContent) in casConfigFiles {
+            for (configPath, configs) in casConfigFiles {
+                if configs.count > 1 {
+                    // Check that all configs for this path have the same content
+                    let firstContent = configs[0].1
+                    for conf in configs[1...] {
+                        if conf.1 != firstContent {
+                            throw StubError.error("Inconsistent CAS configuration for path '\(configPath.str)': multiple targets produce different configurations for the same path")
```
akyrtzi: It'd be useful to also print out the contents as before, so it's obvious what the difference is.

cachemeifyoucan:
@swift-ci please test

Sources/SWBTaskConstruction/TaskProducers/WorkspaceTaskProducers/CASConfigFileTaskProducer.swift
```diff
@@ -27,9 +27,14 @@ final class CASConfigFileTaskProducer: StandardTaskProducer, TaskProducer {
     func generateTasks() async -> [any SWBCore.PlannedTask] {
         var tasks = [any PlannedTask]()
         do {
-            let casConfigFiles = try Dictionary(try await targetContexts.concurrentMap(maximumParallelism: 100) { (targetContext: TaskProducerContext) async throws -> (Path, ByteString)? in
+            let casConfigFiles = try Dictionary(grouping: await targetContexts.concurrentMap(maximumParallelism: 100) { (targetContext: TaskProducerContext) async throws -> (Path, ByteString)? in
                 let scope = targetContext.settings.globalScope
 
+                // Aggregated targets doesn't need to build anything.
```
jakepetroules, line 33: This is not strictly true, as aggregate targets can have script phases and copy files phases, both of which could process source code.

cachemeifyoucan: `.cas-config` file is for compilation caching tasks planned by build system though, and those can't exit in aggregated task?

jakepetroules: I don't think so, unless there's some way for a clang invocation in a script phase to hook up to that?

cachemeifyoucan: I don't think so. They can do clang caching inside script phase but that is not hook into the CAS from build system so this config file won't applied to them. They can do `xcodebuild` in script phase but that xcodebuild should write the config file.

The only situation I can think is trying to write a script but want to infer the CAS configuration from the build system so they can consume this file.

cachemeifyoucan: Is this ok to land as it is?

jakepetroules: Seems fine to me. Check with @owenv as well, he's more familiar with the compilation caching.

PR #1052, opened 2026-01-27
Disable autolink extract and other linker arguments for Wasm SDK

cmcgee1024:
Autolink extract and other library linker line propagation fails on the
mh_object resource type for the Swift standard libraries when pre-linking
using clang because clang doesn't have a resource dir that includes the
libraries. When pre-linking object files into a larger object file
doesn't make use of the libraries at this stage.

Disable autolink extract when using Wasm doesn't require this feature. Block
the adding of the libraries and library search paths for mh_object resource
types. Add wasm as an ImageFormat case and set the computed properties
to the values for this format.

Linkers for non-Apple platforms don't recognize certain linker parameters
such as deterministic mode and dependency info files. Clear these
parameters out in the generic unix platform.

Update the SwiftSDK struct to reflect the scheme of the JSON files that
are used in the popular SDKs.

### DIFF
--- a/Sources/SWBCore/Settings/Settings.swift
+++ b/Sources/SWBCore/Settings/Settings.swift
@@ -2702,6 +2702,8 @@ private class SettingsBuilder: ProjectMatchLookup {
                 imageFormat = .macho
             } else if variant.llvmTargetTripleSys == "windows" {
                 imageFormat = .pe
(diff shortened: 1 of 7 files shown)


cmcgee1024:
@swift-ci please test

PR #1053, opened 2026-01-28
Pass clang module cache pruning options down to swift planning tasks

cyndyishida:
Clang has pruning options to clear out stale scanning PCMs from the module cache. If this is not passed reliably, the clang compiler, embedded into the Swift compiler, would default to only pruning 31-day-old PCM's every 7 days. The `swift build` default is instead pruning 4-day-old PCM's every 1 day. This teaches the swift scanner to respect the same build setting used for clang's modular builds.

resolves: rdar://169063500


Tests/SWBTaskConstructionTests/SwiftTaskConstructionTests.swift
```diff
@@ -1229,6 +1229,10 @@ fileprivate struct SwiftTaskConstructionTests: CoreBasedTests {
                     if LibSwiftDriver.supportsDriverFlag(spelled: "-sdk-module-cache-path") {
                         task.checkCommandLineContains(["-sdk-module-cache-path", "/path/to/DerivedData/ModuleCache.noindex"])
                     }
+
+                    // Check module cache pruning options are passed.
+                    task.checkCommandLineContains(["-Xcc","-fmodules-prune-interval=86400"])
```
jakepetroules, line 1234: would be nice to add some comments like "1 day" / "4 days" to allow fast interpretation of the units

cyndyishida: Added a comment where the default is defined.

owenv (approved):

Tests/SwiftBuildTests/GeneratePreviewInfoTests.swift
```diff
@@ -117,7 +117,7 @@ fileprivate struct GeneratePreviewInfoTests: CoreBasedTests {
...
...build/Test.build/Debug-iphoneos/App.build/App-all-target-headers.hmap", "-Xcc", "-iquote", "-Xcc", "\(tmpDir.str)/Test/build/Test.build/Debug-iphoneos/App.build/App-project-headers.hmap", "-Xcc", "-I\(tmpDir.str)/Test/build/Debug-iphoneos/include", "-Xcc", "-I\(tmpDir.str)/Test/build/Test.build/Debug-iphoneos/App.build/DerivedSources-normal/\(activeRunDestination.targetArchitecture)", "-Xcc", "-I\(tmpDir.str)/Test/build/Test.build/Debug-iphoneos/App.build/DerivedSources/\(activeRunDestination.targetArchitecture)", "-Xcc", "-I\(tmpDir.str)/Test/build/Test.build/Debug-iphoneos/App.build/DerivedSources", "-working-directory", "\(tmpDir.str)/Test", "-experimental-emit-module-separately", "-disable-cmo", "-disable-bridging-pch", "\(tmpDir.str)/Test/build/Test.build/Debug-iphoneos/App.build/Objects-normal/\(activeRunDestination.targetArchitecture)/TestFile4.canary.preview-thunk.swift", "-o", "\(tmpDir.str)/Test/build/Test.build/Debug-iphoneos/App.build/Objects-normal/\(activeRunDestination.targetArchitecture)/TestFile4.canary.preview-thunk.o", "-module-name", "App_PreviewReplacement_TestFile4_canary", "-parse-as-library", "-Onone", "-Xfrontend", "-disable-modules-validate-system-headers"
```
cyndyishida, line 120: I am not sure if it's the right thing to do to pass `-fmodules-prune-interval*`, `-fmodules-prune-after*` down for dynamic replacement invocations for previews. Could someone check my reasoning? I noticed other related module invalidation flags get pruned in `SwiftCompilerSpec.generatePreviewInfo` but it seems like that's only for backwards compatibility with older tools. These pruning flags don't have a compatibility problem.

jonathanpenn: I think pruning them out is probably best unless it will cause issues. Keeps things as similar as possible.

Edit: After thinking about this more let's prune it. These parameters seem safe.

cyndyishida:
@swift-ci please test

cyndyishida:
Sorry, I didn't realize that the merge button didn't automatically squash the commits in this PR. Won't make that mistake again.

PR #1054, opened 2026-01-28
Fix issue where static archive products of root packages are not poduced on windows

daveinglis:
Updated what can be static linked when looking for linker inputs (added .objectLibrary)


daveinglis:
@swift-ci test

daveinglis:
@swift-ci test

PR #1056, opened 2026-01-29
Move SDK modules to subdirectory in MODULE_CACHE_DIR

cyndyishida:
This triggers the clang compiler to prune them with the necessary options set from `CLANG_MODULES_PRUNE_INTERVAL` and `CLANG_MODULES_PRUNE_AFTER`.

Resolves: rdar://169107922


cyndyishida:
This is not a riskless change, & its unclear whether the tradeoff is worth it. So I'm going to decline this PR for now.


Did a little experiment locally & a `module file not found` error is possible and pretty reliable in this scenario, where explicit PCMs can be shared, and the interval/after times are short enough e.g. seconds. Imagine project A imports a bunch of modules and project B imports none.

```
        ┌─ Planning Xcode Instance 1 for project A
        │
        │─ Compiling Xcode Instance 1's explicit PCMs
        | 
    ┌───┤  Planning Xcode Instance 2 for project B # Prunes explicit pcms
    │   │
    │   │
    │   └─ Compiling Xcode Instance 1  💥 BOOM 💥
    │
    │
    └─ Compiling Xcode Instance 2
```

PR #1057, opened 2026-01-29
Stub binaries injected into codeless frameworks should have an explicit install name

daveinglis:
rdar://154514099


Tests/SWBBuildSystemTests/MergeableLibrariesBuildOperationTests.swift
```diff
@@ -340,6 +340,9 @@ fileprivate struct MergeableLibrariesBuildOperationTests: CoreBasedTests {
                                 let (slices, linkage) = try macho.slicesIncludingLinkage()
                                 #expect(try slices.flatMap { try $0.buildVersions().map(\.platform) }.only == .iOS)
                                 #expect(linkage == .macho(.dylib))
+
+                                // Verify the install name is @rpath-based and doesn't contain temporary paths (rdar://154514099)
```
owenv: For the purposes of this test, I think it's enough to check that -install_name appeared on the link line and skip the expensive otool lookup

owenv (approved):

Tests/SWBBuildSystemTests/MergeableLibrariesBuildOperationTests.swift
daveinglis: fixed

daveinglis:
@swift-ci test

PR #1058, opened 2026-01-29
[compilation caching] Prefix map source and build directory

cachemeifyoucan:
In order for better relocatability for source and build directory (for derived source and etc.), prefix map source and build directory as well.

rdar://128486666

### DIFF
--- a/Sources/SWBUniversalPlatform/Specs/Clang.xcspec
+++ b/Sources/SWBUniversalPlatform/Specs/Clang.xcspec
@@ -3297,7 +3297,9 @@
                         "-fdepscan-prefix-map-sdk=/^sdk",
                         "-fdepscan-prefix-map-toolchain=/^toolchain",
                         "-fdepscan-prefix-map=$(DEVELOPER_DIR)=/^xcode",
-                        // FIXME: add source and build directory mappings by default
+                        "-fdepscan-prefix-map=$(PROJECT_DIR)=/^src",
+                        "-fdepscan-prefix-map=$(PROJECT_TEMP_DIR)=/^derived",
+                        "-fdepscan-prefix-map=$(BUILT_PRODUCTS_DIR)=/^built",
                     );
                     NO = ();
                 };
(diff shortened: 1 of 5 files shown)


cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:
@swift-ci please test

owenv (approved):

Sources/SWBUniversalPlatform/Specs/Clang.xcspec
```diff
@@ -3297,7 +3297,8 @@
                         "-fdepscan-prefix-map-sdk=/^sdk",
                         "-fdepscan-prefix-map-toolchain=/^toolchain",
                         "-fdepscan-prefix-map=$(DEVELOPER_DIR)=/^xcode",
-                        // FIXME: add source and build directory mappings by default
+                        "-fdepscan-prefix-map=$(SRCROOT)=/^src",
```
jakepetroules: Please use PROJECT_DIR, not SRCROOT. They are usually the same, but PROJECT_DIR is semantically the right one to use as the latter means something else.

cachemeifyoucan:
@swift-ci please test

Sources/SWBUniversalPlatform/Specs/Clang.xcspec
```diff
@@ -3297,7 +3297,8 @@
                         "-fdepscan-prefix-map-sdk=/^sdk",
                         "-fdepscan-prefix-map-toolchain=/^toolchain",
                         "-fdepscan-prefix-map=$(DEVELOPER_DIR)=/^xcode",
-                        // FIXME: add source and build directory mappings by default
+                        "-fdepscan-prefix-map=$(PROJECT_DIR)=/^src",
+                        "-fdepscan-prefix-map=$(PROJECT_TEMP_DIR)=/^derived",
```
benlangmuir, line 3301: Should we also have a mapping for the build products directory where headers could be installed to?

cachemeifyoucan: Maybe a good idea if one target is search for the installed header from another target? I will add it.

cachemeifyoucan:
@swift-ci please test

PR #1059, opened 2026-01-30
[swift][EmitConst] Use driver flag when available

cachemeifyoucan:
Pass `-const-gather-protocols-list` driver flag when available.

rdar://169280604


cachemeifyoucan:
@swift-ci please test

PR #1060, opened 2026-01-31
Teach Swift Build how to work with Command Line Tools installations

jakepetroules:
This adds a few new extensions in the Apple plugin which load the SDKs, synthesize a macOS platform, and synthesize a toolchain, from a Command Line Tools installation, since its internal layout is substantially different from that of an Xcode-based developer directory.

### DIFF
--- a/Sources/SWBApplePlatform/PluginDCLT.swift
+++ b/Sources/SWBApplePlatform/PluginDCLT.swift
@@ -1,0 +1,141 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import SWBUtil
+import SWBCore
+
+extension MutablePluginManager {
+    func registerDeveloperCommandLineToolsExtensions() {
+        register(DeveloperCommandLineToolsSDKRegistryExtension(), type: SDKRegistryExtensionPoint.self)
+        register(DeveloperCommandLineToolsPlatformInfoExtension(), type: PlatformInfoExtensionPoint.self)
+        register(DeveloperCommandLineToolsToolchainRegistryExtension(), type: ToolchainRegistryExtensionPoint.self)
+    }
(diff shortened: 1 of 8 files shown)


jakepetroules:
@swift-ci test

Sources/SWBApplePlatform/PluginDCLT.swift
```diff
@@ -0,0 +1,141 @@
...
+
+extension MutablePluginManager {
+    func registerDeveloperCommandLineToolsExtensions() {
+        register(DeveloperCommandLineToolsSDKRegistryExtension(), type: SDKRegistryExtensionPoint.self)
+        register(DeveloperCommandLineToolsPlatformInfoExtension(), type: PlatformInfoExtensionPoint.self)
+        register(DeveloperCommandLineToolsToolchainRegistryExtension(), type: ToolchainRegistryExtensionPoint.self)
+    }
+}
+
+struct DeveloperCommandLineToolsSDKRegistryExtension: SDKRegistryExtension {
```
bkhouri, line 25: suggestion: can we add `small` (ie: unit tests) automated tests that would validate the newly added code in isolation?  This would allow us to perform fault injection and al and possibly catch behavioural defect.

We could also add fewer `medium` (integration tests) tests, which could tests various component interactions.

owenv:
23:20:06  /home/build-user/swift-build/Tests/SWBCoreTests/ToolchainRegistryTests.swift:50:19: error: type 'TestDataDelegate' does not conform to protocol 'ToolchainRegistryDelegate'

lgtm once the test is fixed up

owenv:
PluginDCLT.swift needs to be added to the CMake build as well

Sources/SWBApplePlatform/PluginDCLT.swift
jakepetroules: I'd like to come back to that a bit later as I'm not thinking of any tests here that would add meaningful coverage other than a full integration test. I think we'd have to plumb the FSProxy through a lot more places, as there are many places in the Core registries in particular that use the localFS, which makes testing difficult.

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

jakepetroules:
>  /home/build-user/swift-build/Tests/SwiftBuildTests/ConsoleCommands/CLIConnection.swift:214:39: error: sending value of non-Sendable type '() async throws -> Void' risks causing data races [#SendingRisksDataRace]

Did the nightly just get broken by language changes somehow?

jakepetroules:
@swift-ci test

PR #1062, opened 2026-02-02
[Test] Make CLIConnection @unchecked Sendable

cachemeifyoucan:
Fix build failure on nightly swift toolchain.

### DIFF
--- a/Tests/SwiftBuildTests/ConsoleCommands/CLIConnection.swift
+++ b/Tests/SwiftBuildTests/ConsoleCommands/CLIConnection.swift
@@ -27,7 +27,7 @@ import SystemPackage
 #endif
 
 /// Helper class for talking to 'swbuild' the tool
-final class CLIConnection {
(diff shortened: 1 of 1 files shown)


cachemeifyoucan:
@swift-ci please test

jakepetroules:
It is definitely NOT Sendable though. Is the compiler wrong, or what's going on here? At minimum, this should have "FIXME: Work around compiler errors, but does not meet Sendable requirements".

cachemeifyoucan:
> It is definitely NOT Sendable though. Is the compiler wrong, or what's going on here? At minimum, this should have "FIXME: Work around compiler errors, but does not meet Sendable requirements".

Oops, I was supposed to mark this as a draft to test some fix without installing nightly compiler. This is not enough. I am unsure if this should be Sendable but the next error that is still unresolved after this looks like compiler is wrong for sure.

cachemeifyoucan:
Closing. Since this looks like a compiler bug.

PR #1063, opened 2026-02-03
Don't use -T with strip unless targeting Apple platforms

owenv:
... As it's not supported by other implementations of the tool

### DIFF
--- a/Sources/SWBApplePlatform/Specs/StripSymbolsDarwin.xcspec
+++ b/Sources/SWBApplePlatform/Specs/StripSymbolsDarwin.xcspec
@@ -1,0 +1,33 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+(
+    {
+        Domain = darwin;
+        Identifier = com.apple.build-tools.strip;
+        Type = Tool;
+        BasedOn = "default:com.apple.build-tools.strip";
+        Options = (
+            {
+                Name = "STRIP_SWIFT_SYMBOLS";
+                Type = Boolean;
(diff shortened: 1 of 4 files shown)


owenv:
@swift-ci test

Sources/SWBApplePlatform/Specs/StripSymbolsDarwin.xcspec
```diff
@@ -0,0 +1,41 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+(
+    {
+        Domain = darwin;
```
jakepetroules, line 15: I think you can omit everything between BasedOn and Options; it should get inherited.

owenv: yep, you're right

owenv:
@swift-ci test

Sources/SWBApplePlatform/Specs/StripSymbolsDarwin.xcspec
jakepetroules: I think there's a bunch of places where we are re-declaring inherited properties unnecessarily, we should aim to clean that up.

jakepetroules (approved):

Tests/SWBTaskConstructionTests/TaskConstructionTests.swift
```diff
@@ -9298,6 +9298,61 @@ fileprivate struct TaskConstructionTests: CoreBasedTests {
...
+
+            let core = try await getCore()
+            let tester = try TaskConstructionTester(core, testProject)
+
+            try await tester.checkBuild(BuildParameters(configuration: "Debug", overrides: [:]), runDestination: .host, fs: fs) { results in
+                try results.checkTask(.matchRuleType("Strip")) { task in
+                    switch try ProcessInfo.processInfo.hostOperatingSystem() {
+                    case .macOS:
+                        task.checkCommandLineContains(["-T"])
+                    case .windows:
```
jakepetroules: nit: last two cases can be collapsed into default

owenv:
@swift-ci test

PR #1066, opened 2026-02-04
Skip linking during installAPI more often when the only inputs are static archives

owenv:
Tweak the recently-updated logic which determines when we create a link task to avoid breaking compatibility with certain mergeable library configurations. Also, drop some dead code around `perVariantOutputPaths`

rdar://169439369


owenv:
@swift-ci test

PR #1069, opened 2026-02-05
Use decodeIfPresent when decoding optional sdkVariant

rconnell9:
Use decodeIfPresent when decoding optional sdkVariant.  Fixes the following:

$swift package init --type executable
$swift run --build-system xcode
error: error: unable to load --buildParametersFile ('/var/folders/j0/_0tkly6x0hn_0nb98ptbkn7c0000gn/T/TemporaryFile.5Uyd0U'): Failed to decode contents of file '/var/folders/j0/_0tkly6x0hn_0nb98ptbkn7c0000gn/T/TemporaryFile.5Uyd0U' as SWBBuildParameters: {
 ...
  }
}


rconnell9:
@swift-ci test

rconnell9:
> Did this cause an actual problem or were you just switching it to the idiomatic form?

Fixes a bug when using the Xcode build-system.  Added more information to the description.

jakepetroules:
Interesting, so they are not equivalent!  I actually called this out in code review earlier as the less idiomatic form but assumed they were still the same in terms of behavior.

PR #1071, opened 2026-02-06
Linux fixes for excessive Swift invalidation

owenv:
1. Modulewrap was tracked as a compilation requirements output instead of a link requirements output. As a result, it could be written after the compilation requirements task completed, invalidating the compile on the next build.
2. In some cases, the Swift driver will write autolink files alongside the module in the same location we were using for the output of our autolink-extract task. Rename the output to avoid this.
3. Add a null build check to the end to end commandLineTool test to avoid regressions in this area

(found these with https://github.com/swiftlang/swift-llbuild/pull/1017, which I should really try to clean up and land)


owenv:
@swift-ci test

owenv:
@swift-ci test

jakepetroules (approved):

PR #1074, opened 2026-02-09
[6.3]Linux fixes for excessive Swift invalidation

owenv:
Explanation:

1. Modulewrap was tracked as a compilation requirements output instead of a link requirements output. As a result, it could be written after the compilation requirements task completed, invalidating the compile on the next build.
2. In some cases, the Swift driver will write autolink files alongside the module in the same location we were using for the output of our autolink-extract task. Rename the output to avoid this.
3. Add a null build check to the end to end commandLineTool test to avoid regressions in this area

Scope: Swift compilation targeting non-Darwin platforms

Original PRs: https://github.com/swiftlang/swift-build/pull/1071

Risk: Low, the change is relatively minimal and targeted.

Testing: Updated automated tests

Reviewers: @jakepetroules


owenv:
@swift-ci test

PR #1077, opened 2026-02-10
Drop use of -add_ast_path/modulewrap when building with explicit modules

owenv:
The compiler is moving to a new approach for registering swiftmodules required by the debugger using -debug-module-path. Disable the redundant build system registration in these cases to simplify things and shorten link command lines.


owenv:
@swift-ci test

rconnell9 (approved):

owenv:
@swift-ci test

PR #1079, opened 2026-02-11
Don't add strip symbols task on Windows

svmkr-dev:
During CI runs of https://github.com/swiftlang/swift-package-manager/pull/9380 it was discovered that some tests were failing on Windows due to strip command not being found.
As [suggested](https://github.com/swiftlang/swift-package-manager/pull/9380#issuecomment-3880953936) this PR aims to turn off adding strip symbols task on Windows builds.

### DIFF
--- a/Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/ProductPostprocessingTaskProducer.swift
+++ b/Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/ProductPostprocessingTaskProducer.swift
@@ -221,6 +221,11 @@ final class ProductPostprocessingTaskProducer: PhasedTaskProducer, TaskProducer 
             return
         }
 
+        // we can't rely on strip.exe being availalbe on Windows
+        guard !scope.evaluate(BuiltinMacros.HOST_PLATFORM).contains("windows") else {
+            return
+        }
+
         // NOTE: These must be captured here; they are mutable and used to define the task order gating.
         let phaseStartNodes = self.phaseStartNodes
         let phaseEndTask = self.phaseEndTask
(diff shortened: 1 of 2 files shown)


Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/ProductPostprocessingTaskProducer.swift
```diff
@@ -221,6 +221,11 @@ final class ProductPostprocessingTaskProducer: PhasedTaskProducer, TaskProducer
             return
         }
 
+        // we can't rely on strip.exe being availalbe on Windows
+        guard !scope.evaluate(BuiltinMacros.HOST_PLATFORM).contains("windows") else {
```
jakepetroules, line 225: Please use:

```suggestion
        guard context.hostOperatingSystem != .windows else {
```

Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/ProductPostprocessingTaskProducer.swift
```diff
@@ -221,6 +221,11 @@ final class ProductPostprocessingTaskProducer: PhasedTaskProducer, TaskProducer
             return
         }
 
+        // we can't rely on strip.exe being availalbe on Windows
```
jakepetroules, line 224: we can't "rely" on it (implying it's _sometimes_ available), or it simply doesn't apply to Windows at all?

(also, "availalbe" is mis-spelled)

svmkr-dev: that's a good question! I used wording straight from Owen's comment because it aligned with my understanding that by default strip is not present on Windows and it would require GNU binutils installation of some kind, but with that in mind i should probably use a stronger wording.

Though after searching more, I found that official Swift Toolchain for Windows (mine installed using winget as per swift.org instructions) contains llvm-strip tool, which does have `-T` option for stripping swift symbols. I'm thinking better solution is to close this PR and open another where xcspec for llvm-strip on windows is added and `crossPlatformStripSwiftSymbols` test added in #1063 is modified to account for that windows tool.
Should I proceed with this solution?

svmkr-dev:
Closing in favor of #1085

PR #1080, opened 2026-02-11
Prevent CAS size limiting during script phases

benlangmuir:
When a script phase runs a build of its own, we do not want to limit the size of the CAS within the sub build, and only the limit (if any) that is set for the parent build should take effect. This avoids unnecessary work and could prevent correctness issues if we ever want to release the build system's access to the CAS temporarily during a build for any reason.

rdar://170093561


benlangmuir:
@swift-ci test

owenv (approved):

PR #1081, opened 2026-02-11
Introduce an optional post-build CAS validation

benlangmuir:
With COMPILATION_CACHE_VALIDATE_POST_BUILD, perform a post-build CAS validation to check the consistency of the CAS data. Our existing pre-build validation covers a specific functional need to validate the CAS after reboot, and will silently recover from invalid data, since that is a "normal" scenario in the current design. The post-build validation on the other hand should never fail unless there is a bug in the CAS implementation, or data corruption coming from a lower-level system like the filesystem, so it will fail loudly and not attempt to recover data immediately so that a bug can be filed.

rdar://169539794


benlangmuir:
@swift-ci test

PR #1082, opened 2026-02-11
Fix CleanupCompileCache desscription path spelling

benlangmuir:
We were printing `Path(_str: "...")` instead of the expected raw path string in the description text (the rule info had the correct spelling already).


benlangmuir:
@swift-ci test

jakepetroules (approved):

PR #1083, opened 2026-02-11
[Caching] Produce a compilation-prefix-map file

cachemeifyoucan:
On top of CASConfig file, also produce the prefix map file when enabling compilation caching and prefix mapping. This writes a JSON file so other tools can try to remap the file back to its on disk path using the dictionary value.

rdar://170087108


cachemeifyoucan:
@swift-ci please test

Sources/SWBTaskConstruction/TaskProducers/WorkspaceTaskProducers/CompilationCachingConfigFileTaskProducer.swift
```diff
@@ -0,0 +1,136 @@
...
+
+                    // If compilation caching is not on, then there is no file to write.
+                    // The condition here is more relax than the actual check in the compile task generation
+                    // since it won't hurt if the file is not used.
+                    guard scope.evaluate(BuiltinMacros.CLANG_ENABLE_COMPILE_CACHE) || scope.evaluate(BuiltinMacros.SWIFT_ENABLE_COMPILE_CACHE) else {
+                        return nil
+                    }
+
+                    // FIXME: we need consistent CAS configuration across all languages.
+                    if !scope.evaluate(BuiltinMacros.COMPILATION_CACHE_REMOTE_SERVICE_PATH).isEmpty && !scope.evaluate(BuiltinMacros.COMPILATION_CACHE_REMOTE_SUPPORTED_LANGUAGES).isEmpty {
```
owenv, line 53: Do we want to still write the prefix map file if it's enabled but only the local cache is being used?

cachemeifyoucan: Yes. The check above only early terminates if CAS configuration can have conflicts (which is only a certain remote caching configuration). Prefix map should be produced if prefix mapping is used

cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:
ping

PR #1084, opened 2026-02-12
Add the new experimentalWindowDLL to SwiftBuild's copy

dschaefer2:
SwiftBuild has a copy of the ArtifactBundleMetadata structure from SwiftPM. Add the new ArtifactType experimentalWindowsDLL so it can continue to parse the artifact bundle metadata.

Also cleaned up a warning about exectuable artifacts which are used by SwiftPM plugins and should just be ignored.

### DIFF
--- a/Sources/SWBTaskConstruction/TaskProducers/BuildPhaseTaskProducers/SourcesTaskProducer.swift
+++ b/Sources/SWBTaskConstruction/TaskProducers/BuildPhaseTaskProducers/SourcesTaskProducer.swift
@@ -688,8 +688,11 @@ package final class SourcesTaskProducer: FilesBasedBuildPhaseTaskProducerBase, F
                 }
                 for (name, artifact) in metadata.artifacts {
                     switch artifact.type {
-                    case .crossCompilationDestination, .executable, .swiftSDK:
+                    case .crossCompilationDestination, .swiftSDK:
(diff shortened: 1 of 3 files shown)


dschaefer2:
@swift-ci please test

Sources/SWBTaskConstruction/TaskProducers/BuildPhaseTaskProducers/SourcesTaskProducer.swift
```diff
@@ -688,9 +688,12 @@ package final class SourcesTaskProducer: FilesBasedBuildPhaseTaskProducerBase, F
                 }
                 for (name, artifact) in metadata.artifacts {
                     switch artifact.type {
-                    case .crossCompilationDestination, .executable, .swiftSDK:
+                    case .crossCompilationDestination, .swiftSDK:
```
owenv, line 691: This might be a PIF generation issue that needs investigation in SwiftPM, generally speaking executables shouldn't be ending up in a target's linked libraries phase

dschaefer2: This is coming from the info.json, not the PIF. Artifact bundles are allowed to list executables along with static libraries. The paths to the executables are made available to the plugins to add commands that call them.

I saw this warning when playing with using executables for the DLLs. SwiftBuild should be ignoring those entries.

PR #1085, opened 2026-02-12
Use llvm-strip as strip tool on Windows

svmkr-dev:
swiftlang/swift-package-manager#9380 turns on deployment post processing for release builds which caused some tests on Windows CI to fail with "strip.exe not found" error.

strip is part of binutils and is not available by default on Windows, but llvm-strip is part of Swift Toolchain for Windows so we can use it instead.

Mutually exclusive with #1079

### DIFF
--- a/Sources/SWBWindowsPlatform/Specs/WindowsStripSymbols.xcspec
+++ b/Sources/SWBWindowsPlatform/Specs/WindowsStripSymbols.xcspec
@@ -1,0 +1,34 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+(
+    {
+        Domain = windows;
+        Identifier = com.apple.build-tools.strip;
+        Type = Tool;
+        BasedOn = "default:com.apple.build-tools.strip";
+        CommandLine = "llvm-strip [options] [input]";
+        Options = (
+            {
+                Name = "STRIP_SWIFT_SYMBOLS";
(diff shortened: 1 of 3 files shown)


jakepetroules:
@swift-ci test

Sources/SWBWindowsPlatform/Specs/WindowsStripSymbols.xcspec
```diff
@@ -0,0 +1,34 @@
+//===----------------------------------------------------------------------===//
```
owenv, line 1: Please add this to the CMakeLists.txt file for this target too. We'll also need a corresponding PR to the Windows installer like this one: https://github.com/swiftlang/swift-installer-scripts/pull/495

owenv: Other than that this change looks good to me, thanks!

owenv:
Testing the combined toolchain build in https://github.com/swiftlang/swift/pull/79684. Assuming that works, this is good to go

owenv:
@swift-ci test

Sources/SWBWindowsPlatform/Specs/WindowsStripSymbols.xcspec
```diff
@@ -0,0 +1,34 @@
...
+        Domain = windows;
+        Identifier = com.apple.build-tools.strip;
+        Type = Tool;
+        BasedOn = "default:com.apple.build-tools.strip";
+        CommandLine = "llvm-strip [options] [input]";
+        Options = (
+            {
+                Name = "STRIP_SWIFT_SYMBOLS";
+                Type = Boolean;
+                Condition = "$(DEBUG_INFORMATION_FORMAT) == dwarf-with-dsym && $(STRIP_STYLE) != all";
```
compnerd, line 24: Really, we should do this with any DWARF build on windows. The concern I have is that I've seen binary corruption with stripping.

owenv: For now this matches existing behavior that previously was inherited from the parent spec, so I'm ok taking this as-is and following up to figure out the right set of defaults.

PR #1091, opened 2026-02-16
Address SwiftDriver API deprecations

owenv:
Adopt Process EnvironmentBlock and pass compilerIntegratedTooling: false when initializing the driver


owenv:
@swift-ci test

PR #1093, opened 2026-02-16
Avoid reporting all per-file subtasks as failed when a batch compile fails

owenv:
Instead, report the files with errors as failed, and the rest as cancelled. This results in a better build log UX with fewer "x failed without emitting errors" messages.


owenv:
@swift-ci test

mhrawdon (approved):

PR #1094, opened 2026-02-17
Disable autolinking when emitting relocatable objects

owenv:
This fixes some issues when cross-compiling with Swift SDKs because the relocatable objects aren't ending up with unexpected flags to link the standard library when the linker isn't configured with the full set of search paths


owenv:
@swift-ci test

jakepetroules (approved):

PR #1095, opened 2026-02-17
Skip RegisterExecutionPolicyException tasks on non-Darwin platforms

owenv:
These were already a no-op on Darwin, but we shouldn't be planning them in the first place

### DIFF
--- a/Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/ProductPostprocessingTaskProducer.swift
+++ b/Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/ProductPostprocessingTaskProducer.swift
@@ -10,7 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-import SWBCore
+package import SWBCore
 import SWBUtil
 import SWBMacro
 import SWBProtocol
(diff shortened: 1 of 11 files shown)


owenv:
@swift-ci test

Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/ProductPostprocessingTaskProducer.swift
jakepetroules: nit: even better if we could add a post-processing tasks extension and move this to the Apple plugin.

jakepetroules (approved):

owenv:
@swift-ci test

Tests/SWBTaskConstructionTests/PostprocessingTaskConstructionTests.swift
```diff
@@ -296,6 +298,45 @@ fileprivate struct PostprocessingTaskConstructionTests: CoreBasedTests {
...
+                        TestSourcesBuildPhase(["SourceFile.m"]),
+                    ]
+                ),
+            ]
+        )
+
+        try await TaskConstructionTester(getCore(), testProject).checkBuild(runDestination: .host) { results in
+            results.checkNoDiagnostics()
+
+            if try ProcessInfo.processInfo.hostOperatingSystem() == .macOS {
```
jakepetroules, line 332: nit: should use the Core's hostOS given we have a Core in this context

PR #1103, opened 2026-02-19
Remove test which only validates SDK content

owenv:
This test isn't actually exercising the build system, it's just examining stub binaries in the SDK.


owenv:
@swift-ci test

PR #1104, opened 2026-02-19
Fix PRELINK_LIBS dependency tracking

rjmansfield:
Use -dependency_info on Darwin platforms to track dependencies on files specified in PRELINK_LIBS. Manually track file paths on platforms where -dependency_info is not supported.

rdar://140459877

### DIFF
--- a/Sources/SWBCore/SpecImplementations/Tools/PrelinkedObjectLink.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/PrelinkedObjectLink.swift
@@ -86,9 +86,37 @@ public final class PrelinkedObjectLinkSpec: CommandLineToolSpec, SpecImplementat
             commandLine += warningLdFlags
         }
         commandLine += cbc.inputs.map({ $0.absolutePath.str })
-        commandLine += cbc.scope.evaluate(BuiltinMacros.PRELINK_LIBS)
+
+        let prelinkLibs = cbc.scope.evaluate(BuiltinMacros.PRELINK_LIBS)
+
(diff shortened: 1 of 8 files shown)


rjmansfield:
@swift-ci test

Sources/SWBCore/SpecImplementations/Tools/PrelinkedObjectLink.swift
```diff
@@ -86,9 +86,37 @@ public final class PrelinkedObjectLinkSpec: CommandLineToolSpec, SpecImplementat
...
+        let prelinkLibs = cbc.scope.evaluate(BuiltinMacros.PRELINK_LIBS)
+
+        let dependencyInfoFile = cbc.scope.evaluate(BuiltinMacros.PRELINK_DEPENDENCY_INFO_FILE)
+        if !dependencyInfoFile.isEmpty {
+            commandLine += prelinkLibs
+        } else {
+            // Library options like -lfoo can't be tracked
+            for lib in prelinkLibs {
+                if !lib.isEmpty {
+                    if lib.hasPrefix("-l") {
```
jakepetroules, line 99: There's a ton of other types of linkage flags on Apple platforms (like -weak-l), see `searchPathFlagsForLd` for an example. Let's make sure if those are relevant in this context to handle them if so.

rjmansfield: This is for non-Apple platforms where `-dependency_info` isn't supported. I think on other platforms it would just be `-l` that would be valid in `PRELINK_LIBS`

owenv (approved):

PR #1105, opened 2026-02-19
Add new product types to supportsInstallAPI check

pmattos:
### Motivation ###

When building Swift packages with SwiftPM's new PIF builder with the `xcodebuild installapi` action, the build system was skipping `swiftmodule` emission with the following warning (e.g., for the https://github.com/swiftlang/swift-subprocess project):

```
Skipping installAPI swiftmodule emission for target 'Subprocess'
```

This occurred because the new PIF builder uses different product type identifiers (`org.swift.product-type.common.object` and `org.swift.product-type.library.static`) that were not recognized by `supportsInstallAPI`. The old PIF builder used `com.apple.product-type.objfile` which was already supported.

Without `swiftmodule` emission, downstream targets cannot compile against the module, breaking `installapi` builds for Swift packages.

### Changes ###

Extended `ProductTypeIdentifier.supportsInstallAPI` to recognize the product types used by SwiftPM's new PIF builder:

- `org.swift.product-type.common.object` — used for object file targets
- `org.swift.product-type.library.static` — used for static library targets

Both types inherit from `com.apple.product-type.objfile` (in `ProductTypes.xcspec`) but have distinct identifiers that need explicit support.

### Fixed Radar ###

* rdar://168454272 *(Skipping installAPI swiftmodule emission for target)*

### DIFF
--- a/Sources/SWBCore/ProductTypeIdentifier.swift
+++ b/Sources/SWBCore/ProductTypeIdentifier.swift
@@ -20,19 +20,6 @@ public struct ProductTypeIdentifier: Equatable {
 }
 
 public extension ProductTypeIdentifier {
-    var supportsInstallAPI: Bool {
-        // Static frameworks/libraries need to run installAPI if they contain Swift source code because we need .swiftmodule to build downstream targets.
-        return stringValue == "com.apple.product-type.framework"
-            || stringValue == "com.apple.product-type.framework.static"
-            || stringValue == "com.apple.product-type.library.dynamic"
-            || stringValue == "com.apple.product-type.library.static"
-            // Swift packages need to run installAPI.
-            || stringValue == "com.apple.product-type.objfile"
-            // SwiftPM's new PIF builder uses these product types.
-            || stringValue == "org.swift.product-type.common.object"
-            || stringValue == "org.swift.product-type.library.static"
-    }
-
     var supportsEagerLinking: Bool {
         return stringValue == "com.apple.product-type.framework"
             || stringValue == "com.apple.product-type.library.dynamic"
--- a/Sources/SWBCore/SpecImplementations/ProductTypes.swift
+++ b/Sources/SWBCore/SpecImplementations/ProductTypes.swift
@@ -315,7 +315,13 @@ public class ProductTypeSpec : Spec, SpecType, @unchecked Sendable {
     ///
     /// In particular, Swift compilation runs during InstallAPI builds even when building static libraries, but only generates TBD files when building dylibs.
     public var supportsInstallAPI: Bool {
-        return ProductTypeIdentifier(identifier).supportsInstallAPI
+        // Static frameworks/libraries need to run installAPI if they contain Swift
+        // source code because we need .swiftmodule to build downstream targets.
+        return conformsTo(identifier: "com.apple.product-type.framework")
+            || conformsTo(identifier: "com.apple.product-type.framework.static")
+            || conformsTo(identifier: "com.apple.product-type.library.dynamic")
+            || conformsTo(identifier: "com.apple.product-type.library.static")
+            || conformsTo(identifier: "com.apple.product-type.objfile") // Swift packages need to run installAPI.
     }
 
     public var supportsEagerLinking: Bool {
--- a/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
@@ -1229,14 +1229,15 @@ public final class SwiftCompilerSpec : CompilerSpec, SpecIdentifierType, SwiftDi
             }
 
             func addTBDEmissionIfRequired() {
-                let typeStr = cbc.scope.evaluate(BuiltinMacros.PRODUCT_TYPE)
-                let productType = ProductTypeIdentifier(typeStr)
-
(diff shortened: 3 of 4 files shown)


Sources/SWBCore/ProductTypeIdentifier.swift
```diff
@@ -28,6 +28,9 @@ public extension ProductTypeIdentifier {
             || stringValue == "com.apple.product-type.library.static"
             // Swift packages need to run installAPI.
             || stringValue == "com.apple.product-type.objfile"
+            // SwiftPM's new PIF builder uses these product types.
```
owenv: I think we should consider making these productType.conformsTo(identifier: "") checks instead of just looking at equality here, which would likely cover this case + future similar issues

pmattos: Yep, it would be nice to do something like this instead:
```swift
public var supportsInstallAPI: Bool {
      conformsTo(identifier: "com.apple.product-type.framework")
   || conformsTo(identifier: "com.apple.product-type.framework.static")
   || conformsTo(identifier: "com.apple.product-type.library.dynamic")
   || conformsTo(identifier: "com.apple.product-type.library.static")
   || conformsTo(identifier: "com.apple.product-type.objfile")
}
```

But given we would then need to move this code from `ProductTypeIdentifier` up to `ProductTypeSpec`, this might complicate this call site (in `ProductPostprocessingTaskProducer.swift`):
```swift
// If we are creating a TBD for this product, then also sign the .tbd file.
//
// FIXME: This is not strictly correct, because there are situations where these methods 
// return true but we don't actually enable InstallAPI. We should resolve this eventually.
//
// FIXME: This is very gross, we currently only have access to the scope here, 
// so we can't use the actual product type object. However, we will hopefully replace 
// all of this logic by eliminating the
let productType = ProductTypeIdentifier(subscope.evaluate(BuiltinMacros.PRODUCT_TYPE))

// TAPI will only be run when the output is a dylib.
// Swift static library may schedule installAPI phase to generate a swiftmodule.
if productType.supportsInstallAPI &&  ...
```

...and I don't feel we should do this neither:
```swift
if (settings.productType?.supportsInstallAPI ?? false) && { ... }
```

pmattos: ...and given we're in a bit of a rush to get this fix in 😅, can we maybe postpone any further refactorings to another PR? Thanks a bunch!

pmattos:
@swift-ci test

Sources/SWBCore/ProductTypeIdentifier.swift
jakepetroules: You can actually do this refactoring quite trivially without other changes -- just move the code to the existing ProductTypeSpec.supportsInstallAPI implementation in ProductTypes.swift

pmattos: OK... but what should I do with this call site then (in `ProductPostprocessingTaskProducer.swift`)?
```swift
let productType = ProductTypeIdentifier(subscope.evaluate(BuiltinMacros.PRODUCT_TYPE))
if productType.supportsInstallAPI { }
```
...note that it's using the `ProductTypeIdentifier` and not the `ProductTypeSpec` you suggested.

pmattos: As I mentioned above, I don't feel we should do this neither:

```swift
if (settings.productType?.supportsInstallAPI ?? false) { }
```
...which would call the updated `ProductTypeSpec.supportsInstallAPI` API.

jakepetroules: Ah, missed that one. I'd do:

```swift
if let productType = settings.productType, productType.supportsInstallAPI && (shouldUseInstallAPI(subscope, settings) || stubAPIDestination(subscope, settings) == .builtProduct) {
```

jakepetroules: And in the one place in SwiftCompilerSpec:

```swift
guard let productType = cbc.producer.productType else {
    return
}

let supportsInstallAPI = productType.supportsInstallAPI && cbc.scope.evaluate(BuiltinMacros.SUPPORTS_TEXT_BASED_API)
```

pmattos: I'm just afraid that `settings.productType` would be always `nil` given the original `FIXME` comment:
```swift
// FIXME: This is very gross, we currently only have access to the scope here, 
// so we can NOT use the actual product type object. However, we will hopefully replace 
// all of this logic by eliminating the
let productType = ProductTypeIdentifier(subscope.evaluate(BuiltinMacros.PRODUCT_TYPE))
```
...but maybe that's outdated :)

jakepetroules: I expect the comment is outdated.

pmattos: OK, thanks! Just pushed the suggested refactoring...

pmattos:
@swift-ci test

pmattos:
@swift-ci test macOS

Sources/SWBCore/ProductTypeIdentifier.swift
jakepetroules: For some background info, it can be nil because Settings can be constructed for things that aren't targets.

Issue #911, opened 2025-11-14
Toolchain directory layout must be computed in a more flexible way on Unix-like platforms


jakepetroules:
For example, in Fedora 43:

```sh
# swift build --build-system swiftbuild
error: Unexpected toolchain layout for Swift installation path: Path(_str: "/usr/libexec/swift/6.2/bin")
error: missing required default toolchain
error: Could not initialize build system
```

main branch will have the same issue. The install prefix is completely configurable and any given Linux distribution could theoretically put it anywhere, so we need to be able to derive these paths without hardcoding specific prefixes.

p5:
With the release of Swift 6.4, packaging Swift in a private, versioned toolchain directory such as /usr/libexec/swift/6.4.0 is blocked by these hardcoded layout assumptions. SwiftPM and Swift Build fail to locate bundled tools and libraries such as llvm-ar and IndexStore. We currently need downstream patches or launchers to make this layout work, so a prefix-independent solution would be valuable

PR #1106, opened 2026-02-19
Add CI integration tests for the static linux SDK

owenv:
This mostly mirrors the WebAssembly setup. Also, mark these as depending on the soundness checks.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -71,6 +71,7 @@ jobs:
   wasm-integration-tests:
     name: WASM Integration Tests
     uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@0.0.6
+    needs: [soundness, space-format-check]
     with:
       enable_linux_checks: false
       enable_windows_checks: false
@@ -80,6 +81,19 @@ jobs:
       wasm_sdk_pre_build_command: SKIP_ANDROID=1 ./.github/scripts/prebuild.sh
       # This is a hack - replace the build command with a test command and drop the --swift-sdk arg the workflow appends.
       wasm_sdk_build_command: "swift test --filter WebAssemblyIntegrationTests #"
+  static-linux-sdk-integration-tests:
+    name: Static Linux SDK Integration Tests
+    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@0.0.6
+    needs: [soundness, space-format-check]
+    with:
+      enable_linux_checks: false
+      enable_windows_checks: false
+      enable_macos_checks: false
+      enable_linux_static_sdk_build: true
+      linux_static_sdk_versions: '["nightly-main"]'
+      linux_static_sdk_pre_build_command: SKIP_ANDROID=1 ./.github/scripts/prebuild.sh
+      # This is a hack - replace the build command with a test command and drop the --swift-sdk arg the workflow appends.
+      linux_static_sdk_build_command: "swift test --filter StaticLinuxSDKIntegrationTests #"
 
   soundness:
     name: Soundness
--- a/Tests/StaticLinuxSDKIntegrationTests/StaticLinuxSDKIntegrationTests.swift
+++ b/Tests/StaticLinuxSDKIntegrationTests/StaticLinuxSDKIntegrationTests.swift
@@ -17,28 +17,31 @@ import SWBProtocol
 @_spi(Testing) import SWBCore
 import SWBUtil
 
-private func findWebAssemblySwiftSDK() -> SwiftSDK? {
-    let wasmSDKs = try? SwiftSDK.findSDKs(targetTriples: ["wasm32-unknown-wasip1"], fs: localFS, hostOperatingSystem: ProcessInfo.processInfo.hostOperatingSystem())
-    let nonEmbeddedWASMSDK = wasmSDKs?.filter { !$0.identifier.contains("embedded") }.only
-    return nonEmbeddedWASMSDK
+private func findStaticLinuxSwiftSDK() -> SwiftSDK? {
+    guard let hostArch = Architecture.hostStringValue else {
+        return nil
+    }
+    let triple = "\(hostArch)-swift-linux-musl"
+    let sdks = try? SwiftSDK.findSDKs(targetTriples: [triple], fs: localFS, hostOperatingSystem: ProcessInfo.processInfo.hostOperatingSystem())
+    return sdks?.first
 }
 
 extension Trait where Self == Testing.ConditionTrait {
-    static var requiresWebAssemblySwiftSDK: Self {
-        enabled("WebAssembly Swift SDK is not installed", {
-            return findWebAssemblySwiftSDK() != nil
+    static var requiresStaticLinuxSwiftSDK: Self {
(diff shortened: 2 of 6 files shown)


owenv:
The current issue here is that if the currently installed toolchain does not have a matching Swift SDK, the workflow will download another toolchain matching the latest Swift SDK. But when that happens, the unit tests don't know about it and attempt to compile with the system fallback toolchain in /usr/bin/swift...

Tests/StaticLinuxSDKIntegrationTests/StaticLinuxSDKIntegrationTests.swift
```diff
@@ -0,0 +1,97 @@
...
+                """
+            }
+
+            let hostArch = try #require(Architecture.hostStringValue)
+            let triple = "\(hostArch)-swift-linux-musl"
+            let swiftSDK = try #require(findStaticLinuxSwiftSDK())
+            let destination = RunDestinationInfo(buildTarget: .swiftSDK(sdkManifestPath: swiftSDK.manifestPath.str, triple: triple), targetArchitecture: hostArch, supportedArchitectures: [hostArch], disableOnlyActiveArch: false)
+            try await tester.checkBuild(runDestination: destination) { results in
+                results.checkNoErrors()
+                let executionResult = try await Process.getOutput(url: URL(fileURLWithPath: projectDir.join("build").join("Debug-webassembly").join("tool").str), arguments: [])
```
jakepetroules: Debug-linux, not Debug-webassembly

Sources/SWBGenericUnixPlatform/Plugin.swift
```diff
@@ -56,6 +56,11 @@ struct GenericUnixDeveloperDirectoryExtension: DeveloperDirectoryExtension {
             return nil
         }
 
+        if let override = ProcessInfo.processInfo.environment["GENERIC_UNIX_DEVELOPER_DIR_TESTING_OVERRIDE"].map({ Path($0) }), override.isAbsolute {
+            print("GENERIC_UNIX_DEVELOPER_DIR_TESTING_OVERRIDE: \(override)")
```
jakepetroules: no print in libraries please

Sources/SWBGenericUnixPlatform/Plugin.swift
```diff
@@ -56,6 +56,11 @@ struct GenericUnixDeveloperDirectoryExtension: DeveloperDirectoryExtension {
             return nil
         }
 
+        if let override = ProcessInfo.processInfo.environment["GENERIC_UNIX_DEVELOPER_DIR_TESTING_OVERRIDE"].map({ Path($0) }), override.isAbsolute {
```
jakepetroules: I wonder if this overlaps with #911?

Can you add a more detailed comment explaining why this is necessary and how it works around the issue? What might a cleaner solution look like longer term?

owenv: It's not really related, I'm just trying to find the least hacky way of dealing with https://github.com/swiftlang/swift-build/pull/1106#issuecomment-3930613559

owenv:
@swift-ci test

owenv:
This change uncovered some linux toolchain registration issues, so the approach is fairly different now

.github/workflows/pull_request.yml
```diff
@@ -79,7 +80,20 @@ jobs:
       wasm_sdk_versions: '["nightly-main"]'
       wasm_sdk_pre_build_command: SKIP_ANDROID=1 ./.github/scripts/prebuild.sh
       # This is a hack - replace the build command with a test command and drop the --swift-sdk arg the workflow appends.
-      wasm_sdk_build_command: "swift test --filter WebAssemblyIntegrationTests #"
+      wasm_sdk_build_command: "env GENERIC_UNIX_DEVELOPER_DIR_TESTING_OVERRIDE=$(dirname $(dirname $(dirname $(alias swift)))) swift test --filter WebAssemblyIntegrationTests #"
```
owenv: TODO: drop this env var I left in by accident

.github/workflows/pull_request.yml
```diff
@@ -79,7 +80,20 @@ jobs:
...
+    needs: [soundness, space-format-check]
+    with:
+      enable_linux_checks: false
+      enable_windows_checks: false
+      enable_macos_checks: false
+      enable_linux_static_sdk_build: true
+      linux_static_sdk_versions: '["nightly-main"]'
+      linux_static_sdk_pre_build_command: SKIP_ANDROID=1 ./.github/scripts/prebuild.sh
+      # This is a hack - replace the build command with a test command and drop the --swift-sdk arg the workflow appends.
+      linux_static_sdk_build_command: "env GENERIC_UNIX_DEVELOPER_DIR_TESTING_OVERRIDE=$SWIFT_EXECUTABLE_FOR_STATIC_LINUX_SDK swift test --filter StaticLinuxSDKIntegrationTests #"
```
owenv: TODO: drop this env var

Tests/StaticLinuxSDKIntegrationTests/StaticLinuxSDKIntegrationTests.swift
```diff
@@ -0,0 +1,115 @@
...
+@Suite
+fileprivate struct StaticLinuxSDKIntegrationTests: CoreBasedTests {
+
+    // Currently, the integration testing GitHub Actions in CI may download a matching toolchain
+    // for the Swift SDK if the base image isn't a match. Currently, we hardcode knowledge of where
+    // that toolchain will be installed, but we should consider updating the swiftlang shared workflows
+    // with a better way of passing along this path.
+    func getStaticLinuxSDKIntegrationTestingCore() async throws -> Core {
+        // If a matching toolchain was downloaded, it will be in /github/home/.swift-toolchains
+        let userToolchainsDir = Path("/github/home/.swift-toolchains")
```
owenv, line 46: I don't like hardcoding this path, but I haven't figured out a great alternative

jakepetroules: What about `Path.homeDirectory.join(".swift-toolchains")`?

owenv:
@swift-ci test

Tests/StaticLinuxSDKIntegrationTests/StaticLinuxSDKIntegrationTests.swift
owenv: Discussed offline, that doesn't resolve to the right location when running in the container

finagolfin:
[All linux toolchain trunk snapshot CI](https://ci.swift.org/job/oss-swift-package-trigger/5089/console) started [failing](https://ci.swift.org/job/oss-swift-package-trigger/5090/console) some [SwiftPM tests today](https://ci.swift.org/job/oss-swift-package-amazon-linux-2-aarch64/4905/console) after this pull was merged, any idea if the first linux toolchain commit here may be causing it?

owenv:
@finagolfin The noWarningFromRemoteDependencies failure at least is actually from https://github.com/swiftlang/swift-build/pull/1134, I reproduced it locally. I'm going to check the other failures quickly and then get a revert up

PR #1107, opened 2026-02-20
Cleanup old WebAssembly SDK synthesis

owenv:
Now that we have more general SDK synthesis for Swift SDKs included in the BuildRequest, I think we should remove this WebAssembly specific code so that we're exposing one more general interface to SwiftPM.


owenv:
@swift-ci test

owenv:
This needs a bit of work to get linking C-only targets working properly

Sources/SWBWebAssemblyPlatform/Plugin.swift
```diff
@@ -54,85 +53,3 @@ struct WebAssemblyPlatformExtension: PlatformInfoExtension {
...
-                    "Type": .plString("SDK"),
-                    "Version": .plString("1.0.0"),
-                    "CanonicalName": .plString(wasmSDK.identifier),
-                    "Aliases": ["webassembly", "wasi"],
-                    "IsBaseSDK": .plBool(true),
-                    "DefaultProperties": .plDict([
-                        "PLATFORM_NAME": .plString("webassembly"),
-                    ].merging(defaultProperties, uniquingKeysWith: { _, new in new })),
-                    "CustomProperties": .plDict([
-                        "LLVM_TARGET_TRIPLE_OS_VERSION": .plString(os),
```
jakepetroules, line 114: This isn't set in the Swift SDK path and I seem to recall it was necessary for some reason. Can we check on that? The other possibility is that it was copy-pasted from Android, where I believe it is more critical due to the differences in triple format compared to other platforms.

Sources/SWBWebAssemblyPlatform/Plugin.swift
```diff
@@ -54,85 +53,3 @@ struct WebAssemblyPlatformExtension: PlatformInfoExtension {
...
-                        "SWIFT_LIBRARY_PATH": .plString(swiftResourceDir.join("wasi").str),
```
jakepetroules, line 115: Also not present, what about this one?

Sources/SWBWebAssemblyPlatform/Plugin.swift
```diff
@@ -54,85 +53,3 @@ struct WebAssemblyPlatformExtension: PlatformInfoExtension {
...
-                        "SWIFT_RESOURCE_DIR": .plString(swiftResourceDir.str),
-                        // HACK: Ld step does not use swiftc as linker driver but instead uses clang, so we need to add some Swift specific flags
-                        // assuming static linking.
-                        // Tracked in https://github.com/swiftlang/swift-build/issues/3
-                        "OTHER_LDFLAGS": .plArray(["-lc++", "-lc++abi", "-resource-dir", "$(SWIFT_RESOURCE_DIR)/clang", "@$(SWIFT_LIBRARY_PATH)/static-executable-args.lnk"]),
```
jakepetroules, line 120: _All_ of these flags are no longer required?

owenv: A lot of these were needed because we used to be using clang as the linker driver instead of swiftc

owenv:
@swift-ci test

Sources/SWBWebAssemblyPlatform/Plugin.swift
owenv: This gets handled by the linker driver now that we use swiftc

Sources/SWBWebAssemblyPlatform/Plugin.swift
owenv: I haven't been able to dig up the context for this, but AFAICT we're correctly assembling the triple without it

jakepetroules: Yeah, who knows. It'll come back if important.

PR #1109, opened 2026-02-20
Add Swift supplementary outputs support

rjmansfield:
Implements support for generating Swift compiler supplementary output files (SIL, LLVM IR, and optimization records) for code size profiling and analysis.

### DIFF
--- a/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
@@ -1439,6 +1439,12 @@ public final class SwiftCompilerSpec : CompilerSpec, SpecIdentifierType, SwiftDi
                 args.append("-emit-dependencies")
             }
 
+            let codesizeProfileEnabled = cbc.scope.evaluate(BuiltinMacros.ENABLE_CODESIZE_PROFILE)
+            let emitOptRecords = codesizeProfileEnabled || cbc.scope.evaluate(BuiltinMacros.SWIFT_EMIT_OPT_RECORDS)
+            if emitOptRecords {
+                args.append("-save-optimization-record")
+            }
+
             // Generate the .swiftmodule from this compilation to a known location.
             //
             // (We don't care about the intermediate partial swiftmodules, so leave those out of the output file map.)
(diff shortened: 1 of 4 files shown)


rjmansfield:
@swift-ci test

rjmansfield:
@swift-ci test

Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
```diff
@@ -1439,6 +1439,12 @@ public final class SwiftCompilerSpec : CompilerSpec, SpecIdentifierType, SwiftDi
                 args.append("-emit-dependencies")
             }
 
+            let codesizeProfileEnabled = cbc.scope.evaluate(cbc.scope.namespace.parseString("$(ENABLE_CODESIZE_PROFILE)")) == "YES"
```
owenv: We should add ENABLE_CODESIZE_PROFILE to BuiltinMacros.swift and evaluate the constant defined there instead of parsing a string to evaluate here

Sources/SWBUniversalPlatform/Specs/Swift.xcspec
```diff
@@ -1173,6 +1173,59 @@
...
+                Condition = "$(ENABLE_CODESIZE_PROFILE) == YES";
+            },
+
+            {
+                Name = "SWIFT_EMIT_OPT_RECORDS";
+                Type = Boolean;
+                DefaultValue = NO;
+                CommandLineArgs = {
+                    YES = (
+                        "-save-optimization-record",
```
owenv: Since you've defined the flag here you shouldn't also need to add it in SwiftCompiler.xcspec

Sources/SWBUniversalPlatform/Specs/Swift.xcspec
```diff
@@ -1173,6 +1173,59 @@
...
+            {
+                Name = "SWIFT_EMIT_IR_FILES";
+                Type = Boolean;
+                DefaultValue = NO;
+            },
+
+            {
+                Name = "SWIFT_SIL_OUTPUT_DIR";
+                Type = Path;
+                Condition = "$(SWIFT_EMIT_SIL_FILES) == YES";
```
owenv: Since this doesn't apply any arguments in the spec itself this condition is a no-op, same for the examples below

Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
```diff
@@ -2739,6 +2760,22 @@ public final class SwiftCompilerSpec : CompilerSpec, SpecIdentifierType, SwiftDi
                 let bitcodeFilePath = objectFileDir.join(objectFilePrefix + ".bc")
                 fileMapEntry.llvmBitcode = bitcodeFilePath.str
             }
+
+            if emitSIL {
+                let silFilePath = (silOutputDir ?? objectFileDir).join(objectFilePrefix + ".sil")
```
owenv: if silOutputDir is a directory, do we need to append the filename here or will the driver choose one?

Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
```diff
@@ -2739,6 +2760,22 @@ public final class SwiftCompilerSpec : CompilerSpec, SpecIdentifierType, SwiftDi
                 let bitcodeFilePath = objectFileDir.join(objectFilePrefix + ".bc")
                 fileMapEntry.llvmBitcode = bitcodeFilePath.str
             }
+
+            if emitSIL {
```
owenv: Can this code move into computeOutputFileMapContents instead of createCommonFileEntry so we don't need to sometimes `nil`-out the SIL entry below?

Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
rjmansfield: We need to append the filename here. The driver treats output file map entries as full file paths.

rjmansfield:
@swift-ci test

PR #1113, opened 2026-02-21
Improve triple parsing and add tests

jakepetroules:
Our version number handling was mangling WASI triples, and would have similar mangled PS4/PS5 triples or other that have version numbers at the end of a component that are normally treated as part of the name rather than part of the version (going by LLVM's implementation as the authority).


jakepetroules:
@swift-ci test

owenv (approved):

jakepetroules:
@swift-ci test

Issue #1111, opened 2026-02-20
Build planning overhead: multiple caching failures cause 60+ second no-change builds for large SPM projects


jverkoey:
In projects with many local Swift packages (~160 packages, ~2,100 targets), the "Compute target dependency graph" phase takes 60+ seconds on **every single build**, even consecutive no-op builds with zero file changes. This phase is not cached between builds, unlike the downstream `BuildDescription` which has signature-based caching. For large monorepos, this makes iterative development painfully slow — every Cmd+B incurs a fixed 60-second tax before any compilation begins.

## Environment

- Xcode 26.0 beta (26A5082a)
- macOS 26.0 (Sequoia)
- Apple M3 Max, 36 GB RAM
- Swift 6.2

## Project Configuration

- **1 Xcode project** (`ElectricSidecar.xcodeproj`) with ~87 local Swift package references
- **~160 resolved packages** (local packages + their transitive dependencies, all local paths)
- **~2,122 targets** as reported by the build system (from PIF cache analysis)
- **~1,155 target definitions** across all Package.swift files in the monorepo
- All packages are local path dependencies (no remote packages)
- `buildImplicitDependencies` is set to `NO` in all schemes

## Steps to Reproduce

1. Open a workspace with 100+ local Swift packages
2. Build the project (Cmd+B) — observe "Compute target dependency graph for package preparation" takes 60+ seconds
3. Wait for build to complete successfully (build succeeded, no errors)
4. **Without modifying any files**, press Cmd+B again
5. Observe "Compute target dependency graph for package preparation" takes 60+ seconds again

## Expected Behavior

On a no-op rebuild with no file changes, the target dependency graph computation should be cached and effectively free. The build system already caches the `BuildDescription` using a signature-based mechanism (`BuildDescriptionManager` with in-memory and on-disk caching). The `TargetBuildGraph` — which is a prerequisite for computing the `BuildDescription` signature — should have an analogous caching mechanism.

## Actual Behavior

The `TargetBuildGraph` is computed from scratch on every build. In the open-source Swift Build code (`Sources/SWBBuildService/PlanningOperation.swift`), the `plan()` method constructs a new `TargetBuildGraph` unconditionally:

```swift
let graph = await TargetBuildGraph(
workspaceContext: workspaceContext,
buildRequest: buildRequest, ...
)

```

There is no caching of this intermediate result. The `TargetDependencyResolver` is instantiated fresh for each planning operation, performing:

[...]

PR #1117, opened 2026-02-21
Sort dry-run output in generatedFilePaths for deterministic build plans

### LINKED ISSUE #1115: xcstringstool compile --dry-run produces non-deterministic output ordering, invalidating BuildDescription cache
## Summary

`xcstringstool compile --dry-run` returns output file paths in a **non-deterministic order** across invocations. Since `XCStringsCompilerSpec.constructCatalogCompilationTask()` records these paths directly into the build manifest without sorting, the `BuildDescription` hash changes on every build — even when nothing in the project has changed. This defeats the in-memory `BuildDescription` cache entirely for any project using `.xcstrings` files.

## Impact

The `BuildDescription` in-memory `HeavyCache` (max 4 entries, no timeout) is designed to avoid re-serializing the build description on consecutive builds. But because the manifest changes every time due to locale path ordering, the cache **never gets a hit**.

For a ~2,100 target project, `CreateBuildDescription` takes ~22-24 seconds cold. With a stable manifest, this drops to ~1 second on cache hit. Every project with `.xcstrings` files pays this cost on every single build.

## Reproduction

```bash
#!/bin/bash
XCSTRINGSTOOL="/Applications/Xcode.app/Contents/Developer/usr/bin/xcstringstool"
OUTDIR="/tmp/xcstrings-repro"
mkdir -p "$OUTDIR"

cat > "$OUTDIR/Test.xcstrings" << 'XCSTRINGS'
{
  "sourceLanguage" : "en",
  "strings" : {
    "hello" : {
      "localizations" : {
        "de" : { "stringUnit" : { "state" : "translated", "value" : "Hallo" } },
        "es" : { "stringUnit" : { "state" : "translated", "value" : "Hola" } },
        "fr" : { "stringUnit" : { "state" : "translated", "value" : "Bonjour" } },
        "it" : { "stringUnit" : { "state" : "translated", "value" : "Ciao" } },
        "ja" : { "stringUnit" : { "state" : "translated", "value" : "こんにちは" } },
        "ko" : { "stringUnit" : { "state" : "translated", "value" : "안녕하세요" } }
      }
    }
  },
  "version" : "1.0"
}
XCSTRINGS

echo "Running xcstringstool compile --dry-run 10 times..."
PREV=""
FLIPS=0
for i in $(seq 1 10); do
  OUTPUT=$("$XCSTRINGSTOOL" compile --dry-run --output-directory "$OUTDIR/out" "$OUTDIR/Test.xcstrings")
  [ "$i" -eq 1 ] && echo "Run 1:" && echo "$OUTPUT" && echo ""
  if [ -n "$PREV" ] && [ "$OUTPUT" != "$PREV" ]; then
    FLIPS=$((FLIPS + 1))
    [ "$FLIPS" -eq 1 ] && echo "Run $i (first change):" && echo "$OUTPUT" && echo ""
  fi
  PREV="$OUTPUT"
done
echo "Order changed $FLIPS / 9 consecutive comparisons."
rm -rf "$OUTDIR"
```

**Output (Xcode 26.3, macOS 26.3):**
```
Running xcstringstool compile --dry-run 10 times...
Run 1:
/tmp/xcstrings-repro/out/pl.lproj/Test.strings
/tmp/xcstrings-repro/out/ja.lproj/Test.strings
/tmp/xcstrings-repro/out/fr.lproj/Test.strings
...

Run 2 (first change):
/tmp/xcstrings-repro/out/pt-BR.lproj/Test.strings
/tmp/xcstrings-repro/out/es.lproj/Test.strings
/tmp/xcstrings-repro/out/pl.lproj/Test.strings
...

Order changed 9 / 9 consecutive comparisons.
```

## Root Cause

https://github.com/swiftlang/swift-build/blob/1d22d904844f18f9eeb862e6bdcc24c3879c8532/Sources/SWBApplePlatform/XCStringsCompiler.swift#L229-L231

The output of `xcstringstool --dry-run` is used verbatim without sorting. These paths flow into the build manifest's command outputs, which are inputs to the `BuildDescription` hash.

## Suggested Fix

Sort the paths after parsing:

```swift
outputs = try await generatedFilePaths(...) { output in
    return output.unsafeStringValue.split(separator: "
").map(Path.init).sorted()
}
```

This ensures deterministic manifests regardless of `xcstringstool`'s output ordering. Ideally `xcstringstool` itself would also be fixed to produce sorted output, but sorting on the consumer side is the robust defense.

## Evidence from Real Project

Examining the `XCBuildData/` directory of a ~2,100 target project after 4 consecutive no-op builds:

```
4425fbd7ff9c03ff1efdb3da9533f7c6.xcbuilddata   # build 4
829205679cee90d64cc99323a2c2165b.xcbuilddata   # build 3
cdc0bf1a3110f16a3180a24963c9b794.xcbuilddata   # build 2
fe3d0067d8ab1c5a0e1272af4ee34b03.xcbuilddata   # build 1
```

4 builds → 4 different hashes. Diffing the manifests shows: **only 20 commands differ** (out of 76,825), and **all 20 are CompileXCStrings-related** — identical file sets, just different ordering. Sorting the output lists makes all 4 manifests byte-identical.

A comparable project without `.xcstrings` files produces stable hashes and gets cache hits as expected.

jverkoey:
## Summary

Sort the output paths returned by `generatedFilePaths` in `CommandLineToolSpec` to ensure deterministic build manifests for all CLI tools that use dry-run output.

Tools like `xcstringstool compile --dry-run` return output file paths in non-deterministic order (e.g. locale iteration from an unordered collection internally). Since `generatedFilePaths` records these paths directly into the build manifest without sorting, the `BuildDescription` hash can change on every build — even when nothing in the project has changed. This defeats the on-disk `BuildDescription` cache entirely.

The fix applies `.sorted()` in the shared `generatedFilePaths` function so that all CLI tools using it (xcstringstool, intentbuilderc, momc, coremlc) benefit from deterministic output ordering.

## Test plan

- Added a test (`xcstringsOutputsAreSorted`) that verifies the outputs from the dry-run step are sorted, ensuring deterministic task construction regardless of the tool's internal iteration order.

Fixes #1115. Part of #1111.


jakepetroules (approved):
Awesome, thanks for the fix!

jakepetroules:
@swift-ci test

Tests/SWBTaskConstructionTests/XCStringsTaskConstructionTests.swift
```diff
@@ -2111,4 +2111,82 @@ fileprivate struct XCStringsTaskConstructionTests: CoreBasedTests {
...
+                    ]
+                )
+            ],
+            developmentRegion: "en"
+        )
+
+        // Return output paths in non-sorted order to ensure the .sorted() call is working.
+        // The mock tool will return these in the order specified:
+        let xcstringsTool = MockXCStringsTool(relativeOutputFilePaths: [
+            "/tmp/Test/Project/Sources/Localizable.xcstrings" : [
```
jakepetroules: Great test!

One small thing, we prefer to use `testProject.path` in new code rather than hardcoding the `/tmp/Test/Project` prefix -- this gives the flexibility to change or randomize the path in the future without having to edit the test.

jverkoey:
Moving this to Draft status until the tests are solid.

owenv:
@swift-ci test

Sources/SWBApplePlatform/XCStringsCompiler.swift
```diff
@@ -227,7 +227,7 @@ public final class XCStringsCompilerSpec: GenericCompilerSpec, SpecIdentifierTyp
             dryRunCommandLine.insert("--dry-run", at: 2)
 
             outputs = try await generatedFilePaths(cbc, delegate, commandLine: dryRunCommandLine, workingDirectory: cbc.producer.defaultWorkingDirectory, environment: environmentFromSpec(cbc, delegate).bindingsDictionary, executionDescription: "Compute XCStrings \(cbc.input.absolutePath.basename) output paths") { output in
-                return output.unsafeStringValue.split(separator: "\n").map(Path.init)
+                return output.unsafeStringValue.split(separator: "\n").map(Path.init).sorted()
```
jakepetroules: I made some unrelated change that conflict slightly with this... can you move the `.sorted()` call to the `generatedFilePaths` function in CommandLineToolSpec.swift? Then it'll apply for all of the other 3 CLI tools that are using `generatedFilePaths` rather than only `xcstringstool`.

jverkoey: Done — moved the `.sorted()` into `generatedFilePaths` in CommandLineToolSpec.swift so it applies to all CLI tools using that function. Pushed in 32da037.

jverkoey: Also merged upstream/main to pick up your AbsolutePath changes — should be conflict-free now.

jakepetroules:
@swift-ci test

jakepetroules:
@jverkoey There's a few test issues this PR ran into. Would you mind updating those so we can take this fix over the finish line?

mamouneyya:
I've been working on the same area in #1203, so I'm familiar with what's going on here. The test failures are just existing `checkOutputs` expectations that assumed the old unsorted order. I put the fix up in #1207 if you want to cherry-pick it (https://github.com/swiftlang/swift-build/pull/1207/changes/3f1d1ee9f7f919342e535f7819ae8301922421aa) or pull it into your branch. Feel free to ignore it if you'd rather do it differently. 🙈

matthewseaman:
Since this PR has been inactive, the same issue was fixed via #1207.

Issue #1116, opened 2026-02-21
CopySwiftPackageResourcesTaskProducer iterates Set non-deterministically, invalidating BuildDescription cache


jverkoey:
## Summary

`CopySwiftPackageResourcesTaskProducer.generateTasks()` stores bundle names in a `Set`, then iterates it directly. Since `Set` has no guaranteed iteration order in Swift, the resource bundle inputs appear in a different order in the build manifest on each build. This changes the `BuildDescription` hash, defeating the in-memory cache.

This is the same class of bug as #1115 (xcstringstool non-deterministic output), and together they account for all observed non-determinism in build plans for projects with Swift package resource bundles and `.xcstrings` files.

## Root Cause

[`CopySwiftPackageResourcesTaskProducer.swift`, line 21](https://github.com/swiftlang/swift-build/blob/main/Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/CopySwiftPackageResourcesTaskProducer.swift#L21):

```swift
let bundlesToEmbed = Set(scope.evaluate(BuiltinMacros.EMBED_PACKAGE_RESOURCE_BUNDLE_NAMES))
```

Then line 68:

```swift
for bundleName in bundlesToEmbed {
```

`Set` iteration order is non-deterministic across process invocations. The generated copy tasks appear in different orders, changing the manifest, changing the build description hash.

## Impact

Same as #1115 — the `BuildDescription` in-memory `HeavyCache` never gets a hit for projects with multiple SPM resource bundles. For a ~2,100 target project with ~20+ resource bundles, `CreateBuildDescription` takes ~22-24 seconds cold vs ~1 second on cache hit.

## Evidence

After fixing #1115 with a shim that sorts `xcstringstool --dry-run` output, consecutive builds of the same project still produce different manifest hashes. Diffing the manifests shows exactly 4 commands differ (one per target), all in their `inputs` field — the resource bundle paths are the same set but in different order:

```
Build 1 inputs[2]: .../ElectricSidecar.app/DesignSystem_SidecarPlusSymbol.bundle
Build 2 inputs[2]: .../ElectricSidecar.app/UI_GenericSymbol.bundle
```

Sorting the input lists makes the manifests byte-identical.

## Suggested Fix

Sort the bundle names before iterating:

```swift
let bundlesToEmbed = Set(scope.evaluate(BuiltinMacros.EMBED_PACKAGE_RESOURCE_BUNDLE_NAMES)).sorted()
```

[...]

PR #1118, opened 2026-02-21
Sort resource bundle names for deterministic build plans

jverkoey:
## Summary

- Sort `EMBED_PACKAGE_RESOURCE_BUNDLE_NAMES` before iteration to ensure deterministic build manifests

`CopySwiftPackageResourcesTaskProducer.generateTasks()` stores bundle names in a `Set`, then iterates it directly. Since `Set` has no guaranteed iteration order in Swift, the resource bundle copy tasks appear in a different order in the build manifest on each build. This changes the `BuildDescription` hash, defeating the in-memory `HeavyCache`.

The fix is a one-line change: append `.sorted()` to the `Set` before the `for` loop.

## Test plan

- [ ] Build a project with multiple Swift package resource bundles multiple times consecutively
- [ ] Verify the `XCBuildData/*.xcbuilddata` directory name (BuildDescription hash) is stable across builds
- [ ] Verify the resource bundle copy task ordering in `manifest.json` is deterministic

## Evidence

After fixing the related xcstringstool non-determinism (#1115), consecutive builds of a ~2,100 target project still produced different manifest hashes. Diffing the manifests showed exactly 4 commands differing — all in their `inputs` field with resource bundle paths in a different order:

```
Build 1 inputs[2]: .../ElectricSidecar.app/DesignSystem_SidecarPlusSymbol.bundle
Build 2 inputs[2]: .../ElectricSidecar.app/UI_GenericSymbol.bundle
```

Sorting the input lists makes all manifests byte-identical.

Resolves https://github.com/swiftlang/swift-build/issues/1116

### DIFF
--- a/Tests/SWBTaskConstructionTests/PackageProductConstructionTests.swift
+++ b/Tests/SWBTaskConstructionTests/PackageProductConstructionTests.swift
@@ -1361,7 +1361,7 @@ fileprivate struct PackageProductConstructionTests: CoreBasedTests {
             groupTree: TestGroup(
                 "SomeFiles",
                 children: [
-                    TestFile("main.swift"),
+                    TestFile("main.c"),
                 ]),
             buildConfigurations: [
                 TestBuildConfiguration("Debug", buildSettings: [
@@ -1371,8 +1371,8 @@ fileprivate struct PackageProductConstructionTests: CoreBasedTests {
             ],
             targets: [
                 TestStandardTarget(
-                    "MyApp",
-                    type: .application,
+                    "MyTool",
+                    type: .commandLineTool,
                     buildConfigurations: [
                         TestBuildConfiguration("Debug", buildSettings: [
                             "PRODUCT_NAME": "$(TARGET_NAME)",
@@ -1383,7 +1383,7 @@ fileprivate struct PackageProductConstructionTests: CoreBasedTests {
                     ],
                     buildPhases: [
                         TestSourcesBuildPhase([
-                            "main.swift",
+                            "main.c",
                         ]),
                     ]
                 ),


jakepetroules:
Hmm, I'm not sure this is the right fix. Are you saying the nondeterminism is because the returned tasks array is in an arbitrary order? We should be sorting those closer to where the manifest is actually being written, because the order of the tasks in a task producer _shouldn't_ be meaningful. In fact, when the task producer tasks are added to the product plan, they're placed in a set. So an ordering at this level would have no effect.

Tests/SWBTaskConstructionTests/PackageProductConstructionTests.swift
```diff
@@ -1353,4 +1353,50 @@ fileprivate struct PackageProductConstructionTests: CoreBasedTests {
...
+                    ]
+                ),
+            ])
+
+        let core = try await getCore()
+        let tester = try await TaskConstructionTester(core, testProject)
+
+        await tester.checkBuild(runDestination: .macOS) { results in
+            results.checkNoDiagnostics()
+            // The test verifies that the bundles are processed in sorted order
```
jakepetroules, line 1397: The verification stated by the comments doesn't exist though -- is this PR meant to still be a work in progress?

jverkoey:
Still working on testing this one

jakepetroules:
No problem, thanks for moving it to draft to clarify the status!

jverkoey:
You're right — after investigating further, sorting `bundlesToEmbed` here has no effect on task ordering since the tasks end up in a `Set`. The only side effect of the iteration order is the order that `declareGeneratedPrivacyPlistContent()` appends to `_generatedPrivacyContentFilePaths`, but that's already handled at the correct consumption site by #1120 which sorts `appPrivacyContentFiles` in `InfoPlistTool.swift`. Closing this in favor of that fix.

PR #1120, opened 2026-02-21
Sort appPrivacyContentFiles inputs for deterministic build plans

### LINKED ISSUE #1119: Non-deterministic ProcessInfoPlistFile task inputs from privacy content file ordering
## Description

The `ProcessInfoPlistFile` task in `InfoPlistTool.swift` receives privacy content file paths from multiple task producers (`CopyFilesTaskProducer`, `CopySwiftPackageResourcesTaskProducer`, `SourcesTaskProducer`) via `declareGeneratedPrivacyPlistContent`. The order in which these paths arrive depends on task producer execution order, which is non-deterministic.

While the command line arguments for `-scanforprivacyfile` are correctly sorted via `filteredPrivacyLocations()` (which returns `privacyFiles.sorted()`), the task **inputs** array passes `appPrivacyContentFiles` unsorted:

```swift
// InfoPlistTool.swift line 114
let inputs = [inputPath] + effectiveAdditionalContentFilePaths + appPrivacyContentFiles
```

This causes the build manifest to change between consecutive identical builds, preventing `BuildDescription` cache hits and degrading incremental build performance.

## Steps to Reproduce

1. Build an app target that embeds multiple Swift package resource bundles and/or framework dependencies (e.g., 30+ embedded bundles)
2. Build the same project again without any changes
3. Compare the `manifest.json` files in `XCBuildData/*.xcbuilddata/`
4. The `ProcessInfoPlistFile` command will have its `inputs` array in a different order

## Expected Behavior

Consecutive identical builds should produce identical build manifests, enabling `BuildDescription` cache hits.

## Actual Behavior

The `inputs` array for `ProcessInfoPlistFile` has non-deterministic ordering of embedded bundle paths, causing a different manifest hash each build.

## Fix

Sort `appPrivacyContentFiles` before adding to the inputs array:

```swift
let inputs = [inputPath] + effectiveAdditionalContentFilePaths + appPrivacyContentFiles.sorted()
```

This is consistent with how `filteredPrivacyLocations()` already sorts for the command line.

## Environment

- Xcode 26.3
- macOS 26.3
- Project with 30+ embedded Swift package resource bundles

## Related

- #1115 (xcstringstool non-deterministic output ordering)
- #1116 (EMBED_PACKAGE_RESOURCE_BUNDLE_NAMES non-deterministic ordering)

jverkoey:
## Summary

Privacy content file paths (`appPrivacyContentFiles`) are collected from multiple task producers (`CopyFilesTaskProducer`, `CopySwiftPackageResourcesTaskProducer`, `SourcesTaskProducer`) via `declareGeneratedPrivacyPlistContent()`. These producers run in parallel via an aggregation queue, and the paths are stored in a collection on `TaskProducerContext`. The resulting iteration order is non-deterministic, which causes both the `-scanforprivacyfile` command line arguments and the task `inputs` array in the `ProcessInfoPlistFile` task to vary between identical builds. Since command line argument order and task input order both affect build description hashes, this prevents `BuildDescription` cache hits and produces non-deterministic build plans.

## Fix

- Changed `_generatedPrivacyContentFilePaths` in `TaskProducerContext` from `[Path]` to `Set<Path>` (semantically correct — these are unique paths)
- Changed the `appPrivacyContentFiles` parameter on `constructInfoPlistTasks` from `[Path]` to `Set<Path>`
- Sort once at the beginning of `constructInfoPlistTasks` via `let sortedAppPrivacyContentFiles = appPrivacyContentFiles.sorted()`, and use the sorted result in both places where order matters:
  1. The `-scanforprivacyfile` command line arguments (via `filteredPrivacyLocations`)
  2. The task `inputs` array

## Test plan

- Added a test that sets up an app target with `EMBED_PACKAGE_RESOURCE_BUNDLE_NAMES: "zebra_res cherry_res apple_res"` (intentionally unsorted). These flow through `CopySwiftPackageResourcesTaskProducer` → `declareGeneratedPrivacyPlistContent()` → `Set<Path>`, and the test asserts the `ProcessInfoPlistFile` task receives `.bundle` inputs in sorted order.
- Verified the test fails without the `.sorted()` call and passes with it.

Fixes #1119. Part of #1111.

### DIFF
--- a/Sources/SWBCore/SpecImplementations/Tools/InfoPlistTool.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/InfoPlistTool.swift
@@ -22,7 +22,8 @@ public final class InfoPlistToolSpec : GenericCommandLineToolSpec, SpecIdentifie
         fatalError("unexpected direct invocation")
     }
 
-    public func constructInfoPlistTasks(_ cbc: CommandBuildContext, _ delegate: any TaskGenerationDelegate, generatedPkgInfoFile: Path? = nil, additionalContentFilePaths: [Path] = [], requiredArch: String? = nil, appPrivacyContentFiles: [Path] = [], clientLibrariesForCodelessBundle: [String] = []) async {
+    public func constructInfoPlistTasks(_ cbc: CommandBuildContext, _ delegate: any TaskGenerationDelegate, generatedPkgInfoFile: Path? = nil, additionalContentFilePaths: [Path] = [], requiredArch: String? = nil, appPrivacyContentFiles: Set<Path> = [], clientLibrariesForCodelessBundle: [String] = []) async {
+        let sortedAppPrivacyContentFiles = appPrivacyContentFiles.sorted()
         let input = cbc.input
         let inputPath = input.absolutePath
         let outputPath = cbc.output
@@ -82,7 +83,7 @@ public final class InfoPlistToolSpec : GenericCommandLineToolSpec, SpecIdentifie
             commandLine += ["-additionalcontentfile", path.str]
         }
 
-        for path in filteredPrivacyLocations(appPrivacyContentFiles, cbc) {
+        for path in filteredPrivacyLocations(sortedAppPrivacyContentFiles, cbc) {
             commandLine += ["-scanforprivacyfile", path.str]
         }
 
@@ -111,7 +112,7 @@ public final class InfoPlistToolSpec : GenericCommandLineToolSpec, SpecIdentifie
         commandLine += ["-o", outputPath.str]
 
         let context = InfoPlistProcessorTaskActionContext(scope: cbc.scope, productType: cbc.producer.productType, platform: cbc.producer.platform, sdk: cbc.producer.sdk, sdkVariant: cbc.producer.sdkVariant, cleanupRequiredArchitectures: cleanupArchs.sorted(), clientLibrariesForCodelessBundle: clientLibrariesForCodelessBundle)
-        let inputs = [inputPath] + effectiveAdditionalContentFilePaths + appPrivacyContentFiles.sorted()
+        let inputs = [inputPath] + effectiveAdditionalContentFilePaths + sortedAppPrivacyContentFiles
         let serializer = MsgPackSerializer()
         serializer.serialize(context)
         let contextPath = delegate.recordAttachment(contents: serializer.byteString)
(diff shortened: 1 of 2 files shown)


jakepetroules:
Awesome, thanks for the fix!

jakepetroules:
@swift-ci test

jakepetroules (approved):

jverkoey:
Moving this to Draft status until the tests are solid.

jakepetroules:
There's actually a second place earlier in InfoPlistTool.swift where you'll need to sort as well:

```swift
for path in filteredPrivacyLocations(appPrivacyContentFiles, cbc) {
    commandLine += ["-scanforprivacyfile", path.str]
}
```

Like inputs/outputs, the order of command line arguments is significant and will impact build description hashes.

I would suggest changing the `appPrivacyContentFiles` argument to `constructInfoPlistTasks`, and `TaskProducer.generatedPrivacyContentFilePaths`\*, to be a `Set`, and then sort it once near the beginning of the inside of the `constructInfoPlistTasks `method before it's used in the two places where order matters.

\* _There's a ton of similar path lists in `TaskProducerContext`; basically all of those should be `Set`s, actually..._

jverkoey:
Thanks for the feedback! I've made the following changes:

- Changed `_generatedPrivacyContentFilePaths` in `TaskProducerContext` from `[Path]` to `Set<Path>` (and updated the accessor return type and `addPrivacyContentPlistContent` to use `.insert()`)
- Changed the `appPrivacyContentFiles` parameter on `constructInfoPlistTasks` from `[Path]` to `Set<Path>`
- Added `let sortedAppPrivacyContentFiles = appPrivacyContentFiles.sorted()` once at the top of `constructInfoPlistTasks`, and both usage sites (the `-scanforprivacyfile` command line loop and the task `inputs` array) now use that single sorted result

The test verifies this by setting up an app target with `EMBED_PACKAGE_RESOURCE_BUNDLE_NAMES: "zebra_res cherry_res apple_res"` — these flow through `CopySwiftPackageResourcesTaskProducer` into the `Set<Path>`, and the test asserts that the `ProcessInfoPlistFile` task receives the `.bundle` inputs in sorted order. Without the `.sorted()` call the test fails.

owenv (approved):

owenv:
@swift-ci test

jakepetroules:
@swift-ci test

PR #1122, opened 2026-02-21
Normalize BuildDescription signature to exclude volatile PIF GUIDs

jverkoey:
## Summary

`xcodebuild` regenerates all PIF build file GUIDs on every invocation, causing PIF target signatures (content hashes of PIF data) to change even when the actual target structure hasn't changed. This prevents the `BuildDescription` on-disk cache from ever hitting, adding ~20s of unnecessary overhead to every build for large projects.

This PR includes:

1. **A workaround** that replaces the volatile PIF-derived `target.signature` in `BuildDescriptionSignatureComponents` with a stable content-based hash computed from the target's meaningful structural data (file paths resolved from references, build phase types, shell script contents, build file metadata, product type, dependencies, and build configuration names). Also strips the volatile `_subobjects=` suffix from workspace signatures.

2. **A reproduction case** (`repro_cases/nondeterministic-platform-filters/`) — a minimal Swift package with multi-platform conditional dependencies that produces different build description signatures on every `xcodebuild` invocation due to `Set<PlatformFilter>` iteration order varying with per-process hash seed randomization.

See discussion in the review thread for context on why this workaround is needed vs. fixing PIF generation upstream.

## Test plan

Run the reproduction script:
```
cd repro_cases/nondeterministic-platform-filters
./reproduce.sh
```

Expected output (without fix): multiple distinct signatures across runs.

### DIFF
--- a/repro_cases/nondeterministic-platform-filters/Package.swift
+++ b/repro_cases/nondeterministic-platform-filters/Package.swift
@@ -1,0 +1,29 @@
+// swift-tools-version: 6.0
+
+import PackageDescription
+
+let package = Package(
+    name: "PlatformFilterRepro",
+    platforms: [.macOS(.v13), .iOS(.v16), .tvOS(.v16), .watchOS(.v9)],
+    products: [
+        .library(name: "App", targets: ["App"]),
+    ],
+    targets: [
+        .target(
+            name: "App",
+            dependencies: [
+                // Multi-platform conditions produce Set<PlatformFilter> with >1 element.
+                // Set iteration order varies per-process due to hash seed randomization.
+                .target(name: "PlatformSpecific", condition: .when(platforms: [.iOS, .macOS, .tvOS, .watchOS, .visionOS])),
+                .target(name: "Common"),
+            ]
+        ),
+        .target(
+            name: "PlatformSpecific",
+            dependencies: [
+                .target(name: "Common", condition: .when(platforms: [.iOS, .macOS, .tvOS])),
+            ]
+        ),
+        .target(name: "Common"),
+    ]
+)
--- a/repro_cases/nondeterministic-platform-filters/Sources/App/App.swift
+++ b/repro_cases/nondeterministic-platform-filters/Sources/App/App.swift
@@ -1,0 +1,1 @@
+public enum App {}
--- a/repro_cases/nondeterministic-platform-filters/Sources/Common/Common.swift
+++ b/repro_cases/nondeterministic-platform-filters/Sources/Common/Common.swift
@@ -1,0 +1,1 @@
+public enum Common {}
--- a/repro_cases/nondeterministic-platform-filters/Sources/PlatformSpecific/PlatformSpecific.swift
+++ b/repro_cases/nondeterministic-platform-filters/Sources/PlatformSpecific/PlatformSpecific.swift
@@ -1,0 +1,1 @@
+public enum PlatformSpecific {}
--- a/repro_cases/nondeterministic-platform-filters/reproduce.sh
+++ b/repro_cases/nondeterministic-platform-filters/reproduce.sh
@@ -1,0 +1,72 @@
+#!/bin/bash
+#
+# Reproduces nondeterministic PIF platform filter ordering.
+#
+# Uses xcodebuild to build a Swift package with multi-platform conditional
+# dependencies. Each xcodebuild invocation gets a fresh process (and thus
+# a fresh Swift hash seed), causing Set<PlatformFilter> to iterate in a
+# different order during PIF serialization. This produces different build
+# description signatures, preventing the on-disk cache from hitting.
+#
+# Usage:
+#   cd repro_cases/nondeterministic-platform-filters
+#   ./reproduce.sh
+#
+# Expected output (without fix):
+#   Multiple distinct build description signatures across runs.
+#
+# Expected output (with fix):
+#   Same signature every run.
+
+set -euo pipefail
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+cd "$SCRIPT_DIR"
+
+RUNS=5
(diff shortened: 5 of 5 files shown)


owenv (changes requested):
Thanks for the PR. I think that instead of adjusting the signature computation here, we should instead look at eliminating nondeterminism when generating the package PIF. `stableTargetSignature` is currently not sound - it excludes important fields like `customTasks` which must trigger generation of a new build description on changes. We could fix those individual issues, but it's difficult to maintain correctness over time as the project model evolves. I think it'll be easier to test that PIF generation is fully deterministic and maintain that property over the long term.

I was able to reproduce the nondeterministic platform filter ordering in generated PIF, and I think we should sort that when serializing. I wasn't able to reproduce nondeterminism in build file GUIDs when applied to e.g. the swift-build package itself. Do you have an example of an open source package which reproduces this?

jverkoey:
I've reworked the PR to just sort platform filters at the PIF serialization sites, as you suggested. The `stableTargetSignature` approach is gone.

Here's a reproduction case using a minimal Swift package with multi-platform conditional dependencies (`repro_cases/nondeterministic-platform-filters/` in this branch):

```
$ ./reproduce.sh
Building 5 times with xcodebuild to check for nondeterministic PIF signatures...

  Run 1: b694a5fa142bd03dcfc9b03e6b206c97
  Run 2: b72d39dd2283d0166d2fe2162452ee51
  Run 3: f55686b66bde0f4091eabb99a4d52aac
  Run 4: ad000974340a70116337813bc25bbad8
  Run 5: 88ac0fede6771c41aad57e089d562240

REPRODUCED: 5 distinct signatures across 5 runs.
```

5 consecutive `xcodebuild build` invocations of the same unchanged package, 5 different build description signatures. Each invocation gets a fresh Swift hash seed, so `Set<PlatformFilter>` iterates in a different order during PIF serialization.

Regarding build file GUIDs — that issue was specific to a large private Xcode project and I haven't been able to reproduce it with a SwiftPM package. It may be an xcodebuild-specific issue outside swift-build's control. I've scoped this PR down to just the platform filter fix.

jverkoey:
Agreed that this PR was meant more as a workaround; it's probably best if we close it rather than pursue it as-is. That being said, we've added a repro case that is hopefully helpful for fixing this issue upstream.

The repro is at `repro_cases/nondeterministic-platform-filters/` on this branch. It's a minimal Swift package with multi-platform conditional dependencies. Running `./reproduce.sh` shows the nondeterminism:

```
$ ./reproduce.sh
Building 5 times with xcodebuild to check for nondeterministic PIF signatures...

  Run 1: b694a5fa142bd03dcfc9b03e6b206c97
  Run 2: b72d39dd2283d0166d2fe2162452ee51
  Run 3: f55686b66bde0f4091eabb99a4d52aac
  Run 4: ad000974340a70116337813bc25bbad8
  Run 5: 88ac0fede6771c41aad57e089d562240

REPRODUCED: 5 distinct signatures across 5 runs.
```

We believe the root cause is that PIF generation happens in xcodebuild, not the build service. xcodebuild serializes the PIF, computes a content hash to produce `target.signature`, and sends both to the build service. The build service receives that signature and passes it through as-is into `BuildDescriptionSignatureComponents`:

https://github.com/swiftlang/swift-build/blob/a428fd0865954c580229cb6189529b82529ff179/Sources/SWBTaskExecution/BuildDescriptionSignature.swift#L80-L88

Since the PIF content that xcodebuild hashes includes nondeterministic `Set` iteration order (due to Swift's per-process hash seed randomization), the signature changes on every invocation even though the project hasn't changed, and the on-disk cache never hits.

The workaround in this PR side-steps the problem by recomputing `target.signature` on the build service side from the deserialized PIF objects, resolving volatile GUIDs back to stable file paths. It's not sound (as you pointed out, it misses fields like `customTasks`), but it does work in practice via `XCBBUILDSERVICE_PATH` because it runs after the nondeterministic PIF has already been parsed into stable in-memory structures.

Closing this out — hopefully the repro case is useful for tracking down the upstream fix.

jverkoey:
## Learning: in-memory caches must NOT normalize workspace signatures

While working on a separate TargetBuildGraph cache (to avoid recomputing the dependency graph on every build), we applied the same `_subobjects=` normalization from this PR to the cache key. This caused a subtle correctness bug:

**Root cause**: `Target` uses reference identity (`ObjectIdentifier(self)`) for both `hash(into:)` and `==`:

```swift
// Target.swift
public func hash(into hasher: inout Hasher) {
    hasher.combine(ObjectIdentifier(self))
}
public static func ==(lhs: Target, rhs: Target) -> Bool {
    return lhs === rhs
}
```

When the PIF is re-transferred, Xcode creates entirely new `Target` objects. If an in-memory cache normalizes away the `_subobjects=` suffix (making old and new PIF signatures identical), the cache serves a `TargetBuildGraph` containing stale `Target` references from the previous PIF. These stale references:

1. Won't match any key in the new workspace's `projectsByTarget` dictionary (falls back to GUID lookup, which may or may not work depending on context)
2. Won't match `ConfiguredTarget` entries in downstream dictionaries that were built from the new PIF
3. Can cause silent dependency graph corruption — targets may appear to have no dependencies

**Symptom**: intermittent "Unable to find module dependency" errors for targets that are clearly present in the project.

**Fix**: In-memory caches that store live `Target` references must use the **full, un-normalized** workspace signature. The `_subobjects=` normalization is correct for the on-disk `BuildDescription` cache (which serializes/deserializes data without retaining object references) but wrong for any cache that holds live Swift objects using reference identity.

This distinction matters for anyone implementing caching in the build service:
- **On-disk caches** (like `BuildDescriptionManager`): safe to normalize, since data is reconstructed from serialized form
- **In-memory caches** (like TargetBuildGraph topology): must NOT normalize, since cached objects use reference identity

PR #1123, opened 2026-02-22
Make parsing generated file paths from tools more robust

jakepetroules:
Output of invoked tools is expected to always be an absolute path. Output from tools in untrusted, so validate it to ensure we don't fail preconditions and potentially crash on malformed output.

rdar://170834370


jakepetroules:
@swift-ci test

PR #1124, opened 2026-02-23
Cache computed target build graphs across repeated builds

jverkoey:
## Problem

`TargetBuildGraph.init` calls `computeGraph()` on every build, even when nothing in
the workspace has changed. For large workspaces this is expensive — in our monorepo
(2100+ targets) Instruments traces showed `computeGraph()` taking **7–37 seconds per
call**, with 83% of the time spent in the recursive `addDependencies` traversal.

Xcode also issues **multiple graph requests per build action** with different
parameters (dependency graph vs actual build, index preparation vs normal build),
compounding the cost. In our workspace each no-change build was spending **~42 seconds**
in `computeGraph()` across all requests.

## Solution

Add a process-level multi-entry cache (`TargetBuildGraphCache`) that stores computed
graph topologies keyed by a signature of:

- Normalized PIF workspace signature (stripping volatile subobject GUIDs)
- Build parameters and per-target parameter overrides
- Topology-affecting flags (`useImplicitDependencies`, `useParallelTargets`,
  `skipDependencies`, `dependencyScope`, `buildCommand`)
- Graph purpose (`.build` vs `.dependencyGraph`)

The cache is static (process-level) because `WorkspaceContext` is recreated on every
PIF transfer, even when nothing has changed. Multiple entries (up to 8) are stored so
that Xcode's different request types don't evict each other.

On cache hit, unapproved target diagnostics are re-emitted to preserve correctness.

## Measurements

Tested on a 2100+ target monorepo (5159 configured targets in the full graph).
Each no-change build triggers two graph requests:

| Request type | Targets | Without cache | With cache |
|---|---|---|---|
| Dependency graph | 1,654 | 6.4–7.6s | **0s (cache hit)** |
| Build / index prep | 5,159 | 34.6–37.5s | **0s (cache hit)** |
| **Total per build** | | **~42s** | **~0s** |

The first build after launch populates the cache (cold miss); every subsequent
no-change build hits. The cache is automatically invalidated when the PIF workspace
signature changes (i.e. when the project structure actually changes).

### DIFF
--- a/Sources/SWBCore/TargetBuildGraphCache.swift
+++ b/Sources/SWBCore/TargetBuildGraphCache.swift
@@ -10,7 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-import SWBUtil
+package import SWBUtil
 
 /// Process-level cache for computed target dependency graphs.
 ///
@@ -24,19 +24,45 @@ public enum TargetBuildGraphCache {
     /// The data we cache — everything needed by the
     /// `TargetBuildGraph` memberwise init except the live context
     /// objects (workspaceContext, buildRequest, buildRequestContext).
-    struct CachedDependencyGraph: @unchecked Sendable {
-        let allTargets: OrderedSet<ConfiguredTarget>
-        let targetDependencies:
+    package struct CachedDependencyGraph: @unchecked Sendable {
+        /// Signature without workspaceIdentity — used for
+        /// content-based remap matching.
+        package let contentSignature: Int
+        package let allTargets: OrderedSet<ConfiguredTarget>
+        package let targetDependencies:
             [ConfiguredTarget: [ResolvedTargetDependency]]
-        let targetsToLinkedReferencesToProducingTargets:
+        package let targetsToLinkedReferencesToProducingTargets:
             [ConfiguredTarget:
                 [BuildFile.BuildableItem: ResolvedTargetDependency]]
-        let dynamicallyBuildingTargets: Set<Target>
+        package let dynamicallyBuildingTargets: Set<Target>
         /// All non-error diagnostics emitted during resolution,
         /// re-emitted on cache hit.
-        let diagnostics: [Diagnostic]
+        package let diagnostics: [Diagnostic]
         /// Last access time for LRU eviction.
-        var lastAccess: UInt64
+        package var lastAccess: UInt64
+
+        package init(
+            contentSignature: Int,
+            allTargets: OrderedSet<ConfiguredTarget>,
+            targetDependencies: [ConfiguredTarget:
+                [ResolvedTargetDependency]],
+            targetsToLinkedReferencesToProducingTargets:
+                [ConfiguredTarget:
+                    [BuildFile.BuildableItem:
+                        ResolvedTargetDependency]],
+            dynamicallyBuildingTargets: Set<Target>,
+            diagnostics: [Diagnostic],
+            lastAccess: UInt64
+        ) {
+            self.contentSignature = contentSignature
+            self.allTargets = allTargets
+            self.targetDependencies = targetDependencies
+            self.targetsToLinkedReferencesToProducingTargets =
+                targetsToLinkedReferencesToProducingTargets
+            self.dynamicallyBuildingTargets = dynamicallyBuildingTargets
+            self.diagnostics = diagnostics
+            self.lastAccess = lastAccess
+        }
     }
 
     /// Maximum number of cached entries. Xcode typically issues 2-4
@@ -87,6 +113,118 @@ public enum TargetBuildGraphCache {
     /// the memory overhead of caching them is not worth it.
     static func shouldSkipCache(buildCommand: BuildCommand) -> Bool {
         buildCommand.isPrepareForIndexing
+    }
+
+    /// Scan all cached entries for a content signature match.
+    ///
+    /// Unlike `lookup(signature:)` which does an exact key lookup,
+    /// this scans all entries because the full signature (which
+    /// includes workspaceIdentity) will be different even when the
+    /// content is identical.
+    static func lookupByContentSignature(
+        _ contentSig: Int
+    ) -> CachedDependencyGraph? {
+        entries.withLock { entries in
+            for (_, entry) in entries {
+                if entry.contentSignature == contentSig {
+                    return entry
+                }
+            }
+            return nil
+        }
+    }
+
+    /// Remap all Target references in a cached graph to point at
+    /// fresh objects from the new workspace.
+    ///
+    /// Returns nil if any cached Target GUID is not found in the new
+    /// workspace, which means the PIF structure changed and a full
+    /// rebuild is needed.
+    package static func remapGraph(
+        _ cached: CachedDependencyGraph,
+        to workspace: Workspace
+    ) -> CachedDependencyGraph? {
+        // Build GUID → new Target lookup.
+        // If ANY target GUID is missing, bail out.
+        var guidToNewTarget: [String: Target] = [:]
+        for ct in cached.allTargets {
+            let guid = ct.target.guid
+            guard let newTarget = workspace.target(for: guid) else {
+                return nil
+            }
+            guidToNewTarget[guid] = newTarget
+        }
+        for target in cached.dynamicallyBuildingTargets {
+            let guid = target.guid
+            guard let newTarget = workspace.target(for: guid) else {
+                return nil
+            }
+            guidToNewTarget[guid] = newTarget
+        }
+
+        // Remap allTargets
+        var newAllTargets = OrderedSet<ConfiguredTarget>()
+        for ct in cached.allTargets {
+            newAllTargets.append(
+                ct.replacingTarget(guidToNewTarget[ct.target.guid]!)
+            )
+        }
+
+        // Remap targetDependencies
+        var newDeps = [ConfiguredTarget:
+            [ResolvedTargetDependency]](
+            minimumCapacity: cached.targetDependencies.count
+        )
+        for (ct, deps) in cached.targetDependencies {
+            let newCT = ct.replacingTarget(
+                guidToNewTarget[ct.target.guid]!
+            )
+            newDeps[newCT] = deps.map { dep in
+                dep.replacingTarget(
+                    guidToNewTarget[dep.target.target.guid]!
+                )
+            }
+        }
+
+        // Remap targetsToLinkedReferencesToProducingTargets
+        var newLinked = [ConfiguredTarget:
+            [BuildFile.BuildableItem: ResolvedTargetDependency]](
+            minimumCapacity: cached
+                .targetsToLinkedReferencesToProducingTargets.count
+        )
+        for (ct, innerMap) in cached
+            .targetsToLinkedReferencesToProducingTargets
+        {
+            let newCT = ct.replacingTarget(
+                guidToNewTarget[ct.target.guid]!
+            )
+            var newInner = [BuildFile.BuildableItem:
+                ResolvedTargetDependency](
+                minimumCapacity: innerMap.count
+            )
+            for (buildableItem, dep) in innerMap {
+                newInner[buildableItem] = dep.replacingTarget(
+                    guidToNewTarget[dep.target.target.guid]!
+                )
+            }
+            newLinked[newCT] = newInner
+        }
+
+        // Remap dynamicallyBuildingTargets
+        let newDynamic = Set(
+            cached.dynamicallyBuildingTargets
+                .map { guidToNewTarget[$0.guid]! }
+        )
+
+        return CachedDependencyGraph(
+            contentSignature: cached.contentSignature,
+            allTargets: newAllTargets,
+            targetDependencies: newDeps,
+            targetsToLinkedReferencesToProducingTargets: newLinked,
+            dynamicallyBuildingTargets: newDynamic,
+            diagnostics: cached.diagnostics,
+            lastAccess: nextAccessTime()
+        )
     }
 
     /// Compute a cache signature from the inputs that determine the
@@ -161,6 +299,51 @@ public enum TargetBuildGraphCache {
         // Purpose affects diagnostic behavior
         hasher.combine(purpose)
 
+        return hasher.finalize()
+    }
+
+    /// Compute a content-based cache signature from the same inputs
+    /// as `computeSignature`, but **excluding** `workspaceIdentity`.
+    ///
+    /// When the PIF is re-transferred after a source-only change,
+    /// the workspace object identity changes but the content
+    /// (workspace signature, build parameters, target GUIDs) stays
+    /// the same. A content signature match lets us remap the cached
+    /// graph's Target references instead of recomputing the entire
+    /// dependency graph.
+    static func computeContentSignature(
+        workspaceSignature: String,
+        buildRequest: BuildRequest,
+        purpose: TargetBuildGraph.Purpose
+    ) -> Int {
+        var hasher = Hasher()
+        hasher.combine(workspaceSignature)
+        hasher.combine(buildRequest.parameters)
+        for targetInfo in buildRequest.buildTargets {
+            hasher.combine(targetInfo.target.guid)
+            hasher.combine(targetInfo.parameters)
+        }
+        hasher.combine(buildRequest.useImplicitDependencies)
+        hasher.combine(buildRequest.useParallelTargets)
+        hasher.combine(buildRequest.skipDependencies)
+        hasher.combine(buildRequest.dependencyScope)
+        switch buildRequest.buildCommand {
+        case .build:
+            hasher.combine("build")
+        case .generateAssemblyCode:
+            hasher.combine("asm")
+        case .generatePreprocessedFile:
+            hasher.combine("preprocess")
+        case .singleFileBuild:
+            hasher.combine("single")
+        case .prepareForIndexing:
+            hasher.combine("index")
+        case .cleanBuildFolder:
+            hasher.combine("clean")
+        case .preview:
+            hasher.combine("preview")
+        }
+        hasher.combine(purpose)
         return hasher.finalize()
     }
 }
--- a/Sources/SWBCore/TargetDependencyResolver.swift
+++ b/Sources/SWBCore/TargetDependencyResolver.swift
@@ -123,6 +123,11 @@ public struct TargetBuildGraph: TargetGraph, Sendable {
             buildRequest: buildRequest,
             purpose: purpose
         )
+        let contentSignature = TargetBuildGraphCache.computeContentSignature(
+            workspaceSignature: workspaceContext.workspace.signature,
+            buildRequest: buildRequest,
+            purpose: purpose
+        )
 
         let allTargets: OrderedSet<ConfiguredTarget>
         let targetDependencies: [ConfiguredTarget: [ResolvedTargetDependency]]
@@ -138,6 +143,23 @@ public struct TargetBuildGraph: TargetGraph, Sendable {
             // Re-emit all cached diagnostics on cache hit so callers observe
             // the same diagnostic behavior regardless of cache state.
             for diagnostic in cached.diagnostics {
+                delegate.emit(diagnostic)
+            }
+        } else if !skipCache,
+                  let contentMatch = TargetBuildGraphCache.lookupByContentSignature(contentSignature),
+                  let remapped = TargetBuildGraphCache.remapGraph(contentMatch, to: workspaceContext.workspace) {
+            // Content-identical PIF re-transfer — remap Target references
+            // instead of recomputing the full dependency graph.
+            allTargets = remapped.allTargets
+            targetDependencies = remapped.targetDependencies
+            targetsToLinkedReferencesToProducingTargets = remapped.targetsToLinkedReferencesToProducingTargets
+            dynamicallyBuildingTargets = remapped.dynamicallyBuildingTargets
+
+            // Store under the new full signature for future exact hits.
+            TargetBuildGraphCache.store(signature: signature, graph: remapped)
+
+            // Re-emit diagnostics.
+            for diagnostic in remapped.diagnostics {
                 delegate.emit(diagnostic)
             }
         } else {
@@ -156,6 +178,7 @@ public struct TargetBuildGraph: TargetGraph, Sendable {
                 TargetBuildGraphCache.store(
                     signature: signature,
                     graph: .init(
+                        contentSignature: contentSignature,
                         allTargets: allTargets,
                         targetDependencies: targetDependencies,
                         targetsToLinkedReferencesToProducingTargets: targetsToLinkedReferencesToProducingTargets,
--- a/Tests/SWBCoreTests/TargetBuildGraphCacheTests.swift
+++ b/Tests/SWBCoreTests/TargetBuildGraphCacheTests.swift
@@ -1,0 +1,722 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Testing
+@_spi(Testing) import SWBCore
+import SWBProtocol
+import SWBTestSupport
+@_spi(Testing) import SWBUtil
+import Foundation
+
+@Suite fileprivate struct TargetBuildGraphCacheTests: CoreBasedTests {
+
+    /// Verify that two workspace signatures differing only in
+    /// _subobjects= suffix produce DIFFERENT cache signatures.
+    @Test(.requireSDKs(.host))
+    func workspaceSignatureIncludesSubobjects() async throws {
+        let core = try await getCore()
+
+        let workspace = try TestWorkspace(
+            "Workspace",
+            projects: [
+                TestProject(
+                    "P1",
+                    groupTree: TestGroup(
+                        "G1",
+                        children: [TestFile("S1.c")]
+                    ),
+                    buildConfigurations: [
+                        TestBuildConfiguration(
+                            "Debug", buildSettings: [:]
+                        ),
+                    ],
+                    targets: [
+                        TestStandardTarget(
+                            "T1",
+                            type: .framework,
+                            buildConfigurations: [
+                                TestBuildConfiguration(
+                                    "Debug",
+                                    buildSettings: [
+                                        "PRODUCT_NAME":
+                                            "$(TARGET_NAME)"
+                                    ]
+                                ),
+                            ],
+                            buildPhases: [
+                                TestSourcesBuildPhase(["S1.c"])
+                            ]
+                        ),
+                    ]
+                ),
+            ]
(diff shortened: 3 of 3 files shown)


Sources/SWBCore/TargetBuildGraphCache.swift
```diff
@@ -0,0 +1,144 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import SWBUtil
+import Foundation
+
+/// Process-level cache for computed target build graphs.
+///
+/// Xcode issues multiple `TargetBuildGraph` requests per build action
+/// with different parameters (e.g. dependency graph vs actual build,
```
owenv: This comment is incorrect - index preparation and user-initiated builds are separate build actions and it's not in general safe to assume particular instances of each are related to each other. The SwiftPM CLI/BSP will also issue requests differently

Sources/SWBCore/TargetBuildGraphCache.swift
```diff
@@ -0,0 +1,144 @@
...
+/// index preparation vs normal build). A multi-entry cache ensures
+/// these don't evict each other.
+///
+/// The cache is static (process-level) because `WorkspaceContext` is
```
owenv, line 21: All caches should be at least Session-scoped because in the context of an IDE/BSP/etc. the build service may regularly open and close unrelated sessions. Instead of working around the WorkspaceContext recreation, we should look at being less aggressive about tearing it down unnecessarily if we're going to do more caching at this layer.

Sources/SWBCore/TargetBuildGraphCache.swift
```diff
@@ -0,0 +1,144 @@
...
+/// recreated on every PIF transfer, even when nothing has changed.
+public enum TargetBuildGraphCache {
+    /// The data we cache — everything needed by the
+    /// `TargetBuildGraph` memberwise init except the live context
+    /// objects (workspaceContext, buildRequest, buildRequestContext).
```
owenv, line 26: This comment is incorrectly referring to the build request as context, we should consider it a primary input to the computation

Sources/SWBCore/TargetBuildGraphCache.swift
```diff
@@ -0,0 +1,144 @@
...
+    struct CachedTopology: @unchecked Sendable {
```
owenv: Why name this "topology"? It doesn't really fit the traditional definition

Sources/SWBCore/TargetBuildGraphCache.swift
```diff
@@ -0,0 +1,144 @@
...
+            [ConfiguredTarget: [ResolvedTargetDependency]]
+        let targetsToLinkedReferencesToProducingTargets:
+            [ConfiguredTarget:
+                [BuildFile.BuildableItem: ResolvedTargetDependency]]
+        let dynamicallyBuildingTargets: Set<Target>
+    }
+
+    /// Maximum number of cached entries. Xcode typically issues 2-4
+    /// distinct graph requests per build action; 8 gives headroom.
+    private static let maxEntries = 8
```
owenv, line 70: Rather than attempting to size the cache to take anticipate index preparation patterns, I recommend a smaller cache which takes the build request QoS into account. Index preparation requests are intentionally submitted with lower QoS and the added memory overhead of caching their dependency graphs is likely not worth it

Sources/SWBCore/TargetBuildGraphCache.swift
```diff
@@ -0,0 +1,144 @@
...
+            entries[signature]
+        }
+    }
+
+    /// Store a computed topology for the given signature.
+    static func store(signature: Int, topology: CachedTopology) {
+        _entries.withLock { entries in
+            // Evict all entries if we exceed the limit (simple reset
+            // policy). This only happens when the PIF changes or the
+            // user switches between different build configurations.
```
owenv: This comment is currently incorrect since the cache is process-wide and there may be multiple active sessions. That said, the proposed eviction policy doesn't make much sense to me. Why not use an LRU eviction policy instead? This is already implemented in e.g. HeavyCache in SWBUtil

Sources/SWBCore/TargetBuildGraphCache.swift
```diff
@@ -0,0 +1,144 @@
...
+    /// on each other — only the PIF does.
+    static func computeSignature(
+        workspaceSignature: String,
+        buildRequest: BuildRequest,
+        purpose: TargetBuildGraph.Purpose
+    ) -> Int {
+        var hasher = Hasher()
+
+        // Normalized PIF signature (strip volatile subobject GUIDs)
+        if let range = workspaceSignature.range(
```
owenv: We shouldn't make assumptions about the formatting of the signature string. Like I said in the other PR, we should eliminate nondeterminism when serializing instead.

Sources/SWBCore/TargetBuildGraphCache.swift
```diff
@@ -0,0 +1,144 @@
...
+        } else {
+            hasher.combine(workspaceSignature)
+        }
+
+        // Global build parameters
+        hasher.combine(buildRequest.parameters)
+
+        // Top-level build targets and their per-target parameters.
+        // Sort by target GUID for order independence.
+        for targetInfo in buildRequest.buildTargets.sorted(
```
owenv: This shouldn't be sorted, the order targets appear in the build request is meaningful when using certain options like deprecated manual build ordering

Sources/SWBCore/TargetBuildGraphCache.swift
```diff
@@ -0,0 +1,144 @@
...
+            hasher.combine(targetInfo.parameters)
+        }
+
+        // Flags that affect graph topology
+        hasher.combine(buildRequest.useImplicitDependencies)
+        hasher.combine(buildRequest.useParallelTargets)
+        hasher.combine(buildRequest.skipDependencies)
+
+        // Dependency scope affects pruning
+        switch buildRequest.dependencyScope {
```
owenv: This should be made Hashable instead of implementing it inline

Sources/SWBCore/TargetBuildGraphCache.swift
```diff
@@ -0,0 +1,144 @@
...
+        case .workspace:
+            hasher.combine(0)
+        case .buildRequest:
+            hasher.combine(1)
+        }
+
+        // Build command affects the early-return for
+        // assembly/preprocessor
+        switch buildRequest.buildCommand {
```
owenv, line 282: This should be made Hashable instead of implementing it inline

Sources/SWBCore/TargetBuildGraphCache.swift
```diff
@@ -0,0 +1,144 @@
...
+        case .prepareForIndexing:
+            hasher.combine("index")
+        case .cleanBuildFolder:
+            hasher.combine("clean")
+        case .preview:
+            hasher.combine("preview")
+        }
+
+        // Purpose affects diagnostic behavior
+        switch purpose {
```
owenv: This should be made Hashable instead of implementing it inline also

Sources/SWBCore/TargetBuildGraphCache.swift
```diff
@@ -0,0 +1,144 @@
...
+        case .build:
+            hasher.combine("build")
+        case .dependencyGraph:
+            hasher.combine("depgraph")
+        }
+
```
owenv, line 301: This cache key computation is missing at least a few important components, like the signature of xcconfig files loaded while computing settings in the dependency resolver, user preferences, etc.

I think I'd recomment first refactoring the dependency resolver to eliminate direct access to the buildRequestContext and workspaceContext in favor of finer-grained inputs. This should make it easier to audit the inputs they're indirectly introducing today and ensure the cache key is complete.

Sources/SWBCore/TargetDependencyResolver.swift
```diff
@@ -109,13 +109,53 @@ public struct TargetBuildGraph: TargetGraph, Sendable {
...
+        let targetsToLinkedReferencesToProducingTargets: [ConfiguredTarget: [BuildFile.BuildableItem: ResolvedTargetDependency]]
+        let dynamicallyBuildingTargets: Set<Target>
+
+        if let cached = TargetBuildGraphCache.lookup(signature: signature) {
+            allTargets = cached.allTargets
+            targetDependencies = cached.targetDependencies
+            targetsToLinkedReferencesToProducingTargets = cached.targetsToLinkedReferencesToProducingTargets
+            dynamicallyBuildingTargets = cached.dynamicallyBuildingTargets
+
+            // Re-emit unapproved target diagnostics on cache hit
```
owenv: We should cache all diagnostics emitted during dependency resolution instead of just the approval ones. IMO we should also not cache any dependency graph where the resolver emitted errors, for better recovery in the event of a nondeterministic bug in resolution.

Sources/SWBCore/TargetBuildGraphCache.swift
```diff
@@ -0,0 +1,144 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
```
owenv: nit: copyright header

Sources/SWBCore/TargetBuildGraphCache.swift
```diff
@@ -0,0 +1,144 @@
...
+            [ConfiguredTarget:
+                [BuildFile.BuildableItem: ResolvedTargetDependency]]
+        let dynamicallyBuildingTargets: Set<Target>
+    }
+
+    /// Maximum number of cached entries. Xcode typically issues 2-4
+    /// distinct graph requests per build action; 8 gives headroom.
+    private static let maxEntries = 8
+
+    private static let _entries =
```
owenv: nit: I don't think this needs the leading underscore

Sources/SWBCore/TargetDependencyResolver.swift
```diff
@@ -109,13 +109,53 @@ public struct TargetBuildGraph: TargetGraph, Sendable {
     ///
     /// The result closure guarantees that all targets a target depends on appear in the returned array before that target.  Any detected dependency cycles will be broken.
     public init(workspaceContext: WorkspaceContext, buildRequest: BuildRequest, buildRequestContext: BuildRequestContext, delegate: any TargetDependencyResolverDelegate, purpose: Purpose = .build) async {
```
owenv: nit: This should maybe change from an initializer to a static method that indicates it caches results internally

jverkoey:
Thank you for the feedback! While I'm working on the feedback I shared https://github.com/swiftlang/swift-build/issues/1111#issuecomment-3943123110 which includes a few Instruments runs that might be of use for understanding the dragon I'm trying to slay here..

Sources/SWBCore/TargetBuildGraphCache.swift
jverkoey: Agreed. The cached `Target` objects use reference identity, so serving them across sessions with different `WorkspaceContext`s would be incorrect. In practice cross-session signature collisions are unlikely (PIF signatures include all target GUIDs and project paths), but it's not safe in general.

Deferred for now since it requires threading a cache instance through the call chain (`PlanningOperation`, `DependencyGraphMessages`, `CleanOperation`, etc). `Session` seems like the right owner since it already manages `BuildDescriptionManager`. Open to guidance on where you'd want this to live.

Sources/SWBCore/TargetBuildGraphCache.swift
jverkoey: Agreed, the resolver reads inputs through context objects that aren't reflected in the signature. Deferred since the fix requires refactoring the resolver to declare explicit inputs. Low practical risk within a single Xcode session but not correct in general.

Sources/SWBCore/TargetBuildGraphCache.swift
jverkoey: Renamed to `CachedDependencyGraph`.

Sources/SWBCore/TargetBuildGraphCache.swift
jverkoey: Done. `shouldSkipCache(buildCommand:)` skips both lookup and store for `prepareForIndexing`. Cache size reduced to 8 entries.

Sources/SWBCore/TargetBuildGraphCache.swift
jverkoey: Done. Tracks `lastAccess` per entry, evicts the oldest when `maxEntries` is exceeded.

Sources/SWBCore/TargetBuildGraphCache.swift
jverkoey: Done. Uses `workspaceSignature` as-is. The in-memory cache stores live `Target` references with identity semantics, so it must miss on any PIF change including subobject GUIDs.

Sources/SWBCore/TargetBuildGraphCache.swift
jverkoey: Done. `computeSignature` now preserves input order.

Sources/SWBCore/TargetBuildGraphCache.swift
jverkoey: Done. `DependencyScope` now conforms to `Hashable`.

Sources/SWBCore/TargetBuildGraphCache.swift
jverkoey: Done. `Purpose` now conforms to `Hashable`.

Sources/SWBCore/TargetBuildGraphCache.swift
jverkoey: `BuildCommand` is kept as a switch since its associated values (`[Path]`, `[Target]?`) prevent auto-synthesized `Hashable`, and only the discriminator affects graph topology.

Sources/SWBCore/TargetBuildGraphCache.swift
jverkoey: Agreed, the resolver reads inputs through context objects that aren't reflected in the signature. Deferred since the fix requires refactoring the resolver to declare explicit inputs. Low practical risk within a single Xcode session but not correct in general.

Sources/SWBCore/TargetDependencyResolver.swift
jverkoey: Done. Added `DiagnosticCollectingDelegate` that wraps the inner delegate and collects all diagnostics. Stored in `CachedDependencyGraph.diagnostics`, re-emitted on hit. Graphs with errors are not cached.

Sources/SWBCore/TargetBuildGraphCache.swift
jverkoey: Fixed. Updated new files (`TargetBuildGraphCache.swift`, `DiagnosticCollectingDelegate.swift`) to 2026.

Sources/SWBCore/TargetBuildGraphCache.swift
jverkoey: Done. `_entries` → `entries`, `_buildCounter` → `accessCounter`.

Sources/SWBCore/TargetDependencyResolver.swift
jverkoey: Done. Now `TargetBuildGraph.cached(...)`. All call sites updated. Memberwise `init` unchanged.

jverkoey:
All feedback addressed in the latest force-push. Summary:

**Implemented:**
- LRU eviction (tracks `lastAccess` per entry)
- QoS-aware caching (skip `prepareForIndexing` for both lookup and store)
- Preserve build target order (removed `.sorted(by:)`)
- `DependencyScope` and `Purpose` conform to `Hashable`
- `DiagnosticCollectingDelegate` collects all diagnostics during resolution, re-emits on cache hit, skips caching graphs with errors
- Use `workspaceSignature` as-is (no normalization)
- Renamed `CachedTopology` → `CachedDependencyGraph`
- Removed leading underscores (`entries`, `accessCounter`)
- Caching `init` → `TargetBuildGraph.cached(...)` static factory
- Copyright year updated to 2026 on new files

**Remaining work (deferred):**
1. **Session-scoped cache**: The cache is still process-wide (`static`). Scoping to `Session` requires threading a cache instance through the call chain. `Session` seems like the right owner since it already manages `BuildDescriptionManager`. Open to guidance on the preferred approach.
2. **Complete cache key**: The signature doesn't capture xcconfig file contents or user preferences accessed indirectly through context objects. Fixing this requires refactoring the resolver to declare explicit inputs rather than reaching into `buildRequestContext`/`workspaceContext`. Low practical risk within a single session, but not correct in general.

Happy to tackle either of these in a follow-up or in this PR if you'd prefer.

Sources/SWBCore/TargetBuildGraphCache.swift
jverkoey: Fixed. Updated the doc comment to not assume index preparation and user-initiated builds are related. Also changed "Xcode" to "Clients" since SwiftPM CLI/BSP issue requests differently.

jverkoey:
Swift Build performance metrics comparisons between my mainline build and this PR:

Screenshot 2026-02-26 at 12 55 03 AM

Note that the number of builds is relatively small because I just got this infra up and running. Hoping to have a bigger dataset over time but I can confirm on local builds that I'm getting significantly faster builds with this change.

jverkoey:
Moving this to a draft state because I discovered a critical bug in this implementation where changing a file doesn't result in a new build being kicked off.

jverkoey:
Pushed a fix for a cache invalidation bug I discovered during testing: [`91afdfa`](https://github.com/swiftlang/swift-build/pull/1124/commits/91afdfa8b31db1cc18d67ab12b986cfa8ab0779e)

**Bug**: The cached dependency graph stores live `ConfiguredTarget` objects whose `Target` references use reference identity (`ObjectIdentifier`) for `hash(into:)` / `==`. When the PIF is re-transferred, `IncrementalPIFLoader` creates new `Target` objects even when the content is unchanged. The workspace signature (content-based) stays the same → cache hit → stale `Target` references → dictionary lookups in `targetDependencies` and `provisioningInputs` silently fail → build skips recompilation of changed source files.

**Fix**: Include `ObjectIdentifier(workspaceContext.workspace)` in the cache signature. Same `Workspace` instance (reused by the incremental loader within a session) → cache hit. New `Workspace` instance (PIF re-transferred) → cache miss. This is correct because `Target` references are tied to the `Workspace` instance lifetime.

This also partially addresses the session-scoping feedback — within a single session the `Workspace` identity naturally partitions the cache, so cross-session stale hits cannot occur as long as different sessions use different `Workspace` instances.

jverkoey:
## New commit: Remap cached target graph on PIF re-transfer

I've pushed a follow-up commit that addresses the performance cost of the workspace identity fix (91afdfa).

### Problem

The `ObjectIdentifier(workspace)` inclusion in the cache signature is correct — it prevents serving stale Target references after PIF re-transfer. However, it's heavy-handed: every PIF re-transfer forces a full graph recompute (~9.5s for our 2100-target workspace), even when only source files changed and the graph structure is identical. In iterative development, Xcode re-transfers the PIF on every source change, so this cost was paid on every build.

### Solution

When the full signature (with workspace identity) misses, check if any cached entry has the same **content signature** (everything except workspace identity). If found, **remap** all Target references in the cached graph to point at the new workspace's Target objects via `workspace.target(for: guid)`, then store the remapped graph under the new full signature.

New methods on `TargetBuildGraphCache`:
- `computeContentSignature()` — same as `computeSignature()` but omits `workspaceIdentity`
- `lookupByContentSignature()` — scans cached entries for content match
- `remapGraph(_:to:)` — replaces all Target references; returns nil if any GUID is missing (structural change)

The remap path in `TargetBuildGraph.cached()` sits between the exact-match hit and the full computation miss:

```
1. Full signature hit → return cached (unchanged, <1ms)
2. Content signature match + remap succeeds → return remapped (~100-150ms)
3. Both miss → full recompute (~9.5s)
```

### Safety

`remapGraph` bails out (returns nil) if **any** cached Target GUID is not found in the new workspace. This means structural PIF changes (targets added, removed, or renamed) always fall through to a full rebuild. The remap only succeeds when the graph is structurally identical and only the object references differ.

Every Target reference in the remapped graph comes from `workspace.target(for: guid)` on the **current** workspace object, so the stale reference problem from before cannot recur.

### Confirmed with Xcode iterative builds

Tested with 3 consecutive builds in Xcode where only a single source file was changed between each build. Instruments Time Profiler traces confirm:

- `TargetBuildGraph.cached()` calls `remapGraph()` on each build (content signature match)
- `remapGraph` takes **~136ms** (vs **9.5s** for a full recompute)
- The bulk of the remap time is `ConfiguredTarget.computeGuid` during `replacingTarget()`
- **Zero** full dependency resolution (`TargetDependencyResolver.computeGraph`) appears in the trace
- Builds compile correctly with no missing dependencies

### Tests

Added `TargetBuildGraphCacheTests.swift` with 7 tests:
- Existing signature behavior (subobjects differ, identical match, different purposes)
- Content signature matches across workspace objects
- Remap produces valid graph with fresh Target references
- Remap fails when a target is removed (returns nil → full rebuild)
- Anti-regression: dictionary lookups with new-workspace ConfiguredTargets succeed after remap

Sources/SWBCore/Settings/Settings.swift
owenv, line 3115: This seems to have been accidentally included in the history

Sources/SWBCore/TargetBuildGraphCache.swift
owenv: I'm more concerned about ensuring cached dependency graphs don't persist in memory when a session is closed but the process is continuing to run builds for other sessions.

Sources/SWBCore/TargetBuildGraphCache.swift
```diff
@@ -0,0 +1,349 @@
...
+            entries[signature] = graph
+        }
+    }
+
+    /// Whether the given build command should bypass the cache.
+    /// Index preparation (prepareForIndexing) uses low QoS and
+    /// produces large dependency graphs that are rarely reused —
+    /// the memory overhead of caching them is not worth it.
+    static func shouldSkipCache(buildCommand: BuildCommand) -> Bool {
+        buildCommand.isPrepareForIndexing
```
owenv, line 115: My earlier comment was suggesting looking at the QoS property itself rather than hardcoding in a special case for indexing.

Sources/SWBCore/TargetBuildGraphCache.swift
```diff
@@ -0,0 +1,349 @@
...
+        entries.withLock { entries in
+            guard var entry = entries[signature] else { return nil }
+            entry.lastAccess = nextAccessTime()
+            entries[signature] = entry
+            return entry
+        }
+    }
+
+    /// Store a computed dependency graph for the given signature.
+    static func store(signature: Int, graph: CachedDependencyGraph) {
```
owenv, line 95: I'd prefer we use existing cache types instead of reimplementing this from scratch, a lot of the surrounding code is redundant

Sources/SWBCore/TargetBuildGraphCache.swift
```diff
@@ -0,0 +1,349 @@
...
+/// with different parameters (e.g. dependency graph analysis vs actual
+/// build). A multi-entry cache ensures these don't evict each other.
+///
+/// The cache is static (process-level) because `WorkspaceContext` is
+/// recreated on every PIF transfer, even when nothing has changed.
+public enum TargetBuildGraphCache {
+    /// The data we cache — everything needed by the
+    /// `TargetBuildGraph` memberwise init except the live context
+    /// objects (workspaceContext, buildRequest, buildRequestContext).
+    package struct CachedDependencyGraph: @unchecked Sendable {
```
owenv, line 27: Why is this @unchecked?

Sources/SWBCore/TargetBuildGraphCache.swift
```diff
@@ -0,0 +1,349 @@
...
+        }
+    }
+
+    /// Remap all Target references in a cached graph to point at
+    /// fresh objects from the new workspace.
+    ///
+    /// Returns nil if any cached Target GUID is not found in the new
+    /// workspace, which means the PIF structure changed and a full
+    /// rebuild is needed.
+    package static func remapGraph(
```
owenv, line 143: This remapping feels very fragile and error-prone. I think the underlying issue is best addressed by the PIF transfer infrastructure rather than attempting to work around it here.

Sources/SWBCore/TargetBuildGraphCache.swift
```diff
@@ -0,0 +1,349 @@
...
+        hasher.combine(buildRequest.useImplicitDependencies)
+        hasher.combine(buildRequest.useParallelTargets)
+        hasher.combine(buildRequest.skipDependencies)
+
+        // Dependency scope affects pruning
+        hasher.combine(buildRequest.dependencyScope)
+
+        // Build command affects the early-return for
+        // assembly/preprocessor. BuildCommand has associated values
+        // that prevent auto-Hashable, so we hash only the
```
owenv, line 280: This comment is incorrect

Sources/SWBCore/DiagnosticCollectingDelegate.swift
```diff
@@ -0,0 +1,51 @@
+//===----------------------------------------------------------------------===//
```
owenv, line 1: This file is needs to be added to CMakeLists.txt

Sources/SWBCore/TargetDependencyResolver.swift
```diff
@@ -104,19 +104,93 @@ public struct TargetBuildGraph: TargetGraph, Sendable {
 
     // FIXME: Report cycles via the delegate.
     //
-    /// Construct a new graph for the given build request.
+    /// Construct a new dependency graph with caching.
```
owenv, line 107: A lot of these comments are just repeating information from comments elsewhere

Tests/SWBCoreTests/TargetBuildGraphCacheTests.swift
```diff
@@ -0,0 +1,722 @@
+//===----------------------------------------------------------------------===//
```
owenv, line 1: These tests currently fail to compile, and only cover the signature computation itself. I think this change needs more comprehensive testing of incremental builds which tigger hits/misses in the cache and verify rebuilds (or not)

owenv (changes requested):
I left a few additional comments, but at a high level I think we probably need to take a different approach to introducing caching of the dependency graph than what is currently in this PR. The current soundness holes will lead to incorrect incremental builds in real-world projects, for example, when modifying OTHER_LDFLAGS in an xcconfig to introduce a new target dependency.

I think my suggested approach would be something like:
1. Refactor PIF loading to reduce unnecessary invalidation of model objects. This eliminates the need for fragile remapping of cached graphs in addition to potentially speeding up null builds a bit.
2. Extract the workspaceContext from target dependency resolution. This is largely only used for access to the project model objects, user preferences, and the SDKRegistry. The first two can be modeled as explicit inputs to the computation, and the third can largely be treated as immutable for the lifetime of the service process.
3. Extract the buildRequestContext from target dependency resolution. This is largely used for settings computation, which might introduce additional inputs to target dependency resolution, especially of implicit dependencies. Currently, changes to dependencies triggered by settings changes won't always correctly invalidate the cached graph. 
4. Eliminate feature flags which impact the dependency graph or move to modeling them as explicit inputs.
5. Introduce a simpler in-memory cache tied to the lifetime of the session and cleared upon cleaning which explicitly tracks all of the inputs and does not attempt to remap stale project model objects.

PR #1125, opened 2026-02-23
Replace O(n*m) linear scan with dictionary lookup in validateTargetCompletion

jverkoey:
Part of https://github.com/swiftlang/swift-build/issues/1111

## Problem

`OperationSystemAdaptor.validateTargetCompletion(buildSucceeded:)` iterates over every entry in `completedTaskCounts` and, for each one, calls `.filter()` over `allConfiguredTargets` comparing `.guid`. `ConfiguredTarget.guid` is a computed property that sorts the overrides dictionary, filters keys, joins strings, and allocates a new `GUID` on every call. With thousands of configured targets and thousands of completed task entries, this is an O(n*m) loop of expensive string operations running on a single thread at the end of every build.

## Solution

Build a `[GUID: ConfiguredTarget]` dictionary once before the loop (O(n), one `.guid` computation per target), then use O(1) dictionary lookups in the inner loop.

## Measurements

Tested on a 2,100+ target monorepo (~5,159 configured targets). Profiled with
Instruments Time Profiler on Apple M3 Max.

**Before (run 1):**
```
2.74 s  12.6%  OperationSystemAdaptor.validateTargetCompletion(buildSucceeded:)
2.74 s  12.6%   specialized Sequence.filter(_:)
2.74 s  12.6%    closure #1 in OperationSystemAdaptor.validateTargetCompletion(buildSucceeded:)
2.70 s  12.4%     ConfiguredTarget.guid.getter
```

**Before (run 2):**
```
2.47 s  59.7%  OperationSystemAdaptor.validateTargetCompletion(buildSucceeded:)
2.45 s  59.3%   specialized Sequence.filter(_:)
2.45 s  59.3%    closure #1 in OperationSystemAdaptor.validateTargetCompletion(buildSucceeded:)
2.42 s  58.4%     ConfiguredTarget.guid.getter
```

**After:**
```
4.00 ms   0.0%  OperationSystemAdaptor.validateTargetCompletion(buildSucceeded:)
4.00 ms   0.0%   specialized Collection.map<A, B>(_:)
4.00 ms   0.0%    specialized ContiguousArray.append(_:)
3.00 ms   0.0%     ConfiguredTarget.guid.getter
```

**~2.5s → ~4ms** (~625x improvement).

### DIFF
--- a/Sources/SWBTaskExecution/BuildDescription.swift
+++ b/Sources/SWBTaskExecution/BuildDescription.swift
@@ -171,6 +171,8 @@ package final class BuildDescription: Serializable, Sendable, Encodable, Cacheab
     /// The count of tasks in each target, used for target completion tracking.
     package let targetTaskCounts: [ConfiguredTarget: Int]
 
+    package let targetsByGuid: [ConfiguredTarget.GUID: ConfiguredTarget]
+
     package let targetDependencies: [TargetDependencyRelationship]
 
     /// Maps module names to the GUID of the configured target which will define them.
@@ -219,6 +221,7 @@ package final class BuildDescription: Serializable, Sendable, Encodable, Cacheab
         self.dependencyValidationPerTarget = settingsPerTarget.mapValues { $0.globalScope.evaluate(BuiltinMacros.VALIDATE_DEPENDENCIES) }
         self.taskActionMap = taskActionMap
         self.targetTaskCounts = targetTaskCounts
+        self.targetsByGuid = try Self.targetsByGuidMap(targetTaskCounts.keys)
         self.moduleSessionFilePath = moduleSessionFilePath
         self.diagnostics = diagnostics
         self.isBuildDirectoryCache = .init()
@@ -231,6 +234,13 @@ package final class BuildDescription: Serializable, Sendable, Encodable, Cacheab
         self.bypassActualTasks = bypassActualTasks
         self.targetsBuildInParallel = targetsBuildInParallel
         self.emitFrontendCommandLines = emitFrontendCommandLines
+    }
+
+    private static func targetsByGuidMap(_ configuredTargets: any Collection<ConfiguredTarget>) throws -> [ConfiguredTarget.GUID: ConfiguredTarget] {
+        try Dictionary(
+            configuredTargets.map { ($0.guid, $0) },
+            uniquingKeysWith: { _, _ in throw StubError.error("Duplicate GUIDs for configured targets") }
+        )
     }
 
     // MARK: Client API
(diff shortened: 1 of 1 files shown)


owenv (approved):

owenv:
@swift-ci test

Sources/SWBBuildSystem/BuildOperation.swift
```diff
@@ -1715,8 +1715,12 @@ internal final class OperationSystemAdaptor: SWBLLBuild.BuildSystemDelegate, Act
         }
 
         // By the time we've completed the build, we've seen every task for a target by definition. This ensures that the target-completed event is always sent in cases where we may be building slices of a target, as in single file builds or other builds which only build a subset of tasks in the graph. Additionally, if we're building multiple nodes, make sure to mark the target completed regardless, since we don't mark it completed in completedTaskForTarget.
+        let targetsByGuid = Dictionary(
+            operation.buildDescription.allConfiguredTargets.map { ($0.guid, $0) },
+            uniquingKeysWith: { first, _ in first }
```
jakepetroules: nit: in the `uniquingKeysWith` closure, you could do a fatalError; there should never be a duplicate here.

jverkoey: Good defensive instinct, but crashing the process seems potentially frustrating for end-users; is there a more preferred way to surface errors like that in builds?

jakepetroules: Not a great way from this particular location, hence the opposite of my typical advice.

The point is, it's a true precondition - if two targets in this list have the same GUID, something's gone catastrophically wrong. There should be no possible way for a user to reach this condition, just as there should be no possible way for a user to cause an out-of-bounds array indexing.

jverkoey: Ah that’s a bummer. Out of curiosity how does swift build handle surfacing fatal errors like this to Xcode and swift build in general? I suppose crashing just crashes the separate process, in which case a new one will be spun up anyway, but at that point we might get into a crash loop if the uuids are persistent; makes me wonder if something stronger than a crash should happen (I.e. a full cache wipe?)

What do you think?

jakepetroules: We use a combination of `throws`, and emitting errors to the delegate, depending on the context. The other approach you could take is to make `targetsByGuid` a stored property on the `BuildDescription`, and initialize it in the `BuildDescription`. Since that constructor is `throws`, you can throw an error from `uniquingKeysWith` and handle this cleanly.

Also better for performance in theory, since you can have multiple build operations off of one BuildDescription and only need to construct the dictionary once rather than on every build.

mamouneyya:
Can we merge this great change?

jakepetroules:
@swift-ci test

jverkoey:
Thank you for getting this over the finish line! Excited to get this merged upstream ❤️

PR #1126, opened 2026-02-23
Fix nondeterminism serializing platform filters in PIF

owenv:
Previously these were serialized are sets and could be reordered across process launches


owenv:
@swift-ci test

Issue #560, opened 2025-06-05
DEBUG_INFORMATION format should support "codeview" on Windows


owenv:
This is needed for parity with the existing SwiftPM build system

PR #1133, opened 2026-02-24
Add codeview debug format for windows

daveinglis:
- update xcspec for Windows to support codeview debug format
- update ObjectLibraryAssembler to support files with the same basename by finding a common root of the inputs copying the directory structure that are children of that root.

closes: https://github.com/swiftlang/swift-build/issues/560

### DIFF
--- a/Sources/SWBTaskExecution/TaskActions/ObjectLibraryAssemblerTaskAction.swift
+++ b/Sources/SWBTaskExecution/TaskActions/ObjectLibraryAssemblerTaskAction.swift
@@ -35,11 +35,43 @@ public final class ObjectLibraryAssemblerTaskAction: TaskAction {
         do {
             let options = try Options.parse(Array(task.commandLineAsStrings.dropFirst()))
             try? executionDelegate.fs.remove(options.output)
-            try executionDelegate.fs.createDirectory(options.output)
-            _ = try await options.inputs.concurrentMap(maximumParallelism: 10) { input in
-                try executionDelegate.fs.copy(input, to: options.output.join(input.basename))
+            try executionDelegate.fs.createDirectory(options.output, recursive: true)
+
+            // Track basename usage for duplicate detection and build destination mappings
+            var basenameCount: [String: Int] = [:]
+            var inputsWithDestinations: [(source: Path, destination: String)] = []
+
+            // Process each input to determine its destination name
+            for input in options.inputs {
+                let basename = input.basename
+                let count = basenameCount[basename, default: 0]
+                basenameCount[basename] = count + 1
+
+                let destinationName: String
+                if count == 0 {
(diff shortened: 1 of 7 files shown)


Sources/SWBTaskExecution/TaskActions/ObjectLibraryAssemblerTaskAction.swift
```diff
@@ -36,15 +36,89 @@ public final class ObjectLibraryAssemblerTaskAction: TaskAction {
             let options = try Options.parse(Array(task.commandLineAsStrings.dropFirst()))
             try? executionDelegate.fs.remove(options.output)
             try executionDelegate.fs.createDirectory(options.output)
+
+            // Find common root of all input paths
+            let commonRoot = findCommonRoot(of: options.inputs)
```
owenv: Thinking about this some more, there's no guarantee the object file inputs are at all close to each other in the file system. Instead of attempting to nest them, could we copy them all to the top level of the object library, but rename them if needed to avoid collisions? If we iterate over them in deterministic order the names could probably just be foo-1.o, foo-2.o, etc.

daveinglis: So ya, I was thinking about that and wanted to ask you if that was possible... will update with your suggestion.

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test

Tests/SWBBuildSystemTests/BuildCommandTests.swift
```diff
@@ -74,7 +74,7 @@ fileprivate struct BuildCommandTests: CoreBasedTests {
                 results.consumeTasksMatchingRuleTypes(excludedTypes)
                 results.checkTaskExists(.matchRule(["SwiftCompile", "normal", results.runDestinationTargetArchitecture, "Compiling \(swiftFile.basename)", swiftFile.str]))
                 results.checkTaskExists(.matchRule(["SwiftEmitModule", "normal", results.runDestinationTargetArchitecture, "Emitting module for aLibrary"]))
-                if runDestination == .linux {  // FIXME: This needs to be investigated... We should be able to use core.hostOperatingSystem.imageFormat.requiresSwiftModulewrap to test for this, but on Windows using this causes the test to fail as the SwiftModuleWrap does not seem to be added.
+                if try await getCore().hostOperatingSystem.imageFormat.requiresSwiftModulewrap  {
```
jakepetroules: nit: if you call getCore() multiple times in a test, please extract it to a local variable

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test

PR #1134, opened 2026-02-24
[SWBCore/SwiftOutputParsing] Fix performance regression in message serialization for diagnostics

akyrtzi:
After refactoring of 9dfb22531bee966521a0ae96045913ab84f6eb6a diagnostics without a source file were using the signature of the compiler task which can be enormous, causing slowdown when sending a lot of diagnostics.

For a particular target it was taking ~30 seconds to send 455 diagnostics; after this change it goes back to under a second.


akyrtzi:
@swift-ci Please test

PR #1135, opened 2026-02-25
Refactor Swift SDK toolsets to be more testable and usable independent of Swift SDKs

owenv:
Refactor how we handle toolset.json only, both to support SwiftPM initiating a build with only a toolset.json (and no Swift SDK) and to allow for easier testing of the settings overrides we generate based on a toolset. There's now a SWIFT_SDK_TOOLSETS build setting which can be injected via either SDK default properties or the build request to use a toolset, and interpreting the toolset contents is now part of settings construction. Currently, we still only respect a very small subset of the toolset fields, I'm going to tackle that in a follow-up

### DIFF
--- a/Tests/SWBTaskConstructionTests/ToolsetTaskConstructionTests.swift
+++ b/Tests/SWBTaskConstructionTests/ToolsetTaskConstructionTests.swift
@@ -1,0 +1,146 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import Testing
+
+import SWBCore
+import SWBTaskConstruction
+import SWBTestSupport
+@_spi(Testing) import SWBUtil
+
+@Suite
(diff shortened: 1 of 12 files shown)


owenv:
@swift-ci test

owenv:
Added some additional changes here to support the full range of toolset customization options relevant to the build system (excluding things like `debugger` which aren't relevant to Swift Build)

owenv:
@swift-ci test

jakepetroules:
Nice. CC @bkhouri as this has some relationship with swiftlang/swift-package-manager#9756 as well.

owenv:
@swift-ci test

owenv:
@swift-ci test

Tests/SWBTaskConstructionTests/ToolsetTaskConstructionTests.swift
```diff
@@ -0,0 +1,146 @@
...
+@Suite
+fileprivate struct ToolsetTaskConstructionTests: CoreBasedTests {
+    @Test(.requireSDKs(.host))
+    func toolsetCustomization() async throws {
+        try await withTemporaryDirectory { tmpDir in
+            let toolchainBinDir = tmpDir.join("custom-toolchain").join("bin")
+            try localFS.createDirectory(toolchainBinDir, recursive: true)
+            let customClang = toolchainBinDir.join("clang")
+            let customClangxx = toolchainBinDir.join("clang++")
+            let customSwiftc = toolchainBinDir.join("swiftc")
```
jakepetroules: needs executable suffix

owenv:
@swift-ci test

owenv:
@swift-ci test

PR #1140, opened 2026-02-26
Introduce a TARGET_TEMP_DIR_SUFFIX setting

owenv:
The intent is to have SwiftPM set this to differentiate the intermediate directories of PIF targets representing a SwiftPM target and product whose names differ only in case


owenv:
@swift-ci test

PR #1144, opened 2026-02-26
Fix single-threaded WMO invocations

owenv:
Previously, this test would fail because the driver doesn't produce per-file outputs if we don't pass some value to -num-threads.


owenv:
@swift-ci test

owenv:
@swift-ci test macOS

PR #1160, opened 2026-03-05
Ensure install name / soname does not have a leading '/' if LD_DYLIB_INSTALL_NAMEBASE is empty

owenv:
Otherwise, the unexpected leading / may lead to failures to load the library at runtime, like we observed with libPackageDescription.so on Linux


owenv:
@swift-ci test

jakepetroules (approved):

owenv:
@swift-ci test macOS

owenv:
@swift-ci test macOS

owenv:
@swift-ci test macOS

owenv:
@swift-ci test

owenv:
@swift-ci test

PR #1161, opened 2026-03-05
Use a matching linker driver when swift compiler path is overridden

owenv:
If overriding the compiler path with SWIFT_EXEC, we should override the linker driver path as well (when using swiftc) to avoid library mismatches


owenv:
@swift-ci test

owenv:
@swift-ci test

jakepetroules (approved):

PR #1165, opened 2026-03-05
Remove all use of ActorLock type

owenv:
Remove uses of ActorLock. Most of these are better modeled using AsyncOperationQueue or AsyncCache. Works around some issues seen in nightly snapshots and makes the resulting code a bit easier to understand


Sources/SWBBuildService/BuildService.swift
```diff
@@ -56,16 +56,7 @@ open class BuildService: Service, @unchecked Sendable {
     let buildManager = BuildManager()
 
     /// The cache of core objects.
-    ///
-    /// We make this a heavy cache in debug mode, so that it can be explicitly cleared (via `clearAllCaches`), which helps considerably with memory leak debugging.
```
owenv, line 60: Dropped this because I've never personally found it useful, but we could consider restoring it

owenv:
@swift-ci test

Sources/SWBBuildService/BuildService.swift
jakepetroules: I was going to ask you about that, yeah, but I've never found it useful either.

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test

Sources/SWBBuildService/Messages.swift
```diff
@@ -1664,14 +1664,14 @@ extension _Concurrency.TaskPriority {
...
+    fileprivate func withInfoOperation(operation: InfoOperation, buildRequest: BuildRequest, requestForReply: Request, serializationQueue: AsyncOperationQueue, _ work: @escaping @Sendable () async -> any Message) {
+        withInfoOperation(operation: operation, qos: buildRequest.qos, requestForReply: requestForReply, serializationQueue: serializationQueue, work)
     }
 
-    fileprivate func withInfoOperation(operation: InfoOperation, qos: SWBQoS, requestForReply: Request, lock: ActorLock, _ work: @escaping @Sendable () async -> any Message) {
+    fileprivate func withInfoOperation(operation: InfoOperation, qos: SWBQoS, requestForReply: Request, serializationQueue: AsyncOperationQueue, _ work: @escaping @Sendable () async -> any Message) {
         registerInfoOperation(operation)
         operation.addTask(_Concurrency.Task<Void, Never>(priority: .init(buildRequestQoS: qos)) {
-            await lock.withLock {
+            try? await serializationQueue.withOperation {
```
jakepetroules: Are we sure the `try?` here can't result in a dangling async operation or something?

owenv: I think you're right, I updated these to catch the error and reply with an ErrorResponse

owenv:
@swift-ci test

owenv:
@swift-ci test Linux

owenv:
Going to merge this to unblock CI

PR #1166, opened 2026-03-06
Fix shared PCH staleness issues due to missing target ordering.

rjmansfield:
When multiple targets share a precompiled header, ProcessPCH task's ordering comes from whichever target creates it first so if that target lacks a dependency on an upstream target whose headers changed, ProcessPCH runs too early which leads to a stale .gch.

Adds a virtual ordering gate node as an input to the shared ProcessPCH. Each dependent target registers its startCompilingNode and then a workspace level gate task wires them all together.

rdar://169564506
rdar://112855255

### DIFF
--- a/Sources/SWBCore/SpecImplementations/Tools/CCompiler.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/CCompiler.swift
@@ -1611,16 +1611,23 @@ public class ClangCompilerSpec : CompilerSpec, SpecIdentifierType, GCCCompatible
             let (precompFile, pchInfoAny) = delegate.createOrReuseSharedNodeWithIdentifier(sharingIdentString) { () -> (any PlannedNode, any Sendable) in
                 // This block is invoked only when we need to create a task to precompile a header.
 
+                // Ordering gate: a workspace level task will wire together all sharing targets' startCompilingNode.
+                let orderingNode = delegate.createVirtualNode("shared-pch-ordering-\(sharingIdentHashValue)")
+
                 // Construct the full path of the precomp file, which includes the hash to make it unique.
                 // FIXME: This needs to actually include a hash code, and it should ideally (for comparison reasons) be the same as Xcode.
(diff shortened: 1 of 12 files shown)


rjmansfield:
@swift-ci test

rjmansfield:
@swift-ci test

owenv:
> When multiple targets share a precompiled header, ProcessPCH task's ordering comes from whichever target creates it first so if that target lacks a dependency on an upstream target whose headers changed, ProcessPCH runs too early which leads to a stale .gch.

Let's say A and B use the same PCH and encounter this issue, B depends on C but A does not. This change will effectively introduce an implicit dependency from A to C. Is there any situation here where the user shouldn't have explicitly specified that A -> C dependency in the first place? I guess __has_include and such could have an impact?

rjmansfield:
> > When multiple targets share a precompiled header, ProcessPCH task's ordering comes from whichever target creates it first so if that target lacks a dependency on an upstream target whose headers changed, ProcessPCH runs too early which leads to a stale .gch.
> 
> Let's say A and B use the same PCH and encounter this issue, B depends on C but A does not. This change will effectively introduce an implicit dependency from A to C. Is there any situation here where the user shouldn't have explicitly specified that A -> C dependency in the first place? I guess __has_include and such could have an impact?

For them to be treated as a shared PCH don't A and B have to have the same set of compiler options, including -I options, so wouldn't they already share the same header search paths already?

owenv:
I could imagine a case where the headers included by the PCH change based on whether the headers of A have been installed yet, in which case building a PCH with the same args but at different times could yield a different result. Which is maybe an argument for incorporating the transitive dependencies of a target in the PCH hash somehow, but I'd have to think on it. This still might be the more correct change since it ensures that the PCH build depends on all the possible targets which could install headers it might transitively include

rjmansfield:
Obsoleted by https://github.com/swiftlang/swift-build/pull/1288

rjmansfield:
@swift-ci test

rjmansfield:
@swift-ci test

owenv (approved):

PR #1168, opened 2026-03-06
Don't emit package deployment target diags for aggregate targets

owenv:
There are a few places in SwiftPM where aggregate targets are used as a top-level grouping mechanism so a CLI invocation can easily build them together. We shouldn't need to specify settings for these targets, so don't check their deployment targets against that of their dependencies


owenv:
@swift-ci test

PR #1170, opened 2026-03-08
Fix a handful of assorted test issues

jakepetroules:
- Forward DEVELOPER_DIR to appintentsnltrainingprocessor: it uses xcrun to find ssu-cli, which fails if the xcode-select'ed Xcode is not the running one
- Update a test to account for the fact that FREEBSD_DEPLOYMENT_TARGET is only a build setting and not a Clang-recognized environment variable
- Implement a preconditionFailure branch in a test that wasn't actually a precondition failure


jakepetroules:
@swift-ci test

PR #1173, opened 2026-03-09
Add support for cleaning caches in addition to the build folder in CleanOperation.

cwakamo:
Clients are now able to request a cache-clean operation in addition to cleaning the build folder, or instead of cleaning the build folder. This can be used to recover from situations where the caches are stale but we can't automatically fix the problem for the user, or simply to free up disk space that might not be freed otherwise.

### DIFF
--- a/Sources/SWBBuildService/BuildOperationMessages.swift
+++ b/Sources/SWBBuildService/BuildOperationMessages.swift
@@ -507,7 +507,10 @@ final class ActiveBuild: ActiveBuildOperation {
             assert(state == .created)
             state = .started
 
-            if case .cleanBuildFolder = buildRequest.buildCommand {} else if case .cleanBuildFolderAndCaches = buildRequest.buildCommand {} else if case .cleanCaches = buildRequest.buildCommand {} else {
+            switch buildRequest.buildCommand {
+            case .cleanBuildFolder, .cleanCaches, .cleanBuildFolderAndCaches:
+                break
+            default:
                 // Once we have reached this point, we are done reporting preparation progress.
                 let statusMessage = workspaceContext.userPreferences.activityTextShorteningLevel == .full ? "Starting" : "Starting build"
                 preparationProgressDelegate!.updateProgress(statusMessage: statusMessage, showInLog: false)


cwakamo:
@swift-ci please test

Sources/SWBBuildSystem/CleanOperation.swift
```diff
@@ -265,7 +307,8 @@ package final class CleanOperation: BuildSystemOperation, TargetDependencyResolv
                 continue
             }
 
-            if isBuildFolder(buildFolderPath) || ignoreCreatedByBuildSystemAttribute {
+            // Cache folders are always safe to delete; build folders require the CreatedByBuildSystem attribute or arena membership.
```
owenv: I do wonder if we should start explicitly creating the cache dirs so we can add the xattr

neonichu: I don't see a reason why cache dirs are safer to delete. The main reason for the xattr was the worry that users would accidentally set a build setting to a bogus value and delete user data which can also be done with e.g. `COMPILATION_CACHE_CAS_PATH=$HOME`, right?

jakepetroules: Yes, I think if it's something we expect users to set externally we should consider creating it and requiring the xattr

cwakamo: That's reasonable -- do you think this blocks this PR or is it safe to land as-is and file a follow-up issue to track adding the xattr?

cwakamo:
@swift-ci please test

cwakamo:
Remaining tasks here:

- [x] Resolve merge conflicts
- [x] Make SwiftBuild automatically create the ModuleCache and CompilationCache directories as needed with the xattr indicating they're safe to delete

cwakamo:
@swift-ci please test

cwakamo:
@swift-ci please test

Sources/SWBBuildSystem/CleanOperation.swift
cwakamo: Added support for creating cache directories with CreateBuildDirectory and updated the tests.

Sources/SWBBuildService/BuildOperationMessages.swift
```diff
@@ -499,7 +507,7 @@ final class ActiveBuild: ActiveBuildOperation {
             assert(state == .created)
             state = .started
 
-            if case .cleanBuildFolder(_) = buildRequest.buildCommand {} else {
+            if case .cleanBuildFolder = buildRequest.buildCommand {} else if case .cleanBuildFolderAndCaches = buildRequest.buildCommand {} else if case .cleanCaches = buildRequest.buildCommand {} else {
```
jakepetroules: This is really hard to read, could you restructure to use switch + break?

cwakamo:
@swift-ci please test

PR #1174, opened 2026-03-10
Check ASSETCATALOG_COMPILER_GENERATE_ASSET_SYMBOLS and STRING_CATALOG_GENERATE_SYMBOLS build settings before moving compilation of Asset Catalog and String Catalog into Souece phase

Tantalum73:
Asset Catalogs and String Catalogs can generate code to allow for easy access to colors, images, and strings. Because of this, their compilation needs to happen in the Sources build phase. When configured to be in a Resources phase, we move them into Sources automatically.
#705 refactored the code to generalize the idea.

By accident, that changed a behavior where build settings to generate these symbols were not taken into account. Before #705, Asset Catalogs were only moved when `ASSETCATALOG_COMPILER_GENERATE_ASSET_SYMBOLS` was `YES`. After #705, the move happened unconditionally.

This PR fixes that by only moving Catalogs to the Sources phase when they actually produce sources to compile.


Tests/SWBTaskConstructionTests/XCStringsSymbolGenTests.swift
```diff
@@ -1083,4 +1083,95 @@ fileprivate struct XCStringsSymbolGenTests: CoreBasedTests {
...
+                results.checkTask(.matchTarget(target), .matchRule(["CopyStringsFile", "/tmp/Test/Project/build/Debug/MyFramework.framework/Versions/A/Resources/de.lproj/Localizable.strings", "/tmp/Test/Project/build/Project.build/Debug/MyFramework.build/de.lproj/Localizable.strings"])) { _ in }
+                results.checkTask(.matchTarget(target), .matchRule(["CopyStringsFile", "/tmp/Test/Project/build/Debug/MyFramework.framework/Versions/A/Resources/de.lproj/Localizable.stringsdict", "/tmp/Test/Project/build/Project.build/Debug/MyFramework.build/de.lproj/Localizable.stringsdict"])) { _ in }
+
+                results.checkNoTask(.matchTarget(target), .matchRuleType("CopyStringsFile"))
+
+                // The xcstrings file must stay in the Resources phase when symbol generation is
+                // disabled. Verify by checking that xcstrings compilation follows Swift compilation
+                // through the phase gate (Resources runs after Sources).
+                let targetArchitecture = results.runDestinationTargetArchitecture
+                let swiftRuleType = swiftFeatures.has(.emitLocalizedStrings) ? "SwiftDriver Compilation" : "CompileSwiftSources"
```
owenv, line 1169: What's this check for? It's probably a bug if lack of emit-localized-strings support is turning off the integrated driver

owenv:
@swift-ci test

Tests/SWBTaskConstructionTests/XCStringsSymbolGenTests.swift
owenv: discussed offline, this is copied over from similar tests. We'll look at cleaning them up in one batch because I suspect some of these checks are obsolete

owenv (approved):

PR #1175, opened 2026-03-10
[WIP] Consider SWIFT_EXEC when searching for testing library

bkhouri:
When calculating the testing library search path, take into consideration if SWIFT_EXEC is defined

### DIFF
--- a/Sources/SWBCore/Settings/Settings.swift
+++ b/Sources/SWBCore/Settings/Settings.swift
@@ -4528,11 +4528,18 @@ private class SettingsBuilder: ProjectMatchLookup {
             table.push(BuiltinMacros.SWIFT_SYSTEM_INCLUDE_PATHS, BuiltinMacros.namespace.parseStringList(["$(inherited)", "$(TEST_LIBRARY_SEARCH_PATHS$(TEST_BUILD_STYLE))"]))
 
             // If the toolchain contains a copy of Swift Testing, prefer it.
-            let toolchainPath = Path(scope.evaluateAsString(BuiltinMacros.TOOLCHAIN_DIR))
-            if let toolchain = core.toolchainRegistry.toolchains.first(where: { $0.path == toolchainPath }) {
+            let swiftExecPath = Path(scope.evaluateAsString(BuiltinMacros.SWIFT_EXEC))
+            if swiftExecPath.exists(fs: workspaceContext.fs) {
                 let platformName = scope.evaluate(BuiltinMacros.PLATFORM_NAME)
-                if let testingLibrarySearchPath = toolchain.testingLibrarySearchPath(forPlatformNamed: platformName) {
-                    table.push(BuiltinMacros.TEST_LIBRARY_SEARCH_PATHS, BuiltinMacros.namespace.parseStringList(["$(inherited)", testingLibrarySearchPath.str]))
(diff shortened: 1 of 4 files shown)


Sources/SWBCore/Settings/Settings.swift
```diff
@@ -4528,11 +4528,18 @@ private class SettingsBuilder: ProjectMatchLookup {
             table.push(BuiltinMacros.SWIFT_SYSTEM_INCLUDE_PATHS, BuiltinMacros.namespace.parseStringList(["$(inherited)", "$(TEST_LIBRARY_SEARCH_PATHS$(TEST_BUILD_STYLE))"]))
 
             // If the toolchain contains a copy of Swift Testing, prefer it.
-            let toolchainPath = Path(scope.evaluateAsString(BuiltinMacros.TOOLCHAIN_DIR))
-            if let toolchain = core.toolchainRegistry.toolchains.first(where: { $0.path == toolchainPath }) {
+            let swiftExecPath = Path(scope.evaluateAsString(BuiltinMacros.SWIFT_EXEC))
```
owenv, line 4531: To be fully correct this might need to do something like the compiler tool lookup:
```
// Get the path to the compiler.
        let path = scope.evaluate(BuiltinMacros.SWIFT_TOOLS_DIR).nilIfEmpty.map(Path.init)?.join(compilerFileName)
            ?? scope.evaluate(BuiltinMacros.SWIFT_EXEC).nilIfEmpty.map({ $0.isAbsolute
                ? $0
                : Path(producer.hostOperatingSystem.imageFormat.executableName(basename: $0.str)) })
            ?? Path(compilerFileName)
        let userSpecifiedBlocklists = scope.evaluate(BuiltinMacros.BLOCKLISTS_PATH).nilIfEmpty.map { Path($0) }
        let toolPath = self.resolveExecutablePath(producer, path)
```

Then try to determine if that's outside any of the toolchains we've registered

Sources/SWBCore/Settings/Settings.swift
```diff
@@ -4528,11 +4528,18 @@ private class SettingsBuilder: ProjectMatchLookup {
...
+            if swiftExecPath.exists(fs: workspaceContext.fs) {
```
owenv, line 4532: We need to be careful about accessing the filesystem during settings construction because it's a hot path and needs to be invalidated when files change. I wonder if we can defer this to when we construct the actual command line in SwiftCompilerSpec

Sources/SWBUtil/Path.swift
```diff
@@ -594,6 +594,14 @@ public struct Path: Serializable, Sendable {
         return nil
     }
 
+    /// Does the path exists on the filesystem `fs`
+    ///
+    /// - parameter fs: A file system proxy
+    /// - returns: `true` if the file exists. Otherwise, `false`
+    public func exists(fs: any FSProxy) -> Bool {
```
owenv, line 601: I think we should drop this, IMO fs.exists(path) does a better job of emphasizing which FSProxy is used

bkhouri: we should maybe also drop for the same reasons ;)

```
    /// Resolves all symlinks in the path and returns a new path containing no symlinks.
    public func resolveSymlink(fs: any FSProxy) throws -> Path {
        return try fs.realpath(self)
    }
```

PR #1178, opened 2026-03-10
Fix the search path order for the clang module verifier

ian-twilightcoder:
The clang module verifier is putting BUILT_PRODUCTS_DIR at the end of the search path list, but it's very important that it comes first. Otherwise includes from the framework being tested can resolve to a different copy of the framework depending on what gets added on the command line/from the SDK/etc.

rdar://161918113


ian-twilightcoder:
@swift-ci test

PR #1179, opened 2026-03-11
Handle libc-related Windows flags that are currently hardcoded on the SwiftPM side

jakepetroules:


jakepetroules:
Draft because I still need to update tests

jakepetroules:
@swift-ci test

PR #1180, opened 2026-03-11
Don't delete relative paths computed when cleaning

owenv:
If a macro like OBJROOT evaluates to a relative path, we shouldn't attempt to delete it while cleaning because build configuration can't make assumptions about the build service executable's current working directory, so the resulting behavior will be unpredictable.

I discovered this while debugging an occasional SQLite assert in the test suite which reported a build db being unlinked while open. It turns out:
1. The Swift dependency scanner is configuring the clang importer with a working directory, which clang is then setting with chdir (we should fix this in the compiler)
2. The external target build system tests which run a clean operation had OBJROOT evaluate to 'build/'
3. When these two raced in just the right way, the cleaning tests could delete another random test's build directory at just the right time to trigger the llbuild assert

### DIFF
--- a/Sources/SWBBuildSystem/CleanOperation.swift
+++ b/Sources/SWBBuildSystem/CleanOperation.swift
@@ -265,6 +265,11 @@ package final class CleanOperation: BuildSystemOperation, TargetDependencyResolv
                 continue
             }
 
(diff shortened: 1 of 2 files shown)


owenv:
@swift-ci test

PR #1182, opened 2026-03-11
Stop passing -add_ast_path / running modulewrap when compiler supports explicit modules based debug info

owenv:
Reapply the change to drop use of -add_ast_path when the compiler is capable of using explicit modules based debug info


owenv:
@swift-ci test

owenv:
@swift-ci test

PR #1183, opened 2026-03-11
Remove stale EagerLinkingTBDs across workspace switches

rjmansfield:
When switching between workspaces that share the same OBJROOT, llbuild's delta stale file removal fails because each workspace uses a different build.db with no previous state. This means stale TBDs in EagerLinkingTBDs/ can persist and cause link/runtime errors.

rdar://109491531
rdar://165932649

### DIFF
--- a/Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/ProductPostprocessingTaskProducer.swift
+++ b/Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/ProductPostprocessingTaskProducer.swift
@@ -38,6 +38,17 @@ enum InstallAPIDestination {
             return []
         }
     }
+
+    /// EagerLinkingTBDs are in a shared directory so their outputs should be
+    /// tracked by the workspace SFR.
+    var staleFileRemovalScope: StaleFileRemovalScope {
+        switch self {
+        case .eagerLinkingTBDDir:
+            return .workspace
+        case .builtProduct:
+            return .target
+        }
+    }
 }
 
 package final class ProductPostprocessingTaskProducer: PhasedTaskProducer, TaskProducer {
@@ -794,7 +805,8 @@ func addCommonInstallAPITasks(_ producer: PhasedTaskProducer, _ scope: MacroEval
         dsymPath = producer.context.producedDSYM(forVariant: variant)
 
     }
-    let delegate = PhasedProducerBasedTaskGenerationDelegate(producer: producer, context: producer.context, taskOptions: destination.correspondingTaskOrderingOptions, phaseStartNodes: phaseStartNodes, phaseEndTask: phaseEndTask)
+    let delegate = PhasedProducerBasedTaskGenerationDelegate(producer: producer, context: producer.context, taskOptions: destination.correspondingTaskOrderingOptions, staleFileRemovalScope: destination.staleFileRemovalScope, phaseStartNodes: phaseStartNodes, phaseEndTask: phaseEndTask)
+
     let swiftTBDFiles = producer.context.generatedTBDFiles(forVariant: variant)
     await producer.context.tapiSpec.constructTAPITasks(CommandBuildContext(producer: producer.context, scope: scope, inputs: inputs, output: tapiOutputNode.path, commandOrderingInputs: dependencyInputs, commandOrderingOutputs: [tapiOrderingNode]), delegate, generatedTBDFiles: swiftTBDFiles, builtBinaryPath: builtBinaryPath, fileListPath: jsonPath, dsymPath: dsymPath)
 
@@ -816,7 +828,7 @@ extension FrameworkProductTypeSpec {
                 output = producer.context.createNode(scope.evaluate(BuiltinMacros.EAGER_LINKING_INTERMEDIATE_TBD_DIR).join(scope.evaluate(BuiltinMacros.WRAPPER_NAME)).join(fileName))
             }
 
-            await producer.appendGeneratedTasks(&tasks, options: destination.correspondingTaskOrderingOptions) { delegate in
+            await producer.appendGeneratedTasks(&tasks, options: destination.correspondingTaskOrderingOptions, staleFileRemovalScope: destination.staleFileRemovalScope) { delegate in
                 producer.context.symlinkSpec.constructSymlinkTask(CommandBuildContext(producer: producer.context, scope: scope, inputs: [], output: output.path), delegate, toPath: target, repairViaOwnershipAnalysis: true)
             }
         }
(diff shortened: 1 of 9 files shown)


rjmansfield:
@swift-ci test

owenv:
Cleaning these up is the right thing to do, but there are a variety of other filetypes that might fall into a similar category. It's been a while since I looked at the llbuild stale file removal, but I wonder if there's something we can do to trigger it to run on the first clean build with a new build database

rjmansfield:
> Cleaning these up is the right thing to do, but there are a variety of other filetypes that might fall into a similar category. It's been a while since I looked at the llbuild stale file removal, but I wonder if there's something we can do to trigger it to run on the first clean build with a new build database

I'll admit I didn't dig too deep into a llbuild solution, but isn't the problem with a full clean-on-new build.db is that the roots can be shared across workspaces/build.db ? e.g. DSTROOT, OBJROOT, SYMROOT, etc so removing stuff not in expectedOutputs from one may wipe out other from other workspaces?

I went this route because removing EagerLinkingTBDs is the only one, AFAICT, where it's safe because all the outputs are statically declared.  The other reason is that if they're accidentally deleted they should be fast to regenerate, but leaving a stale tbd can cause more serious problems.

owenv:
I'm surprised we're getting a totally clean build DB since usually that's saved relative to OBJROOT - do you know what mechanism is causing it to get dropped? 

> I went this route because removing EagerLinkingTBDs is the only one, AFAICT, where it's safe because all the outputs are statically declared. The other reason is that if they're accidentally deleted they should be fast to regenerate, but leaving a stale tbd can cause more serious problems.

I'm not entirely opposed to doing this if it turns out to be the most pragmatic solution, but it would be better to tweak the existing mechanism if we can find a way to make it work. unfortunately it's been a long time since I last looked at stale file removal so I don't recall all the details of how it works

rjmansfield:
> I'm surprised we're getting a totally clean build DB since usually that's saved relative to OBJROOT - do you know what mechanism is causing it to get dropped?
 
Looks so a workspace change means a new entry in the build.db. Not an entirely new build.db being used, but the same problem that llbuild is not able to detect prior state.

`Sources/SWBTaskConstruction/StaleFileRemovalContext.swift:59:            staleFileRemovalIdentifier += "target-\(target.name)-\(target.guid)-\(configurationName)\(targetPlatform)"`

so workspace change means a different guid which means different SFR identifier so then llbuild has no prior expectedOutputs for that key and skips deletion. In both of the radars mentioned, there's cross workspace switching involved.

owenv:
IDK how realistic it is, but I wonder if these outputs should be tied to the workspace-wide stale file removal task instead, since that has a stable ID, and these are products in a global location in search paths. Not sure if that might have unexpected implications on other build behavior

rjmansfield:
OK, I can look into that . I don't have any understanding of the broader implications either. Part of the reason I went with this current approach is because it is limited in scope.

rjmansfield:
@swift-ci test

Sources/SWBTaskExecution/BuildDescription.swift
```diff
@@ -544,6 +544,10 @@ package final class BuildDescriptionBuilder {
 
     private let allOutputPaths: Set<Path>
 
+    /// EagerLinkingTBDs directories used to move TBD outputs from per-target
+    /// SFR to workspace SFR.
+    private let eagerLinkingTBDDirs: Set<Path>
```
owenv: Maybe we don't need to do it in this PR, but I think it's be interesting to generalize this beyond "eagerLinkingTBDDirs" to "workspaceWideStaleFileRemovalDirs". In theory it might make sense to treat e.g. the products directory the same way

Sources/SWBTaskExecution/BuildDescription.swift
```diff
@@ -603,7 +608,33 @@ package final class BuildDescriptionBuilder {
...
+            // delta can remove stale TBDs across workspace switches.
+            var sfrOutputPathsPerTarget = outputPathsPerTarget
+            if !eagerLinkingTBDDirs.isEmpty && staleFileRemovalIdentifierPerTarget[nil] != nil {
+                var workspaceTBDOutputs = [Path]()
+                for (configuredTarget, outputPaths) in sfrOutputPathsPerTarget {
+                    guard let configuredTarget else { continue }
+                    var tbdOutputs = [Path]()
+                    var otherOutputs = [Path]()
+                    for path in outputPaths {
+                        if eagerLinkingTBDDirs.contains(where: { $0.isAncestorOrEqual(of: path) }) {
```
owenv: I could see this check being slow on large builds with many tasks/outputs. I wonder if we could do this check when we first create outputPathsPerTarget instead. Right now that looks like:

```
if !(task is GateTask) && task.ruleInfo.first != "CreateBuildDirectory" {
                let target = task.forTarget

                // Ignore resign tasks for stale file removal, <rdar://problem/42642132> tracks a more generic solution for this.
                if task.outputs.filter({ $0.name.hasPrefix("ReSign") }).count > 0 || task.ruleInfo.first == "ClangStatCache" {
                    continue
                }

                var outputs = outputPathsPerTarget[target] ?? [Path]()
                outputs.append(contentsOf: task.outputs.map { $0.path }.filter { !$0.str.isEmpty })
                outputPathsPerTarget[target] = outputs
            }
```

Potentially `task` could get some new field like `staleFileRemovalScope` with `target` and `workspace` cases? Then if the scope is `workspace` we assign it to  `outputPathsPerTarget[nil]` instead of `outputPathsPerTarget[target]`, and the tapi task which produces the TBD would be responsible for setting `workspace` scope, and could do so without an ancestor check? WDYT?

rjmansfield: Yeah, this seems like a better approach and more extensible.

owenv:
@swift-ci test

Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/ProductPostprocessingTaskProducer.swift
```diff
@@ -795,6 +795,13 @@ func addCommonInstallAPITasks(_ producer: PhasedTaskProducer, _ scope: MacroEval
 
     }
     let delegate = PhasedProducerBasedTaskGenerationDelegate(producer: producer, context: producer.context, taskOptions: destination.correspondingTaskOrderingOptions, phaseStartNodes: phaseStartNodes, phaseEndTask: phaseEndTask)
```
jakepetroules: Can we please pass in the stale file removal scope into the delegate initializer? This being mutable makes me nervous, and it doesn't look like the mutability is necessary.

rjmansfield:
@swift-ci test

rjmansfield:
@swift-ci test

PR #1186, opened 2026-03-13
Fix handling of static archive archiver inputs on non-Darwin/Windows

owenv:
The unix archiver is unable to merge static archives. Instead, we need to extract the individual objects from each input archive, then feed those to the final archiver invocation. This is implemented as a new step in the existing LinkerTaskAction.

Fixes an issue where building Swift toolchain components with --multiroot-data-file would fail on linux because we'd produce malformed archives for products of root packages depending on static libraries produced by other root packages

### DIFF
--- a/Sources/SWBTaskExecution/TaskActions/LinkerTaskAction.swift
+++ b/Sources/SWBTaskExecution/TaskActions/LinkerTaskAction.swift
@@ -24,24 +24,33 @@ public final class LinkerTaskAction: TaskAction {
     private let expandResponseFiles: Bool
     private let responseFileFormat: ResponseFileFormat
 
-    public init(expandResponseFiles: Bool, responseFileFormat: ResponseFileFormat) {
+    /// Whether static archive inputs should be extracted to their constituent
+    /// object files before invoking the archiver. Apple's libtool is capable of
+    /// directly merging static archives, but other tools like ar/llvm-ar/GNU libtool
+    /// are not.
+    private let extractArchiveInputs: Bool
+
+    public init(expandResponseFiles: Bool, responseFileFormat: ResponseFileFormat, extractArchiveInputs: Bool) {
         self.expandResponseFiles = expandResponseFiles
         self.responseFileFormat = responseFileFormat
+        self.extractArchiveInputs = extractArchiveInputs
         super.init()
     }
 
     public override func serialize<T: Serializer>(to serializer: T) {
-        serializer.beginAggregate(3)
+        serializer.beginAggregate(4)
         serializer.serialize(expandResponseFiles)
         serializer.serialize(responseFileFormat)
+        serializer.serialize(extractArchiveInputs)
         super.serialize(to: serializer)
         serializer.endAggregate()
     }
 
     public required init(from deserializer: any Deserializer) throws {
-        try deserializer.beginAggregate(3)
+        try deserializer.beginAggregate(4)
         self.expandResponseFiles = try deserializer.deserialize()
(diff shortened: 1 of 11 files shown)


Sources/SWBTaskExecution/TaskActions/LinkerTaskAction.swift
```diff
@@ -24,24 +24,32 @@ public final class LinkerTaskAction: TaskAction {
     private let expandResponseFiles: Bool
     private let responseFileFormat: ResponseFileFormat
 
-    public init(expandResponseFiles: Bool, responseFileFormat: ResponseFileFormat) {
+    /// Whether static archive inputs should be extracted to their constituent
+    /// object files before invoking the archiver. Required on platforms where
+    /// `ar` cannot merge archive inputs directly (e.g. Linux).
```
jakepetroules: Does `ar` actually merge archive inputs directly on Darwin, or is that just because we use `libtool` for that instead of `ar`?

owenv: It's because we use libtool on Darwin

owenv:
@swift-ci test

Sources/SWBTaskExecution/TaskActions/LinkerTaskAction.swift
```diff
@@ -68,6 +76,101 @@ public final class LinkerTaskAction: TaskAction {
...
+
+        let arPath = commandLine[0]
+        let outputPaths = Set(task.outputPaths.map(\.str))
+
+        var result = commandLine
+        var insertionOffset = 0
+        var archiveIndex = 0
+
+        for (originalIndex, arg) in commandLine.enumerated() {
+            guard arg.hasSuffix(".a"), !outputPaths.contains(arg) else { continue }
```
jakepetroules, line 136: Some comments might be nice - "Skip the output archive, as we only want to handle input archives here" or similar

Sources/SWBTaskExecution/TaskActions/LinkerTaskAction.swift
```diff
@@ -68,6 +76,101 @@ public final class LinkerTaskAction: TaskAction {
...
+                environment: task.environment.bindingsDictionary,
+                workingDirectory: extractDir,
+                processDelegate: processDelegate
+            )
+
+            if let error = processDelegate.executionError {
+                throw StubError.error(error)
+            }
+            guard success else {
+                throw StubError.error("Failed to extract static archive: \(arg)")
```
jakepetroules, line 156: Does processDelegate give you the exit code and such back? Would be nice to use ProcessNonZeroExitError here if possible.

Sources/SWBTaskExecution/TaskActions/LinkerTaskAction.swift
jakepetroules: Can you update the comment to be correct? It's basically: Apple libtool has capabilities that GNU libtool and GNU ar or LLVM llvm-ar do not; it's not about the behavior of ar itself differing.

jakepetroules (approved):
Will approve to unblock you to merge whenever ready but would prefer to see feedback addressed.

Sources/SWBTaskExecution/TaskActions/LinkerTaskAction.swift
owenv: We don't do a very good job of preserving this as it threads its way through the stack today, but we should improve it

owenv:
@swift-ci test

owenv:
@swift-ci test

PR #1188, opened 2026-03-13
Model Windows testing library paths as a list so whitespace is properly handled

owenv:
Otherwise if testingLibraryPath.strWithPosixSlashes has spaces it can split the search path arguments unexpectedly


owenv:
@swift-ci test

PR #1189, opened 2026-03-14
Clean up how environment is passed to llbuild

jakepetroules:
Much of this came out of the investigation into another issue which turned out to be a dead end, but as a secondary effect, this centralizes and makes it easier to understand to full base environment that is given to the llbuild BuildSystem and merged with individual tasks' environment overrides. That's now in WorkspaceContext.mergedBuildEnvironment.

This fixed a number of issues, including where in some build operation tests we were not correctly replicating the full environment due to not considering all extension points that could add to it.

It also ensures that the error messages we print when a process failed to execute _actually_ show the environment they were invoked with. This was often missing. The only exception to this is the LLBUILD_ meta environment variables set internally by llbuild, which can't easily be propagated back without more changes Some of this was due to the aforementioned changes, and some of it was due to passing inconsistent arguments to Process.getOutput vs RunProcessNonZeroExitError. There are some additional opportunities for improvement there...

It updates some comments in the explicitModulesEnvironment() test and in the llbuild manifest construction related to #835, since the comments that were there turned out to be false.

Finally, a buildEnvironment() test was added which leverages `/usr/bin/env` to guarantee that llbuild invokes subprocesses with a clean environment controlled by the client APIs without inheriting anything from the process-level environment.

### DIFF
--- a/Tests/SWBBuildSystemTests/BuildTaskBehaviorTests.swift
+++ b/Tests/SWBBuildSystemTests/BuildTaskBehaviorTests.swift
@@ -19,10 +19,13 @@ import SwiftBuildTestSupport
 import SWBTaskExecution
 @_spi(Testing) import SWBUtil
 import SWBLibc
+import SWBLLBuild
 
 import class SWBBuildSystem.BuildOperation
 import class SWBTaskExecution.Task
 import SWBProtocol
+import Foundation
+import SWBServiceCore
 
 private final class MockTaskTypeDescription: TaskTypeDescription {
     init(isUnsafeToInterrupt: Bool = false) {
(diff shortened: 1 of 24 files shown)


jakepetroules:
@swift-ci test

Tests/SWBBuildSystemTests/BuildTaskBehaviorTests.swift
```diff
@@ -972,6 +975,118 @@ fileprivate struct BuildTaskBehaviorTests: CoreBasedTests {
...
+            public override func serialize<T: Serializer>(to serializer: T) { fatalError("not used") }
+            public required init(from deserializer: any Deserializer) throws { fatalError("not used") }
+        }
+
+        let envTask = createTask(ruleInfo: ["env"], commandLine: ["/usr/bin/env"], additionalOutput: [], inputs: [], outputs: [MakePlannedVirtualNode("<ECHO>")], action: EnvTaskAction())
+
+        let core = try await Core.createInitializedTestingCore(pluginLoadingFilter: { identifier in
+            // Prevent extra environment from platform plugins like Android, etc.
+            // Also need Apple because not all specs are properly modularized yet
+            identifier == "com.apple.dt.SWBUniversalPlatformPlugin" || identifier == "com.apple.dt.SWBApplePlatformPlugin"
```
owenv: I think this needs the host plugin as well for any plugins which register a developer directory extension

jakepetroules: Yep, fixed.

jakepetroules:
@swift-ci test

PR #1247, opened 2026-03-26
Preliminary OpenBSD support with swiftpm.

3405691582:
Here, we require manually set ar otherwise swb tries to call ar-lld,
which does not exist. Additionally, since the swift-build tests call
openpty, linking with libutil is required for the tests to build and run.

There was reticence to ensure that swiftmodule names are versioned with a
preference for unversioning triples for the platform, which landed on the
compiler side in https://github.com/swiftlang/swift/pull/88224.

Lastly, there are some other changes still required to improve swb here,
and address unit tests failing for the platform, but this should be the
minimal set.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -328,7 +328,11 @@ let package = Package(
             resources: [
                 .copy("TestData")
             ],
-            swiftSettings: swiftSettings(languageMode: .v5)), // Temporarily downgraded from Swift 6 mode due to a source break in 1/31/26 nightly snapshot (rdar://169461269)
+            swiftSettings: swiftSettings(languageMode: .v5), // Temporarily downgraded from Swift 6 mode due to a source break in 1/31/26 nightly snapshot (rdar://169461269)
+            linkerSettings: [
+                // required for openpty.
+                .linkedLibrary("util", .when(platforms: [.openbsd])),
+            ]),
         .testTarget(
             name: "SWBProjectModelTests",
             dependencies: ["SWBProjectModel"],
--- a/Sources/SWBGenericUnixPlatform/Plugin.swift
+++ b/Sources/SWBGenericUnixPlatform/Plugin.swift
@@ -124,6 +124,8 @@ func _deploymentTargetSettingName(os: String) -> String? {
     switch os {
     case "freebsd":
         return "FREEBSD_DEPLOYMENT_TARGET"
+    case "openbsd":
+        return "OPENBSD_DEPLOYMENT_TARGET"
     default:
         return nil
     }
@@ -157,14 +159,20 @@ struct GenericUnixSDKRegistryExtension: SDKRegistryExtension {
 
                     "AR": "llvm-ar",
                 ]
+            case .openbsd:
+                defaultProperties = [
+                    "GENERATE_TEXT_BASED_STUBS": "NO",
+                    "GENERATE_INTERMEDIATE_TEXT_BASED_STUBS": "NO",
+                    "AR": "ar",
+                ]
             default:
                 defaultProperties = [:]
             }
 
             let shouldUseLLD = {
                 switch operatingSystem {
-                case .freebsd:
-                    // FreeBSD is always LLVM-based.
+                case .freebsd, .openbsd:
+                    // FreeBSD and OpenBSD are always LLVM-based.
                     return true
                 case .linux:
                     // Amazon Linux 2 has a gold linker bug see: https://sourceware.org/bugzilla/show_bug.cgi?id=23016.
@@ -190,9 +198,9 @@ struct GenericUnixSDKRegistryExtension: SDKRegistryExtension {
                 tripleEnvironment = ""
             }
 
+            let realTripleVersion = try Version(ProcessInfo.processInfo.operatingSystemVersion).zeroTrimmed.description
             let deploymentTargetSettings: [String: PropertyListItem]
             if let deploymentTargetSettingName = _deploymentTargetSettingName(os: tripleSystem) {
-                let realTripleVersion = try Version(ProcessInfo.processInfo.operatingSystemVersion).zeroTrimmed.description
                 deploymentTargetSettings = [
                     "DeploymentTargetSettingName": .plString(deploymentTargetSettingName),
                     "DefaultDeploymentTarget": .plString(realTripleVersion),
@@ -205,7 +213,7 @@ struct GenericUnixSDKRegistryExtension: SDKRegistryExtension {
 
             return try (.root, platform, [
                 "Type": .plString("SDK"),
-                "Version": .plString(Version(ProcessInfo.processInfo.operatingSystemVersion).zeroTrimmed.description),
+                "Version": .plString(realTripleVersion),
                 "CanonicalName": .plString(operatingSystem.xcodePlatformName),
                 "IsBaseSDK": .plBool(true),
                 "DefaultProperties": .plDict([
--- a/Sources/SWBUtil/Architecture.swift
+++ b/Sources/SWBUtil/Architecture.swift
@@ -99,7 +99,7 @@ public struct Architecture: Sendable {
                 return withUnsafeBytes(of: &buf.machine) { buf in
                     let data = Data(buf)
                     let value = String(decoding: data[0...(data.lastIndex(where: { $0 != 0 }) ?? 0)], as: UTF8.self)
-                    #if os(FreeBSD)
+                    #if os(FreeBSD) || os(OpenBSD)
                     switch value {
                     case "amd64":
                         return "x86_64"
--- a/Tests/SWBBuildServiceTests/BuildServiceTests.swift
+++ b/Tests/SWBBuildServiceTests/BuildServiceTests.swift
@@ -78,7 +78,7 @@ fileprivate struct BuildServiceTests: CoreBasedTests {
         .init(triple: "x86_64-unknown-freebsd14", platformName: "freebsd", sdkVariant: nil, deploymentTargetSettingName: "FREEBSD_DEPLOYMENT_TARGET", deploymentTarget: "14"),
 
         // OpenBSD
-        .init(triple: "x86_64-unknown-openbsd", platformName: "openbsd", sdkVariant: nil, deploymentTargetSettingName: nil, deploymentTarget: nil),
+        .init(triple: "x86_64-unknown-openbsd7.8", platformName: "openbsd", sdkVariant: nil, deploymentTargetSettingName: "OPENBSD_DEPLOYMENT_TARGET", deploymentTarget: "7.8"),
 
         // QNX
         .init(triple: "aarch64-unknown-nto-qnx", platformName: "qnx", sdkVariant: nil, deploymentTargetSettingName: nil, deploymentTarget: nil),
--- a/Tests/SWBTaskConstructionTests/ShellScriptTaskConstructionTests.swift
+++ b/Tests/SWBTaskConstructionTests/ShellScriptTaskConstructionTests.swift
@@ -155,6 +155,7 @@ extension CoreBasedTests {
                     // Not actually recognized by clang
                     "ANDROID_DEPLOYMENT_TARGET",
                     "FREEBSD_DEPLOYMENT_TARGET",
+                    "OPENBSD_DEPLOYMENT_TARGET",
                     "QNX_DEPLOYMENT_TARGET",
                 ])
                 let suffix = "_DEPLOYMENT_TARGET"


Sources/SWBGenericUnixPlatform/Plugin.swift
```diff
@@ -190,9 +198,13 @@ struct GenericUnixSDKRegistryExtension: SDKRegistryExtension {
                 tripleEnvironment = ""
             }
 
+            let realTripleVersion = try Version(ProcessInfo.processInfo.operatingSystemVersion).zeroTrimmed.description
+            if operatingSystem == .openbsd {
+                defaultProperties["SWIFT_TARGET_TRIPLE"] = .plString("$(CURRENT_ARCH)-unknown-$(SWIFT_PLATFORM_TARGET_PREFIX)\(realTripleVersion)")
```
owenv: Here, I think we might want to try defining an OPENBSD_DEPLOYMENT_TARGET setting, which might lead us to automatically construct the appropriate target triple. @jakepetroules setup FREEBSD_DEPLOYMENT_TARGET and might know if we can safely reuse that pattern

owenv (commented):
Thanks! This looks great. One comment about the triple handling I want to get a second opinion on before approving

Sources/SWBGenericUnixPlatform/Plugin.swift
jakepetroules: Yup, look for anywhere that's doing FREEBSD_DEPLOYMENT_TARGET and we should just be able to do OPENBSD_DEPLOYMENT_TARGET the same way.

Sources/SWBGenericUnixPlatform/Plugin.swift
```diff
@@ -157,14 +157,22 @@ struct GenericUnixSDKRegistryExtension: SDKRegistryExtension {
 
                     "AR": "llvm-ar",
                 ]
+            case .openbsd:
+                defaultProperties = [
+                    // Workaround to avoid `-dependency_info`.
+                    "LD_DEPENDENCY_INFO_FILE": .plString(""),
```
jakepetroules: You probably don't need this one (nor does FreeBSD) since this is zeroed out in UnixLd.xcspec already.

Package.swift
```diff
@@ -328,7 +328,10 @@ let package = Package(
             resources: [
                 .copy("TestData")
             ],
-            swiftSettings: swiftSettings(languageMode: .v5)), // Temporarily downgraded from Swift 6 mode due to a source break in 1/31/26 nightly snapshot (rdar://169461269)
+            swiftSettings: swiftSettings(languageMode: .v5), // Temporarily downgraded from Swift 6 mode due to a source break in 1/31/26 nightly snapshot (rdar://169461269)
+            linkerSettings: [
+                .linkedLibrary("util", .when(platforms: [.openbsd])),
```
jakepetroules, line 334: nit: adding a comment stating this is needed for `openpty` would be nice

jakepetroules:
@3405691582 while we have you here... I tried building Swift Build using your container and ran into an issue where stdatomic.h was not found via `#include <stdatomic.h>`. I had to do this prior to `swift test`:

```
cp /usr/lib/clang/19/include/stdatomic.h /usr/include/stdatomic.h
```

Any ideas?

3405691582:
I can answer this one quickly...

> @3405691582 while we have you here... I tried building Swift Build using your container and ran into an issue where stdatomic.h was not found via `#include <stdatomic.h>`. I had to do this prior to `swift test`:
> 
> ```
> cp /usr/lib/clang/19/include/stdatomic.h /usr/include/stdatomic.h
> ```
> 
> Any ideas?

This issue specifically is a local toolchain bug where the system clang includes aren't symlinked on installation. I've fixed this locally and should have a new container image with the 6.2.4 tool chain including this fix available soon.

Sources/SWBGenericUnixPlatform/Plugin.swift
3405691582: Done. This way necessitated changing SwiftCompiler.swift in SWBCore in order to ensure that the deployment target ended up in the triple when copying swiftmodule files named after triples; please take a close look at this change to make sure it is appropriate.

Sources/SWBGenericUnixPlatform/Plugin.swift
3405691582: Great; removed.

Package.swift
3405691582: Done.

Sources/SWBGenericUnixPlatform/Plugin.swift
jakepetroules: Should do openbsd here too, to match the other place that's stating lld should be used for OpenBSD

Sources/SWBGenericUnixPlatform/Plugin.swift
jakepetroules: Did you mean you want to ensure the version does _not_ appear in the swiftmodule filename?

Sources/SWBGenericUnixPlatform/Plugin.swift
```diff
@@ -157,14 +161,21 @@ struct GenericUnixSDKRegistryExtension: SDKRegistryExtension {
 
                     "AR": "llvm-ar",
                 ]
+            case .openbsd:
+                defaultProperties = [
+                    "GENERATE_TEXT_BASED_STUBS": "NO",
+                    "GENERATE_INTERMEDIATE_TEXT_BASED_STUBS": "NO",
+                    "OPENBSD_DEPLOYMENT_TARGET": .plString("\(realTripleVersion)"),
```
jakepetroules: This should be covered by the DefaultDeploymentTarget case below...

Sources/SWBGenericUnixPlatform/Plugin.swift
3405691582: The version *must* appear in the swiftmodule filename on OpenBSD, otherwise the module will fail to match the versioned target. As I understand it, the lookup in `destinationModuleFileName` replaces the `SWIFT_DEPLOYMENT_TARGET` with an empty string in the `SWIFT_TARGET_TRIPLE`, which we don't want for this platform.

Sources/SWBGenericUnixPlatform/Plugin.swift
3405691582: Done.

Sources/SWBGenericUnixPlatform/Plugin.swift
3405691582: Done; removed.

Sources/SWBGenericUnixPlatform/Plugin.swift
jakepetroules: I think the problem lies elsewhere, because I recall we had a similar issue with FreeBSD and I think @etcwilde or someone else had to fix something in the compiler.

If you download the FreeBSD toolchain https://download.swift.org/tmp-ci-nightly/development/freebsd-14_ci_latest.tar.gz you'll see the module filenames don't contain the version.

If it absolutely needs to be different for OpenBSD, then ok, but I want to be certain before we move forward since I'm not an expert on the compiler level pieces here and at first glance it _seems_ OpenBSD should follow the same pattern as FreeBSD (and most platforms, for that matter).

3405691582: Yes, FreeBSD originally used versioned triples in the toolchain and then revised everything such that it didn't any more. However, OpenBSD always used versioned triples in the toolchain. I'm not convinced yet that unversioned triples make sense for OpenBSD based on my current understanding (though I could be in the future).

If you like, I can wrap the removed lines in SwiftCompiler.swift with `#if !os(OpenBSD)` so that builds on OpenBSD work and won't affect existing behavior, but that seems slightly off.

etcwilde: There's the module file name and the target triple encoded into the module.

The swiftmodule filenames aren't supposed to have a version number. When you run `swiftc --print-target-info` on OpenBSD, what do you see for the module triple? That's the ground-truth of what the frontend will look for.

The target triple encoded into the module is more nuanced based on what the platform ABI story looks like.

3405691582: The module triple is versioned, `aarch64-unknown-openbsd7.8` for example. 

For completeness, here's the target sub-object:

```
...
  "target": {
    "triple": "aarch64-unknown-openbsd7.8",
    "unversionedTriple": "aarch64-unknown-openbsd",
    "moduleTriple": "aarch64-unknown-openbsd7.8",
    "platform": "openbsd",
    "arch": "aarch64",
    "pointerWidthInBits": 64,
    "pointerWidthInBytes": 8,
    "compatibilityLibraries": [ ],
    "openbsdBTCFIEnabled": false,
    "librariesRequireRPath": false
  },
...
```

If the module filename is `aarch64-unknown-openbsd.swiftmodule`, then the frontend won't use it and the build fails as you point out. If the module filename is `aarch64-unknown-openbsd7.8.swiftmodule`, then everything is fine.

etcwilde: Okay, yeah, the frontend is looking for a versioned triple then. I guess the question is, do we want to make this change here, or update `swift::getTargetSpecificModuleTriple` to look for an un-versioned openBSD triple for the module? We've generally recommended un-versioned triples since availability is encoded into the module so the module could be good for multiple OS versions (depending on the ABI stability guarantees of that OS).
Keeping the version number would make OpenBSD different from all of the other platforms too.

3405691582: Typically userland and kernel are treated more of a unit on the BSDs to be updated together between versions, though I don't know offhand whether Free is special or not and whether there is any "official" promise of backward compatibility, which is why I've been reticent to think about using unversioned triples in the past (though in practice, sometimes one point release after a system upgrade is probably fine but possibly janky, especially if there are library differences, etc.).

jakepetroules: Maybe we should be universally consistent about the swiftmodule filename format, and if there are concerns about OpenBSD version compatibility, we handle that at a different layer, e.g. in the compiler or something?

I think I'd prefer that if there is not strong opposition from either of you, who probably have a better handle on the compiler side of things than I do.

3405691582: If we want to fix it on the compiler side, so long as it looked for an unversioned triple first and then a versioned one as a fallback, then that would probably work for me, since there are possibly other side effects where changing behavior might break things elsewhere.

etcwilde: The base FreeBSD system is ABI stable within a major version. The user-installed packages (libxml2, I'm looking at you) can change on a whim.

There might be some issues with build script mixing pieces of installations that it shouldn't, but so long as the new compiler that built the stdlib and modules is used to consume them, migrating shouldn't be too bad.

Just switching should be a matter of having FreeBSD and OpenBSD do the same thing here: 

```
  if (triple.isOSFreeBSD()) {
    return swift::getUnversionedTriple(triple);
  }

  if (triple.isOSOpenBSD()) {
    StringRef arch = swift::getMajorArchitectureName(triple);
    return llvm::Triple(arch, triple.getVendorName(), triple.getOSName());
  }
```

https://github.com/swiftlang/swift/blob/9d6a4f17bc864c8232b2d3d9a9b0bf93fc89738c/lib/Basic/Platform.cpp#L466-L473

Falling back on the old spelling will need changes in a handful of places. There's `ModuleIntefaceLoader.cpp` for the textual swift interface files, `ModuleLoader.cpp` for the cross-imports, and `SerializedModuleLoader.cpp` for the binary swiftmodule files:

https://github.com/swiftlang/swift/blob/9d6a4f17bc864c8232b2d3d9a9b0bf93fc89738c/lib/Frontend/ModuleInterfaceLoader.cpp#L704
https://github.com/swiftlang/swift/blob/9d6a4f17bc864c8232b2d3d9a9b0bf93fc89738c/lib/AST/ModuleLoader.cpp#L148-L163
https://github.com/swiftlang/swift/blob/9d6a4f17bc864c8232b2d3d9a9b0bf93fc89738c/lib/Serialization/SerializedModuleLoader.cpp#L55-L67

I can't think of anything else compiler-side that needs changing. The CMake bits in the old build system will likely need changes. The new runtime build system keys off of the output from `swiftc -print-target-info -target <triple>` to get all of that information, so that should just work if the compiler is working correctly.

3405691582: Looks like adding new fallback logic in some places is going to be complex...

For now, I'll pull the controversial hunk from this pr so we can at least progress the uncontroversial parts.

3405691582:
Revised description and scope based on above discussion.

Sources/SWBGenericUnixPlatform/Plugin.swift
jakepetroules: That sounds good as a starting point, thanks.

jakepetroules (approved):
@swift-ci test

Sources/SWBGenericUnixPlatform/Plugin.swift
jakepetroules: Maybe we should change getTargetSpecificModuleTriple to unconditionally call getUnversionedTriple or similar for all platforms, and then add a triple property like allowsVersionedSwiftmoduleFallback, which OpenBSD could set to true to get the behavior of falling back to the versioned variant.

jakepetroules:
@3405691582 The `buildTargetInfo` test is failing, can you please update that?

3405691582:
> @3405691582 The `buildTargetInfo` test is failing, can you please update that?

@jakepetroules I'm having difficulty spotting a specific test, can you direct me further? The GHA checks also seem to have known issues which makes it hard to spot actual failures.

jakepetroules:
In Tests/SWBBuildServiceTests/BuildServiceTests.swift, there's a buildTargetInfo(_:) test which has an OpenBSD entry around line 81.

Currently that's like:

```swift
 .init(triple: "x86_64-unknown-openbsd", platformName: "openbsd", sdkVariant: nil, deploymentTargetSettingName: nil, deploymentTarget: nil),
```

and should be changed to something like:

```swift
 .init(triple: "x86_64-unknown-openbsd7.8", platformName: "openbsd", sdkVariant: nil, deploymentTargetSettingName: "OPENBSD_DEPLOYMENT_TARGET", deploymentTarget: "7.8"),
```

3405691582:
Ah, I couldn't see it because I needed to rebase. Fixed now; looks like there is an unrelated build break on HEAD, though.

jakepetroules:
> looks like there is an unrelated build break on HEAD, though.

If you mean the Android jobs, then just ignore that. It's fixed by #1270 but needs to land in a toolchain before the CI will turn green.

jakepetroules (approved):
This looks good to land. Thanks!

3405691582:
> > looks like there is an unrelated build break on HEAD, though.
> 
> If you mean the Android jobs, then just ignore that. It's fixed by #1270 but needs to land in a toolchain before the CI will turn green.

No, it's something about `error: type 'SWBBuildServer' does not conform to protocol 'QueueBasedMessageHandler'`.

jakepetroules:
@swift-ci test

jakepetroules:
You just need to run `swift package update`.

3405691582:
> @3405691582 while we have you here... I tried building Swift Build using your container and ran into an issue where stdatomic.h was not found via `#include <stdatomic.h>`. I had to do this prior to `swift test`:
> 
> ```
> cp /usr/lib/clang/19/include/stdatomic.h /usr/include/stdatomic.h
> ```
> 
> Any ideas?

@jakepetroules I've updated the openbsd-swift-amd64 container to 6.3.1 and this should also resolve this issue. Feel free to download and try again.

PR #1251, opened 2026-03-26
Rewrite TouchTool to use FSProxy instead of shell utilities

daveinglis:
Replace platform-specific shell commands with a cross-platform builtin task action using FSProxy.touch().

This fixes a bug on Windows where the `copy /b +,,` trick failed for directories, attempting to copy contents instead of updating timestamps.

The new TouchTaskAction uses FileManager.setAttributes() internally via FSProxy, providing consistent cross-platform behavior with better error handling and no external dependencies.

depends on: https://github.com/swiftlang/swift-foundation/pull/1855

### DIFF
--- a/Tests/SWBTaskExecutionTests/TouchTaskActionTests.swift
+++ b/Tests/SWBTaskExecutionTests/TouchTaskActionTests.swift
@@ -1,0 +1,158 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import Testing
+import SWBUtil
+import SWBCore
+import SWBTaskExecution
+import SWBTestSupport
+
+#if compiler(<6.4)
+// Swift compiler versions prior to 6.4 ship with a Foundation implementation on Windows
+// that does not properly support setting timestamps on directories via touch operations.
+// This test is disabled on Windows for those compiler versions to avoid spurious failures.
+let enableTouchDirTest = (try? ProcessInfo.processInfo.hostOperatingSystem()) != .windows
+#else
+let enableTouchDirTest = true
+#endif
+
+@Suite
+fileprivate struct TouchTaskActionTests {
+
+    @Test(.enabled(if: enableTouchDirTest))
+    func touchDirectory() async throws {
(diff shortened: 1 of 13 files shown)


Tests/SWBTaskExecutionTests/TouchTaskActionTests.swift
```diff
@@ -0,0 +1,141 @@
...
+                outputDelegate: outputDelegate
+            )
+
+            // Verify success
+            #expect(result == .succeeded)
+            #expect(outputDelegate.errors.isEmpty)
+
+            // Verify timestamp was updated (touch sets it to current time)
+            let newTimestamp = try fs.getFileTimestamp(bundlePath)
+            #expect(newTimestamp >= initialTimestamp, "Timestamp should have been updated to current time")
```
jakepetroules: This may be subject to race conditions, especially because the timestamp granularity on Windows is quite coarse on some filesystems. See FSProxy.updateTimestamp for how we've worked around this before.

daveinglis: but we are testing that the new timestamp is the same or newer then it was. Not sure I understand what race there is here.

jakepetroules: If it's >= that means it might have not changed, which doesn't prove that anything happened, right?

daveinglis: I guess, will put a similar sleep loop like what's in updateTimestamp after getting the initial time stamp

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test windows

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test macos

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test windows

daveinglis:
The nightly docker image for windows is now updated with the Foundation fix 🎉

daveinglis:
@jakepetroules @owenv are we good for this to go in?

owenv:
I think we should do a cross-repo test to make sure this doesn't break compiler CI, but if that passes I think this is good to go

daveinglis:
@swift-ci test

daveinglis:
> I think we should do a cross-repo test to make sure this doesn't break compiler CI, but if that passes I think this is good to go

https://github.com/swiftlang/swift/pull/75354#issuecomment-4352903558 passed

daveinglis:
ok I finally got a complete toolchain build pass using this PR  , https://github.com/swiftlang/swift/pull/75354#issuecomment-4379744772

owenv (approved):

PR #1258, opened 2026-03-27
Add XCBuild build setting evaluation operator for truncating to a 3 tuple version number

liyanage:
**Context and problem:**

XCBuild's build setting evaluation system supports operators that transform macro values inline (e.g. `$(VAR:upper)`, `$(VAR:standardizepath)`). However, there is currently no operator to truncate a version string to its first three components (major.minor.patch).

In Apple's build system, version numbers sometimes come in 4 or 5-tuple forms (e.g. `1.2.3.4`), but many contexts — such as bundle versions, SDK version constraints, or compatibility checks — require a standard 3-tuple semver format (`1.2.3`). Today, there's no way to perform this truncation directly within a build setting expression; you'd need to work around it with custom scripts or intermediate settings.

**What this change does:**

This adds a new `truncatedversion` build setting evaluation operator, so you can write `$(MY_VERSION:truncatedversion)` and it will return up to the first three dot-separated components of the version string. For example:

- `1.2.3.4` → `1.2.3`
- `1.2.3` → `1.2.3` (no change)
- `1.2` → `1.2` (shorter versions pass through unchanged)

This follows the same pattern as the existing operators (`upper`, `lower`, `suffix`, `standardizepath`, etc.) and includes tests covering the 4-tuple, 3-tuple, and 2-tuple cases.

**Usage Example**

```
PROJECT_VERSION = 100.4.0.0.1
PROJECT_SEMANTIC_VERSION = $(PROJECT_VERSION:truncatedversion)
// PROJECT_SEMANTIC_VERSION now 100.4.0

// for use with CFBundleVersion in Info.plist which "is a machine-readable string composed of one to three period-separated integers"
CURRENT_PROJECT_VERSION = $(PROJECT_SEMANTIC_VERSION)

// LC_LOAD_DYLIB is 32-bit: will truncate to three components in linker, but this avoids linker warning
DYLIB_CURRENT_VERSION = $(PROJECT_SEMANTIC_VERSION)
```

### DIFF
--- a/Sources/SWBMacro/MacroEvaluationProgram.swift
+++ b/Sources/SWBMacro/MacroEvaluationProgram.swift
@@ -11,6 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 public import SWBUtil
+internal import Foundation
 
 /// A “program” consisting of a sequence of instructions for building a macro evaluation result buffer.  Regular clients of macro evaluation don’t need to be aware of this class.
 final class MacroEvaluationProgram: Serializable, Sendable {
@@ -185,6 +186,7 @@ final class MacroEvaluationProgram: Serializable, Sendable {
         case suffix
         case standardizepath
         case not
+        case tripleversion
 
         /// Creates and returns a new retrieval operator with the given `name`.  Returns nil if the `name` is not a supported operator.
         init?(_ name: String) {
@@ -203,6 +205,7 @@ final class MacroEvaluationProgram: Serializable, Sendable {
             case "suffix": self = .suffix
             case "standardizepath": self = .standardizepath
             case "not": self = .not
+            case "tripleversion": self = .tripleversion
             default:
                 return nil
             }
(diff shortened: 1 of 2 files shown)


jakepetroules:
Can you provide some concrete examples of how you would use this?

liyanage:
> Can you provide some concrete examples of how you would use this?

I updated the PR description with the "Usage Example" section.

owenv (dismissed):

Sources/SWBMacro/MacroEvaluationProgram.swift
```diff
@@ -194,6 +185,7 @@ final class MacroEvaluationProgram: Serializable, Sendable {
         case suffix
         case standardizepath
         case not
+        case tripleversion
```
owenv: I think we should call this something other than tripleversion to avoid conflating it with target triples. Possibly something like truncatedversion. Other than that this change lgtm

liyanage: Good point, `tripleversion` indeed already occurs in other places in this project.

Renamed to `truncatedversion` as suggested.

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv (approved):

PR #1259, opened 2026-03-27
[172848427] Rework plug-in initialization to support a new plug-in layout.

cwakamo:
* Plug-ins are now allowed to be frameworks instead of bundles.
* Assume plug-ins are relative to the SWBBuildService library instead of the SWBBuildService bundle.
* In support of this, removed the `pluginsDirectory` parameter from the build service entry point.
* Also reworked the responsibility for plug-in registration in tests so it's part of CoreTestSupport.swift directly.

This is part of &lt;rdar://problem/172848427&gt;.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -290,7 +290,7 @@ let package = Package(
             swiftSettings: swiftSettings(languageMode: .v6)),
         .target(
             name: "SWBTestSupport",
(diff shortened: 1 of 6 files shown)


cwakamo:
@swift-ci please test

cwakamo:
@swift-ci please test

owenv (approved):

cwakamo:
@swift-ci please test

PR #1271, opened 2026-04-02
Route swift-driver incremental build diagnostics through the build message system

daveinglis:
`writeIncrementalBuildInformation()` was emitting diagnostics (e.g. "warning: next compile won't be incremental") directly to stderr via `Driver.stderrDiagnosticsHandler`. In `inProcessStatic` mode (used by swift-package-manager), this bypasses the build message system entirely, causing warnings to appear on the wrong output stream with no way for the caller to intercept, deduplicate, or properly position them relative to progress output.

Changes:
- `LibSwiftDriver.writeIncrementalBuildInformation()` now captures diagnostics emitted during serialization using a snapshot-delta of the accumulated `diagnosticsEngine.diagnostics` array and returns them as `[SWBUtil.Diagnostic]` instead of writing to stderr.
- `DiagnosticsEngine` in `createAndPlan` and `frontendCommandLine` is created with an empty handler list instead of `[stderrDiagnosticsHandler]`. All diagnostics from those paths are already captured via `diagnosticsEngine.diagnostics` and returned to callers; the stderr handler was the source of double-writes.
- `SwiftModuleDependencyGraph.cleanUp(key:)` and `cleanUpForAllKeys()` gain an optional `diagnosticsHandler` callback (default `nil`, preserving backward compatibility at all existing call sites).
- `BuildOperation` wires up the handler at both call sites to emit cleanup diagnostics via `buildOutputDelegate.emit(_:)`, routing them through the same structured build output path as all other diagnostics.

### DIFF
--- a/Sources/SWBCore/LibSwiftDriver/LibSwiftDriver.swift
+++ b/Sources/SWBCore/LibSwiftDriver/LibSwiftDriver.swift
@@ -252,20 +252,26 @@ public final class SwiftModuleDependencyGraph: SwiftGlobalExplicitDependencyGrap
     }
 
     /// Serialize incremental build state for the given key and removes its state from memory
-    public func cleanUpForAllKeys() {
+    public func cleanUpForAllKeys(diagnosticsHandler: (([SWBUtil.Diagnostic]) -> Void)?) {
         registryQueue.async {
             for driver in self.registry.values {
-                driver.writeIncrementalBuildInformation()
+                let diagnostics = driver.writeIncrementalBuildInformation()
+                if !diagnostics.isEmpty {
+                    diagnosticsHandler?(diagnostics)
+                }
             }
             self.registry.removeAll()
         }
     }
 
     /// Serialize incremental build state for the given key and removes its state from memory
-    public func cleanUp(key: String) {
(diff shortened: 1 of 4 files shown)


daveinglis:
@swift-ci test

daveinglis:
@swift-ci test

Sources/SWBCore/LibSwiftDriver/LibSwiftDriver.swift
```diff
@@ -252,20 +252,26 @@ public final class SwiftModuleDependencyGraph: SwiftGlobalExplicitDependencyGrap
     }
 
     /// Serialize incremental build state for the given key and removes its state from memory
-    public func cleanUpForAllKeys() {
+    public func cleanUpForAllKeys(diagnosticsHandler: (([SWBUtil.Diagnostic]) -> Void)? = nil) {
```
jakepetroules: suggest removing the `= nil` (and on cleanUp); I find default parameters tend to cause more trouble than they're worth unless they're things like very low level stdlib types where the defaults are much less likely to be overridden.

daveinglis: updated

daveinglis:
@swift-ci test

jakepetroules (approved):

daveinglis:
@swift-ci test

PR #1274, opened 2026-04-02
Correctly detect llbuild-cancelled tasks on Windows via STATUS_CONTROL_C_EXIT

daveinglis:
On Windows, llbuild cancels tasks via TerminateProcess(STATUS_CONTROL_C_EXIT), which the NTSTATUS parsing maps to .uncaughtSignal(0x13A). wasCanceled was unconditionally returning false on Windows, misreporting interrupted tasks as failures. Fix by recognizing STATUS_CONTROL_C_EXIT in the Windows branch.

### DIFF
--- a/Sources/SWBUtil/Process.swift
+++ b/Sources/SWBUtil/Process.swift
@@ -223,6 +223,13 @@ public enum Processes: Sendable {
         case exit(_ code: Int32)
         case uncaughtSignal(_ signal: Int32)
 
+        #if os(Windows)
+        /// Mask applied to HRESULT/NTSTATUS exit codes to extract the signal value stored in uncaughtSignal.
+        private static let ntStatusSignalMask: DWORD = 0x3FFFFFFF
+        /// NTSTATUS code used by llbuild when cancelling a process via TerminateProcess. Equivalent to Ctrl+C termination.
+        static let statusControlCExit: Int32 = Int32(bitPattern: 0xC000013A & ntStatusSignalMask)
+        #endif
+
         public init?(rawValue: Int32) {
             #if os(Windows)
             let dwExitCode = DWORD(bitPattern: rawValue)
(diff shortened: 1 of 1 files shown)


Sources/SWBUtil/Process.swift
```diff
@@ -223,6 +223,13 @@ public enum Processes: Sendable {
         case exit(_ code: Int32)
         case uncaughtSignal(_ signal: Int32)
```
jakepetroules, line 224: I think we should make uncaughtSignal ifdef'ef for !os(Windows). We recently did that for swift-subprocess, as Windows has no such concept.

This would make it easier to check for STATUS_CONTROL_C_EXIT by comparing to the value directly, rather than having to deal with bit masks.

daveinglis: just want to be clear here, but are you suggesting I ifdef all the `.uncaughtSignal` references throughout including changing the `ExitStatus.?init()` for windows? If so that becomes a rather larger change over all of swift-build doesn't it?

jakepetroules: Yes

daveinglis: Are all those changes worth it?  This does something similar to what's in swift-corelibs-foundation https://github.com/swiftlang/swift-corelibs-foundation/blob/c35c6a0cf49bdd7ef65c964be504878b1e25485b/Sources/Foundation/Process.swift#L644, just not sure what the benefit of a change like that would give us?

jakepetroules: Foundation NSTask's method is fundamentally flawed as it tries to apply Unix concepts that don't exist on Windows. That bit-masking stuff is a heuristic only, as exit codes on Windows have NO formal meaning whatsoever. In Subprocess, we moved away from that strategy at my encouragement: https://github.com/swiftlang/swift-subprocess/pull/229, and the v1.0 API will just give you a single numeric exit code on Windows.

If you want to skip the refactoring for now, I can take that on instead as part of https://github.com/swiftlang/swift-build/pull/638

daveinglis: > If you want to skip the refactoring for now, I can take that on instead as part of #638

Yes, let do that. Thanks

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test windows

owenv (approved):

daveinglis:
@swift-ci test

PR #1289, opened 2026-04-08
Track dSYM bundle directory for stale file removal

rjmansfield:
SFR only tracked the DWARF binary inside the dSYM bundle, leaving the dSYM directory and skeleton behind when switching DEBUG_INFORMATION_FORMAT from dwarf-with-dsym to dwarf.

Add additionalSFRPaths to PlannedTask so tasks can register paths for SFR tracking that aren't formal task outputs and use it to register the dSYM bundle directory.

rdar://19063463

### DIFF
--- a/Sources/SWBCore/PlannedTask.swift
+++ b/Sources/SWBCore/PlannedTask.swift
@@ -107,6 +107,9 @@ public protocol PlannedTask: AnyObject, CustomStringConvertible, Sendable, Plann
 
     /// The scope for stale file removal tracking of this task's outputs.
     var staleFileRemovalScope: StaleFileRemovalScope { get }
+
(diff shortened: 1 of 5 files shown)


rjmansfield:
@swift-ci please test

owenv:
Can we add a case for a test which first builds with something like `BUILD_VARIANTS=normal asan` and then incrementally with `BUILD_VARIANTS=normal`? There's a chance that because both create a dSYM in the same bundle, the stale file removal might cause unnecessary churn with this change. A test like `BuildOperationTests.multiArchAndVariantNeutrality` might be a decent starting point for a new test case with multiple variants

rjmansfield:
@swift-ci please test

jakepetroules (approved):

owenv (approved):

PR #1292, opened 2026-04-09
Add explicit precompiled module cache pruning

rjmansfield:
Prune stale .pcm files from ExplicitPrecompiledModules/ and SwiftExplicitPrecompiledModules/ directories using the libclang clang_ModuleCache_prune API.

rdar://173786020

### DIFF
--- a/Sources/SWBBuildSystem/BuildOperation.swift
+++ b/Sources/SWBBuildSystem/BuildOperation.swift
@@ -1148,6 +1148,10 @@ private class InProcessCommand: SWBLLBuild.ExternalCommand, SWBLLBuild.ExternalD
         self.adaptor = adaptor
     }
 
+    var outputs: [String] {
+        task.isDynamic ? task.outputPaths.map(\.str) : []
+    }
+
     func getSignature(_ command: Command) -> [UInt8] {
         let signature = action.getSignature(task, executionDelegate: adaptor.operation).bytes
         return signature
@@ -1535,6 +1539,7 @@ internal final class OperationSystemAdaptor: SWBLLBuild.BuildSystemDelegate, Act
             cleanupCompilationCache()
         }
         cleanupGlobalModuleCache()
+        pruneExplicitPrecompiledModules()
 
         await queue.sync {
             self.isCompleted = true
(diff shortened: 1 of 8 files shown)


rjmansfield:
@swift-ci please test

Sources/SWBBuildSystem/BuildOperation.swift
```diff
@@ -1635,6 +1640,64 @@ internal final class OperationSystemAdaptor: SWBLLBuild.BuildSystemDelegate, Act
...
+    //  are created but old variants are never removed.
+    /// Uses the libclang clang_ModuleCache_prune API which removes stale .pcm files based on
+    /// access time. Uses CLANG_MODULES_PRUNE_AFTER and CLANG_MODULES_PRUNE_INTERVAL.
+    func pruneExplicitPrecompiledModules() {
+        var directories = Set<Path>()
+        var pruneAfter: TimeInterval = 345600  // 4 days default
+        var pruneInterval: TimeInterval = 86400 // 1 day default
+        var libclangPath: Path?
+        for target in operation.buildDescription.targetTaskCounts.keys {
+            let targetSettings = operation.requestContext.getCachedSettings(operation.request.parameters, target: target.target)
```
owenv, line 1659: We are not guaranteed to get a cache hit here, so we may want to compute these settings during build planning and save them in the build description so we never need to do expensive work here

Sources/SWBBuildSystem/BuildOperation.swift
```diff
@@ -1635,6 +1640,64 @@ internal final class OperationSystemAdaptor: SWBLLBuild.BuildSystemDelegate, Act
...
+            return
+        }
+
+        for dirPath in existingDirectories {
+            let signatureCtx = InsecureHashContext()
+            signatureCtx.add(string: "PruneExplicitPrecompiledModules")
+            signatureCtx.add(string: dirPath.str)
+            let signature = signatureCtx.signature
+
+            withActivity(ruleInfo: "PruneExplicitPrecompiledModules \(dirPath.str)", executionDescription: "Prune stale explicit modules at \(dirPath.str)", signature: signature, target: nil, parentActivity: nil) { _ in
```
owenv, line 1694: If we keep the settings computation, we should move that inside the withActivity block here so we're accounting for the time it takes

Sources/SWBBuildSystem/BuildOperation.swift
```diff
@@ -1144,6 +1144,10 @@ private class InProcessCommand: SWBLLBuild.ExternalCommand, SWBLLBuild.ExternalD
         self.adaptor = adaptor
     }
 
+    var outputs: [String] {
```
owenv, line 1151: Is anything using this right now?

rjmansfield: Yes, it's used in the PrecompileModule and SwiftDriverJob tasks so pcm paths can be passed to llbuild so it can track them. IIUC, the only mechanism for dynamic tasks to pass output paths to llbuild is through ExternalCommand.outputs, so without this llbuild can’t detect when the pcm was deleted.

Sources/SWBBuildSystem/BuildOperation.swift
rjmansfield: cleanupGlobalModuleCache() which is called immediately before pruneExplicitPrecompiledModules() already uses getCachedSettings so I think we should always hit the cache here.  If you want this moved to the build description, I should probably also do the same for cleanupGlobalModuleCache(). LMK what you prefer.

owenv (approved):

rjmansfield:
@swift-ci please test

rjmansfield:
@swift-ci please test

PR #1295, opened 2026-04-09
Ensure we look through PackageProductTargets when computing artifact bundle info

owenv:
Ensure artifact bundle info computation accounts for both regular and package product targets. Resolves an issue building swift-temporal-sdk on linux where search paths were lost when loaded by a transitive dependency through a package product.


owenv:
@swift-ci test

jakepetroules (approved):

PR #1298, opened 2026-04-10
Cache ConfiguredTarget.guid and replace linear scan in XCFrameworkContext

### LINKED ISSUE #1145: ConfiguredTarget.guid computed property causes O(n²) task generation for large projects
## Summary

`ConfiguredTarget.guid` is a computed property that performs dictionary sorting and string joining on every access. During parallel task generation in `BuildPlan.init`, it is called thousands of times across multiple threads, dominating CPU time for large projects. Caching it as a stored property (plus a secondary index in `XCFrameworkContext`) reduced a 2100-target index build from **7.07 minutes to 2.77 minutes wall clock** (61% reduction).

## Root cause

`ConfiguredTarget.guid` (in `SWBCore/ConfiguredTarget.swift`) is a computed property that:

1. Sorts `parameters.overrides` dictionary by key — O(n log n) string comparisons
2. Filters to SDKROOT/SDK_VARIANT entries
3. Joins strings into the final GUID

```swift
public var guid: GUID {
    var parameters: [String] = self.parameters.overrides.sorted(byKey: <).compactMap {
        if ["SDKROOT", "SDK_VARIANT"].contains($0.key) {
            return "\($0.key):\($0.value)"
        } else { return nil }
    }
    // ...
    return .init(id: ["target", target.name, target.guid, parameters.joined(separator: ":")].joined(separator: "-"))
}
```

Since `ConfiguredTarget` is a `final class` with all `let` properties, the guid value is deterministic — but it's recomputed on every access.

### Call amplification

`TargetOrderTaskProducer.generateTasks()` calls `createStartTask` 5 times per target (begin, start compiling, start scanning, start linking, start immediate). Each of these calls `XCFrameworkContext.outputFiles(for:)`, which does a **linear scan** over all copy configurations comparing `.guid` on every key:

```swift
func outputFiles(for target: ConfiguredTarget) -> [Path] {
    return state.withLock { state in
        return state.copyConfigurations.flatMap { (key, config) in 
            key.guid == target.guid ? config.outputs : [] 
        }
    }
}
```

For T targets and C configurations, this is **5 × T × (C + 1)** guid computations during the parallel task generation phase. With 2100 targets running across 7-8 threads, the Instruments trace showed:

- `TargetOrderTaskProducer.generateTasks`: **293s CPU** across all threads (virtually all in guid/outputFiles)
- ARC retain/release: **159s** (30% of total) — driven by string/dictionary value copying in guid computation
- String operations: **63s** (12%) — from `joined(separator:)` and string comparisons in sorting

## Fix

Two changes, both minimal:

**1. Cache `ConfiguredTarget.guid` as a stored property**

Extract the computation to a static helper, compute once in `init`, store as `private let _guid`:

```swift
private let _guid: GUID

public init(parameters: BuildParameters, target: Target, specializeGuidForActiveRunDestination: Bool = false) {
    self.parameters = parameters
    self.target = target
    self.specializeGuidForActiveRunDestination = specializeGuidForActiveRunDestination
    self._guid = Self.computeGuid(parameters: parameters, target: target, 
                                   specializeGuidForActiveRunDestination: specializeGuidForActiveRunDestination)
}

public var guid: GUID { _guid }
```

A `CodingKeys` enum excludes `_guid` from `Encodable` auto-synthesis. The deserialization `init(from:)` also computes and stores the guid.

**2. Build a guid-keyed index in `XCFrameworkContext`**

Add a `[ConfiguredTarget.GUID: [Path]]` secondary index built during `freeze()`, replacing the linear scan with a dictionary lookup:

```swift
func freeze() {
    state.withLock { state in
        var index: [ConfiguredTarget.GUID: [Path]] = [:]
        for (key, config) in state.copyConfigurations {
            index[key.guid, default: []].append(contentsOf: config.outputs)
        }
        state.outputsByGuid = index
        state.isFrozen = true
    }
}

func outputFiles(for target: ConfiguredTarget) -> [Path] {
    state.withLock { state in
        state.outputsByGuid?[target.guid] ?? []
    }
}
```

Together this transforms **5 × T × C × O(n log n)** into **5 × T × O(1)**.

## Impact

Measured on a 2100-target monorepo (iOS app with ~209,000 index build tasks):

| Metric | Before | After | Change |
|--------|--------|-------|--------|
| Wall clock (index build) | 7.07 min | 2.77 min | **-61%** |
| TargetOrderTaskProducer.generateTasks | 293s CPU | 411ms CPU | **712× faster** |
| ARC retain/release (flat profile) | 159s (30%) | Not in top 30 | Eliminated |
| String operations (flat profile) | 63s (12%) | Not in top 30 | Eliminated |

The guid caching eliminates the dictionary sorting and string joining that was the root cause of the ARC and string overhead across all task generation threads.

## Files

- `Sources/SWBCore/ConfiguredTarget.swift` — cache guid as stored property
- `Sources/SWBTaskConstruction/XCFrameworkContext.swift` — guid-keyed index + O(1) lookup

## Environment

- macOS 26.3, Xcode 26.4 beta
- 2100 targets, ~209,000 tasks in index build description
- Apple M3 Max, 14 threads available
- Profiled with Instruments Time Profiler

tosh7:
## Summary

Two minimal optimizations targeting the task generation hot path identified in #1145:
the `ConfiguredTarget.guid` computed property and the linear scan in `XCFrameworkContext.outputFiles(for:)`.
For a 2100-target index build, these eliminate ~293s of CPU spent in `TargetOrderTaskProducer.generateTasks`.

Resolves #1145.

## Background

`ConfiguredTarget.guid` was a computed property that sorted the overrides
dictionary and joined strings on every access. During parallel task generation
in `BuildPlan.init`, it is called thousands of times across multiple threads,
dominating CPU time for large projects. In addition, `XCFrameworkContext.outputFiles(for:)`
performed a linear scan over all `copyConfigurations` and called `.guid` on every key,
multiplying the cost.

For a 2100-target index build, `TargetOrderTaskProducer.generateTasks` was
spending ~293s of CPU across all threads (almost entirely in guid/outputFiles),
with ARC retain/release accounting for ~30% of the flat profile due to repeated
dictionary/string copies in guid computation.

## Changes

Two minimal, independent changes:

1. **Cache `ConfiguredTarget.guid` as a stored property** (`SWBCore/ConfiguredTarget.swift`)
   - Extract the computation to a `private static func computeGUID(...)` helper.
   - Compute once in `init` and `init(from:)` and store as `public let guid: GUID`.
   - Add a `CodingKeys` enum to exclude the new stored property from `Encodable` auto-synthesis.

2. **Replace linear scan with dictionary lookup in `XCFrameworkContext.outputFiles(for:)`** (`SWBTaskConstruction/XCFrameworkContext.swift`)
   - Build a `[ConfiguredTarget.GUID: [Path]]` index in `freeze()` once `copyConfigurations` is finalized.
   - `outputFiles(for:)` now performs an O(1) lookup instead of an O(n) scan, while keeping the existing `precondition(state.isFrozen)` invariant.

Together this transforms the hot path from `5 × T × C × O(n log n)` to `5 × T × O(1)`.

## Testing

- `swift test --filter SWBCoreTests` — 524 tests pass
- `swift test --filter SWBTaskConstructionTests` — 741 tests pass
- `swift test --filter SWBProtocolTests` — 16 tests pass (serialization compatibility verified)

### DIFF
--- a/Sources/SWBCore/ConfiguredTarget.swift
+++ b/Sources/SWBCore/ConfiguredTarget.swift
@@ -25,6 +25,33 @@ public final class ConfiguredTarget: Hashable, CustomStringConvertible, Serializ
     /// This is utilized by the index build description to differentiate targets that have `SDKROOT` set (it's not 'auto'), so they don't need an override for SDKROOT and SDK_VARIANT, but are configured for multiple platforms and their `guid` needs be different.
     @_spi(BuildDescriptionSignatureComponents) public let specializeGuidForActiveRunDestination: Bool
 
+    public struct GUID: Hashable, Sendable, Comparable, CustomStringConvertible, Encodable {
+        public let stringValue: String
+
+        public init(id: String) {
+            self.stringValue = id
+        }
+
+        public static func < (lhs: GUID, rhs: GUID) -> Bool {
+            return lhs.stringValue < rhs.stringValue
+        }
+
+        public var description: String {
+            stringValue
+        }
+    }
+
+    /// An identifier that contributes to all task identifiers in order to disambiguate them.
+    ///
+    /// Note that making significant changes to the formatting of the `guid` will likely require updating quite a few tests.
+    public let guid: GUID
+
+    private enum CodingKeys: String, CodingKey {
+        case parameters
+        case target
+        case specializeGuidForActiveRunDestination
+    }
+
     /// Create a new configured target instance.
     /// - parameter parameters: The build parameters the target is configured with.
     /// - parameter target: The target to be configured.
@@ -33,6 +60,7 @@ public final class ConfiguredTarget: Hashable, CustomStringConvertible, Serializ
         self.parameters = parameters
         self.target = target
         self.specializeGuidForActiveRunDestination = specializeGuidForActiveRunDestination
+        self.guid = Self.computeGUID(parameters: parameters, target: target, specializeGuidForActiveRunDestination: specializeGuidForActiveRunDestination)
     }
 
     public func replacingTarget(_ target: Target) -> ConfiguredTarget {
@@ -80,27 +108,8 @@ public final class ConfiguredTarget: Hashable, CustomStringConvertible, Serializ
         return "<\(type(of: self)) \(string)>"
     }
 
-    public struct GUID: Hashable, Sendable, Comparable, CustomStringConvertible {
-        public let stringValue: String
-
-        public init(id: String) {
-            self.stringValue = id
-        }
-
-        public static func < (lhs: GUID, rhs: GUID) -> Bool {
-            return lhs.stringValue < rhs.stringValue
-        }
-
-        public var description: String {
-            stringValue
-        }
-    }
-
-    /// An identifier that contributes to all task identifiers in order to disambiguate them.
-    ///
-    /// Note that making significant changes to the formatting of the `guid` will likely require updating quite a few tests.
-    public var guid: GUID {
-        var parameters: [String] = self.parameters.overrides.sorted(byKey: <).compactMap {
+    private static func computeGUID(parameters: BuildParameters, target: Target, specializeGuidForActiveRunDestination: Bool) -> GUID {
+        var components: [String] = parameters.overrides.sorted(byKey: <).compactMap {
             if ["SDKROOT", "SDK_VARIANT"].contains($0.key) {
                 return "\($0.key):\($0.value)"
             } else {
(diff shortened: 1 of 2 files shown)


owenv (commented):
one comment but otherwise this lgtm, but in the future please open separate PRs for unrelated changes

owenv:
@swift-ci test

owenv (approved):

PR #1299, opened 2026-04-10
Fix xcfilelist changes not invalidating build description

rjmansfield:
BuildRuleTaskProducer is created transiently and discarded after task generation so its accessedPaths were getting lost. Move accessedPaths to TaskProducerContext so all producers contribute to invalidation paths.

rdar://117046957

### DIFF
--- a/Sources/SWBTaskConstruction/TaskProducers/TaskProducer.swift
+++ b/Sources/SWBTaskConstruction/TaskProducers/TaskProducer.swift
@@ -111,7 +111,19 @@ public class TaskProducerContext: StaleFileRemovalContext, BuildFileResolution
     /// Whether a task planned by this producer has requested frontend command line emission.
     var emitFrontendCommandLines: Bool
 
+    /// Registers a path accessed during task construction.
+    func access(path: Path) {
+        state.withLock { $0.accessedPaths.insert(path) }
+    }
+
+    /// Paths accessed during task construction that invalidate the build description.
+    /// On the context so ephemeral producers are included.
+    var accessedPaths: Set<Path> {
+        state.withLock { $0.accessedPaths }
+    }
+
     private struct State: Sendable {
+        fileprivate var accessedPaths = Set<Path>()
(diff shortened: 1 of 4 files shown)


rjmansfield:
@swift-ci please test

Sources/SWBTaskConstruction/TaskProducers/TaskProducer.swift
```diff
@@ -111,6 +111,15 @@ public class TaskProducerContext: StaleFileRemovalContext, BuildFileResolution
     /// Whether a task planned by this producer has requested frontend command line emission.
     var emitFrontendCommandLines: Bool
 
+    /// Paths accessed during task construction that invalidate the build description.
+    /// On the context so ephemeral producers are included.
+    private(set) var accessedPaths = Set<Path>()
```
owenv: The TaskProducerContext is shared among multiple TaskProducers, so this will need to move into the State structure below or otherwise guard against concurrent access. 

BuildPlan.swift also has this:
```
// Now that tasks have been generated, aggregate the invalidation paths
        let invalidationPaths = Set(producersToEvaluate.flatMap { $0.producer.invalidationPaths })
```

Which will include each path multiple times when they're all in the same set

owenv (changes requested):

Sources/SWBTaskConstruction/TaskProducers/TaskProducer.swift
rjmansfield: Good catch, will fix. 

Should the invalidation path collection be moved after deferred producers complete (after line 221)? It looks like currently no deferred producers call access(path:) but it could be a potential problem if they ever do.

owenv (approved):

Sources/SWBTaskConstruction/TaskProducers/TaskProducer.swift
owenv: It probably wouldn't hurt since we plausibly might start relying on that in the future

rjmansfield:
@swift-ci please test

PR #1301, opened 2026-04-10
Emit build error when resource files generate sources for target without a Compile Sources phase.

brooke-callahan:
Problem:
* Targets with source-generating file types (e.g., `.xcstrings`) in the Resources phase but no Compile Sources phase silently dropped the generated sources, leading to confusing build failures or missing symbols at runtime.

Changes:
* Introduced `SourceGeneratingFileType` and `MissingSourcesPhaseBehavior` in `InputFileGroupingStrategyExtension.swift`.
  * Extension points can now declare whether a missing Sources phase should silently keep the file in Resources (`.keepInResources`) or emit a build error (`.error`).
* Renamed `fileTypesCompilingToSwiftSources` to `fileTypesProducingGeneratedSources` and threaded a `MacroEvaluationScope` through the call chain.
  * The scope is needed for extensions that conditionally produce sources based on build settings.
* In `ResourcesTaskProducer`, `sourceGenerationInputFiles` is now called even when there is no Sources phase.
  * This ensures file types opting into `.error` behavior produce a diagnostic instead of being silently ignored.
* Added a default implementation of `sourceGeneratingFileTypes` that bridges to `fileTypesProducingGeneratedSources` so existing conformers don't break.

### DIFF
--- a/Sources/SWBTaskConstruction/TaskProducers/BuildPhaseTaskProducers/FilesBasedBuildPhaseTaskProducer.swift
+++ b/Sources/SWBTaskConstruction/TaskProducers/BuildPhaseTaskProducers/FilesBasedBuildPhaseTaskProducer.swift
@@ -328,9 +328,18 @@ extension PluginManager {
     func fileTypesProducingGeneratedSources(scope: MacroEvaluationScope) -> [String] {
         var compileToSwiftFileTypes : [String] = []
         for groupingStragegyExtensions in extensions(of: InputFileGroupingStrategyExtensionPoint.self) {
-            compileToSwiftFileTypes.append(contentsOf: groupingStragegyExtensions.fileTypesCompilingToSwiftSources(scope: scope))
+            compileToSwiftFileTypes.append(contentsOf: groupingStragegyExtensions.fileTypesProducingGeneratedSources(scope: scope))
         }
         return compileToSwiftFileTypes
+    }
+
+    /// Returns source-generating file types with metadata about missing Sources phase behavior.
+    func sourceGeneratingFileTypes(scope: MacroEvaluationScope) -> [SourceGeneratingFileType] {
+        var result: [SourceGeneratingFileType] = []
+        for extension_ in extensions(of: InputFileGroupingStrategyExtensionPoint.self) {
+            result.append(contentsOf: extension_.sourceGeneratingFileTypes(scope: scope))
+        }
+        return result
     }
 }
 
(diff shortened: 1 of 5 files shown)


Sources/SWBTaskConstruction/TaskProducers/BuildPhaseTaskProducers/FilesBasedBuildPhaseTaskProducer.swift
```diff
@@ -811,6 +822,14 @@ package class FilesBasedBuildPhaseTaskProducerBase: PhasedTaskProducer {
                 continue
             }
 
+            // Emit a build error if this file type requires a Sources phase but none exists.
+            if !hasSourcesPhase {
+                if let matchingType = sourceGeneratingTypes.first(where: { fileRef.fileType.conformsTo(identifier: $0.identifier) }),
```
jakepetroules: I think this can be expressed more directly as:

```swift
if sourceGeneratingTypes.contains(where: { $0.missingSourcesPhaseBehavior == .error && fileRef.fileType.conformsTo(identifier: $0.identifier) })
```

"first" is generally a code smell for things which are not ordering sensitive

brooke-callahan: How do you feel about:
```
                let shouldErrorForMissingSourcesPhase = sourceGeneratingTypes.contains(where: {
                    $0.missingSourcesPhaseBehavior == .error && fileRef.fileType.conformsTo(identifier: $0.identifier)
                })

                if shouldErrorForMissingSourcesPhase {
```

brooke-callahan: Or am I going the wrong direction?

owenv (approved):

Sources/SWBTaskConstruction/TaskProducers/BuildPhaseTaskProducers/FilesBasedBuildPhaseTaskProducer.swift
jakepetroules: Yes, that's fine, that's the same as what I suggested minus formatting. The main thing was to use `contains(where:)` instead of `first(where:)` to make it obvious there's no ordering sensitivity in play.

jakepetroules (approved):

jakepetroules:
@swift-ci test

PR #1303, opened 2026-04-13
Resolve relative derivedDataPath to absolute in swbuild build

### LINKED ISSUE #944: `swbuild build` cannot handle a relative path pointing to DerivedData
I found `swbuild build` (v6.2, release/6.2) causes an error when it receives a relative path that points to DerivedData.

```sh
swbuild build ./relative/path/to/pif \
    --configuration <configuration>
    --derivedDataPath ./relative/path/to/DerivedData \
    --buildParametersFile ./relative/path/to/json \
    --target <target>
```

Error message:

> unable to write module session file at './relative/path/to/DerivedData/ModuleCache.noindex/Session.modulevalidation': unknown error

# Investigation

## Cause of the Error

The issue seems to originate here:

https://github.com/swiftlang/swift-build/blob/3c541cc6c6773dbff995707c0860bf51a3cc1f7a/Sources/SWBBuildSystem/BuildOperation.swift#L765-L774

https://github.com/swiftlang/swift-build/blob/3c541cc6c6773dbff995707c0860bf51a3cc1f7a/Sources/SWBUtil/FSProxy.swift#L409-L412

## Note

* `swbuild build` works correctly when I pass an absolute path to the same place.
* Other parameters accept both relative and absolute paths.

tosh7:
## Summary

Resolve relative `--derivedDataPath` to an absolute path in `swbuild build` and `prepareForIndex` commands, consistent with how `containerPath` is already handled.

Resolves #944.

## Background

`swbuild build --derivedDataPath ./relative` fails because the relative path is passed through to the build pipeline without resolution. The `containerPath` parameter already resolves relative paths via `makeAbsolute(relativeTo:)` (with an explicit comment explaining why), but `derivedDataPath` was missing the same treatment.

## Changes

- Apply `makeAbsolute(relativeTo: baseDirectory) + normalize()` to `derivedDataPath` before calling `setDerivedDataPath`, in both `SWBServiceConsoleBuildCommand` and `SWBServiceConsolePrepareForIndexCommand`
- Add a test `buildCommandWithPIFRelativeDerivedDataPath` verifying that a relative `--derivedDataPath` produces a successful build

## Design discussion

I considered consolidating the absolute path conversion into a single step. Currently the flow is:

1. `makeAbsolute` + `normalize` → resolve relative path to absolute
2. `setDerivedDataPath(resolved)` → store in `arenaInfo` (generates 6 derived paths)
3. `AbsolutePath(validating:)` → read back from `arenaInfo` and validate

Step 3 re-validates a value we just resolved in step 1, which is redundant when `--derivedDataPath` is given. However, step 3 also covers the case where `derivedDataPath` comes from `--buildRequestFile` (without `--derivedDataPath`), so removing it would require splitting the logic into two explicit branches.

I opted for the minimal change (adding `makeAbsolute` before `setDerivedDataPath`) to keep the diff small and preserve the existing error handling. The redundant validation is harmless — it will always succeed for paths we resolved — but I wanted to flag this for reviewers.

Would it be worth refactoring the two-step conversion into a single unified path in a follow-up, or is the current approach preferred?

## Testing

- Added `buildCommandWithPIFRelativeDerivedDataPath` test

### DIFF
--- a/Sources/SwiftBuild/ConsoleCommands/SWBServiceConsoleBuildCommand.swift
+++ b/Sources/SwiftBuild/ConsoleCommands/SWBServiceConsoleBuildCommand.swift
@@ -206,7 +206,7 @@ class SWBServiceConsoleBuildCommand: SWBServiceConsoleCommand {
             // arena info to both the request-global build parameters as well as the target-specific
             // build parameters, since they may have been deserialized from the build request file above,
             // overwriting the build parameters we set up earlier in this method.
-            if let path = derivedDataPath {
+            if let path = derivedDataPath?.makeAbsolute(relativeTo: Path(baseDirectory.pathString))?.normalize() ?? derivedDataPath {
                 request.setDerivedDataPath(path)
             }
 
@@ -337,7 +337,7 @@ class SWBServiceConsolePrepareForIndexCommand: SWBServiceConsoleCommand {
             request.continueBuildingAfterErrors = true
 
             // Override the arena, if requested.
-            if let path = derivedDataPath {
+            if let path = derivedDataPath?.makeAbsolute(relativeTo: Path(baseDirectory.pathString))?.normalize() ?? derivedDataPath {
                 request.setDerivedDataPath(path)
             }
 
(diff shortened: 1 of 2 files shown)


jakepetroules (approved):

owenv (approved):

owenv:
@swift-ci test

PR #1304, opened 2026-04-13
Simplify target==destination condition for SDK resolution

shrudge:
Fixes #1267.

Context:
This removes redundant hardcoded platform checks in `addRunDestinationSettingsPlatformSDK` when `targetPlatform === destinationPlatform`. This cleanup naturally restores support for Swift SDK destinations that was inadvertently removed during the recent `BuildTarget `refactor (commit 385edd5 ), while maintaining correct override behavior for physical devices.

### DIFF
--- a/Sources/SWBCore/Settings/Settings.swift
+++ b/Sources/SWBCore/Settings/Settings.swift
@@ -3718,7 +3718,6 @@ private class SettingsBuilder: ProjectMatchLookup {
         }
 
         let destinationPlatformIsMacOS = destinationPlatform.name == "macosx"
-        let destinationPlatformIsLinux = destinationPlatform.name == "linux"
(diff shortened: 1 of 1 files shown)


owenv:
Closing for now as the PR description doesn't seem to match the changes here. Feel free to reopen when this is ready for review

shrudge:
@owenv, so sorry for the confusing phrasing.
I did not mean to imply I was adding the test file to this commit. I noticed the disabled tag was already removed on main. When I wrote "re-enables", what I meant was that the fix in `addRunDestinationSettingsPlatformSDK` is what actually makes the `hostToolBuildsForHostPlatformWithSwiftSDK` test pass again, since the recent refactor broke that Swift SDK fallback behavior. Poor choice of words on my part.

I updated the PR description so it is much clearer, could you please reopen this and run tests when you have a second? Thanks alot...

PR #1306, opened 2026-04-13
Add regression test for rdar://133321635

rjmansfield:
The radar reports that a Run Script phase whose input depends on a copied symlinked directory does not rerun when the symlink target is modified. This was previously fixed by f5001861.

This adds a test to verify the incremental behaviour: after modifying a file behind the symlink, both the Copy and downstream Script phases rerun.


rjmansfield:
@swift-ci please test

owenv (approved):

jakepetroules (approved):

PR #1308, opened 2026-04-14
Skip test on Windows which relies on reliable atomic writes

owenv:
This is failing in some Windows CI because swift-driver fails to write to files atomically due to https://github.com/swiftlang/swift-foundation/issues/1507


owenv:
@swift-ci test

jakepetroules (approved):

PR #1310, opened 2026-04-15
Support swift build --experimental-xcframeworks-on-linux

vsarunas:
Allow use of `--experimental-xcframeworks-on-linux` flag as per https://github.com/swiftlang/swift-build/issues/1250

These changes allows us to compile our code base on Linux using `--build-system swiftbuild`.

### DIFF
--- a/Sources/SWBCore/XCFramework.swift
+++ b/Sources/SWBCore/XCFramework.swift
@@ -370,7 +370,7 @@ public struct XCFramework: Hashable, Sendable {
             return false
         }
         // SDK names from Info.plist.
-        let appleSDKNames: Set<String> = [
+        let sdkCanonicalNames = [
             "macosx",
             "iphoneos", "iphonesimulator",
             "appletvos", "appletvsimulator",
@@ -378,7 +378,7 @@ public struct XCFramework: Hashable, Sendable {
             "xros", "xrsimulator",
             "driverkit",
         ]
-        return !appleSDKNames.contains(platform)
+        return !sdkCanonicalNames.contains(platform)
     }
 
     /// Searches the `libraries` based on the current SDK being used.
--- a/Sources/SWBCore/XCFramework.swift
+++ b/Sources/SWBCore/XCFramework.swift
@@ -363,41 +363,42 @@ public struct XCFramework: Hashable, Sendable {
     /// The minimum XCFramework version required to support mergeable metadata.
     @_spi(Testing) public static let mergeableMetadataVersion = Version(1, 1)
 
+    static func hasPerArchSlices(_ platform: String) -> Bool {
+        return platform == "linux"
+    }
+
     /// Searches the `libraries` based on the current SDK being used.
-    public func findLibrary(sdk: SDK?, sdkVariant: SDKVariant?) -> XCFramework.Library? {
+    public func findLibrary(sdk: SDK?, sdkVariant: SDKVariant?, architectures: [String] = []) -> XCFramework.Library? {
         // Lookup the LC_BUILD_VERSION information as that it is how xcframeworks platform and variant values are defined.
         guard let platformName = sdkVariant?.llvmTargetTripleSys else {
             return nil
         }
-        return findLibrary(platform: platformName, platformVariant: sdkVariant?.llvmTargetTripleEnvironment ?? "")
+        return findLibrary(platform: platformName, platformVariant: sdkVariant?.llvmTargetTripleEnvironment ?? "", architectures: architectures)
     }
 
     /// Given a platform and the variant, attempt to find an library within the XCFramework that can be used.
     public func findLibrary(platform: String, platformVariant: String = "", architectures: [String] = []) -> XCFramework.Library? {
-        #if canImport(Darwin)
-        return self.libraries.filter { lib in
-            // Due to the fact that macro evaluation of empty settings returns empty strings, there is no meaningful distinction between nil and empty here.
-            lib.supportedPlatform == platform && (lib.platformVariant ?? "") == platformVariant
-        }.first
-        #else
-        // Linux ships per-arch XCFramework slices and typically omits SupportedPlatformVariant, so for
-        // that platform default archs to the host and allow a no-variant fallback.
-        let platformMatches = libraries.filter { $0.supportedPlatform == platform }
-        var effectiveArchs = architectures
-        if platform == "linux", effectiveArchs.isEmpty, let hostArch = Architecture.hostStringValue {
-            effectiveArchs = [hostArch]
+        guard Self.hasPerArchSlices(platform) else {
+            return self.libraries.filter { lib in
+                // Due to the fact that macro evaluation of empty settings returns empty strings, there is no meaningful distinction between nil and empty here.
+                lib.supportedPlatform == platform && (lib.platformVariant ?? "") == platformVariant
+            }.first
         }
-        let allowNoVariantFallback = platform == "linux" && !platformVariant.isEmpty
-        let variants = allowNoVariantFallback ? [platformVariant, ""] : [platformVariant]
+        // Linux ships per-arch XCFramework slices and typically omits SupportedPlatformVariant, so for
+        // that platform require exactly one target arch and allow a no-variant fallback.
+        guard architectures.count == 1 else { return nil }
+        let targetArch = architectures[0]
+        let variants: [String] = platformVariant.isEmpty ? [""] : [platformVariant, ""]
         for variant in variants {
-            if let match = platformMatches.first(where: {
-                ($0.platformVariant ?? "") == variant && effectiveArchs.allSatisfy($0.supportedArchitectures.contains)
+            if let match = libraries.first(where: {
+                $0.supportedPlatform == platform
+                    && ($0.platformVariant ?? "") == variant
+                    && $0.supportedArchitectures.contains(targetArch)
             }) {
                 return match
             }
         }
         return nil
-        #endif
     }
 }
 
@@ -409,7 +410,7 @@ extension XCFramework {
         let architectures: [String]
 
         // Per-arch slice platforms — sibling slices must not collide on `(platform, variant)` alone.
-        private var usesPerArchSlices: Bool { platform == "linux" }
+        private var usesPerArchSlices: Bool { XCFramework.hasPerArchSlices(platform) }
 
         func hash(into hasher: inout Hasher) {
             // identifier is only used for tracking the offending library
--- a/Sources/SWBCore/XCFramework.swift
+++ b/Sources/SWBCore/XCFramework.swift
@@ -363,8 +363,22 @@ public struct XCFramework: Hashable, Sendable {
     /// The minimum XCFramework version required to support mergeable metadata.
     @_spi(Testing) public static let mergeableMetadataVersion = Version(1, 1)
 
+    /// Whether the platform ships one slice per architecture (no fat binaries).
     static func hasPerArchSlices(_ platform: String) -> Bool {
-        return platform == "linux"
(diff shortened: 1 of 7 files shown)


owenv (changes requested):
As I said in the linked issue, I don't think it makes sense to take this change - we shouldn't attempt to process malformed XCFrameworks that don't contain MachO binaries, and we should ensure the compiler is prepared to enforce the appropriate stability guarantees before allowing packages to vend binary shared libraries via artifact bundle on non-Apple platforms.

Sources/SWBCore/XCFramework.swift
```diff
@@ -373,11 +373,31 @@ public struct XCFramework: Hashable, Sendable {
     }
 
     /// Given a platform and the variant, attempt to find an library within the XCFramework that can be used.
-    public func findLibrary(platform: String, platformVariant: String = "") -> XCFramework.Library? {
+    public func findLibrary(platform: String, platformVariant: String = "", architectures: [String] = []) -> XCFramework.Library? {
+        #if canImport(Darwin)
```
jakepetroules: You can't use ifdefs here because this will not work for cross-compilation.

I think the actual condition here is whether the platform uses per-arch slices? You can extract the logic in the `usesPerArchSlices` property you added to LibraryKey and re-use that here.

Sources/SWBCore/XCFramework.swift
```diff
@@ -373,11 +373,31 @@ public struct XCFramework: Hashable, Sendable {
...
         return self.libraries.filter { lib in
             // Due to the fact that macro evaluation of empty settings returns empty strings, there is no meaningful distinction between nil and empty here.
             lib.supportedPlatform == platform && (lib.platformVariant ?? "") == platformVariant
         }.first
+        #else
+        // Linux ships per-arch XCFramework slices and typically omits SupportedPlatformVariant, so for
+        // that platform default archs to the host and allow a no-variant fallback.
+        let platformMatches = libraries.filter { $0.supportedPlatform == platform }
+        var effectiveArchs = architectures
+        if platform == "linux", effectiveArchs.isEmpty, let hostArch = Architecture.hostStringValue {
```
jakepetroules: You can't use `Architecture.hostStringValue` here because this will not work for cross-compilation. `architectures` should never be an empty list - for the Linux case you can check that it contains exactly one value, and look up that value.

jakepetroules (changes requested):

Sources/SWBCore/XCFramework.swift
vsarunas: OK! ifdef removed, originally I was trying to minimise the changes required and not thread through this argument.

Sources/SWBCore/XCFramework.swift
vsarunas: Adjusted.

Sources/SWBCore/XCFramework.swift
```diff
@@ -363,21 +363,42 @@ public struct XCFramework: Hashable, Sendable {
     /// The minimum XCFramework version required to support mergeable metadata.
     @_spi(Testing) public static let mergeableMetadataVersion = Version(1, 1)
 
+    static func hasPerArchSlices(_ platform: String) -> Bool {
+        return platform == "linux"
```
jakepetroules: Please change this to:

```suggestion
        return !["macosx", "iphoneos", "iphonesimulator", "appletvos", "appletvsimulator", "watchos", "watchsimulator", "xros", "xrsimulator", "driverkit"].contains(platform)
```

It's a bit verbose, but this is more correct and protects against edge cases if we end up splitting linux into multiple platforms, which @owenv and I have talked about doing to solve some cross compilation issues.

Sources/SWBCore/XCFramework.swift
```diff
@@ -363,21 +363,42 @@ public struct XCFramework: Hashable, Sendable {
...
-            lib.supportedPlatform == platform && (lib.platformVariant ?? "") == platformVariant
-        }.first
+    public func findLibrary(platform: String, platformVariant: String = "", architectures: [String] = []) -> XCFramework.Library? {
+        guard Self.hasPerArchSlices(platform) else {
+            return self.libraries.filter { lib in
+                // Due to the fact that macro evaluation of empty settings returns empty strings, there is no meaningful distinction between nil and empty here.
+                lib.supportedPlatform == platform && (lib.platformVariant ?? "") == platformVariant
+            }.first
+        }
+        // Linux ships per-arch XCFramework slices and typically omits SupportedPlatformVariant, so for
```
jakepetroules:
```suggestion
        // Non-Apple platforms ship per-arch XCFramework slices and typically omit SupportedPlatformVariant, so for
```

Sources/SWBCore/XCFramework.swift
```diff
@@ -363,21 +363,42 @@ public struct XCFramework: Hashable, Sendable {
...
+        // that platform require exactly one target arch and allow a no-variant fallback.
```
jakepetroules:
```suggestion
        // those platforms, require exactly one target arch and allow a no-variant fallback.
```

Sources/SWBCore/XCFramework.swift
```diff
@@ -386,15 +407,25 @@ extension XCFramework {
...
+        let architectures: [String]
+
+        // Per-arch slice platforms — sibling slices must not collide on `(platform, variant)` alone.
+        private var usesPerArchSlices: Bool { XCFramework.hasPerArchSlices(platform) }
 
         func hash(into hasher: inout Hasher) {
             // identifier is only used for tracking the offending library
             hasher.combine(platform)
             hasher.combine(platformVariant)
+            if usesPerArchSlices {
```
jakepetroules: This needs to be unconditionally added to the hash; Hashable and Equatable conformances are not the right place for this sort of business logic.

Sources/SWBCore/XCFramework.swift
```diff
@@ -386,15 +407,25 @@ extension XCFramework {
...
+                hasher.combine(architectures)
+            }
         }
 
         static func == (lhs: LibraryKey, rhs: LibraryKey) -> Bool {
-            return lhs.platform == rhs.platform && lhs.platformVariant == rhs.platformVariant
+            guard lhs.platform == rhs.platform, lhs.platformVariant == rhs.platformVariant else {
+                return false
+            }
+            return lhs.usesPerArchSlices ? (lhs.architectures == rhs.architectures) : true
```
jakepetroules: This needs to be unconditionally compared; Hashable and Equatable conformances are not the right place for this sort of business logic.

Sources/SWBCore/XCFramework.swift
```diff
@@ -363,21 +363,42 @@ public struct XCFramework: Hashable, Sendable {
...
+        guard Self.hasPerArchSlices(platform) else {
+            return self.libraries.filter { lib in
+                // Due to the fact that macro evaluation of empty settings returns empty strings, there is no meaningful distinction between nil and empty here.
+                lib.supportedPlatform == platform && (lib.platformVariant ?? "") == platformVariant
+            }.first
+        }
+        // Linux ships per-arch XCFramework slices and typically omits SupportedPlatformVariant, so for
+        // that platform require exactly one target arch and allow a no-variant fallback.
+        guard architectures.count == 1 else { return nil }
+        let targetArch = architectures[0]
```
jakepetroules: You can combine the .count==1 check with this:


```suggestion
        guard let targetArch = architectures.only else { return nil }
```

jakepetroules (changes requested):
Looking better! Commented on a few more small things...

Sources/SWBCore/XCFramework.swift
vsarunas: Adjusted, even more verbose as an existing test was failing.

Sources/SWBCore/XCFramework.swift
vsarunas: Adjusted.

Sources/SWBCore/XCFramework.swift
vsarunas: Adjusted.

Sources/SWBCore/XCFramework.swift
vsarunas: Business logic moved to validate().

Sources/SWBCore/XCFramework.swift
vsarunas: Changed.

Sources/SWBCore/XCFramework.swift
vsarunas: Changed.

Sources/SWBCore/XCFramework.swift
```diff
@@ -363,21 +363,55 @@ public struct XCFramework: Hashable, Sendable {
...
     @_spi(Testing) public static let mergeableMetadataVersion = Version(1, 1)
 
+    /// Whether the platform ships one slice per architecture (no fat binaries).
+    static func hasPerArchSlices(_ platform: String) -> Bool {
+        // Triple-sys spellings (e.g. "macos") — from findLibrary.
+        if BuildVersion.Platform(platform: platform, environment: nil) != nil {
+            return false
+        }
+        // SDK names from Info.plist.
+        let sdkCanonicalNames = [
```
jakepetroules: Sorry, I forgot these platform names are in terms of the triples, not the Xcode platform names. You shouldn't check both -- just the BuildVersion.Platform code path is sufficient.

jakepetroules (changes requested):

Sources/SWBCore/XCFramework.swift
vsarunas: Simplified.

jakepetroules (approved):

jakepetroules:
@swift-ci test

vsarunas:
Ubuntu failure https://github.com/swiftlang/swift-build/actions/runs/24925300005/job/73036104365?pr=1310#step:14:43080 doesn't seem to be related to the changes:
```
◇ Test case passing 1 argument buildSystem → .swiftbuild to enableDisableCache(buildSystem:) started.
**** FAILURE EXECUTING SUBPROCESS ****
output: 
stderr: 
*** Signal 11: Backtracing from 0x7f19595ce9aa... failed ***

*** Backtracer crashed during backtracing; attempting to backtrace backtracer: ***

*** Signal 4: Backtracing from 0x7f2082b4983c... done ***

*** Program crashed: Illegal instruction at 0x00007f2082b4983c ***

Platform: x86_64 Linux (Ubuntu 22.04.5 LTS)

Thread 0 "swift-backtrace" crashed:

 0                0x00007f2082b4983c specialized ImageSource.fetchSLEB128(from:) + 220 in libswiftRuntime.so
 1 [ra]           0x00007f2082afcce7 DwarfReader.read(form:at:addressSize:isDwarf64:unit:shouldFetchIndirect:constantValue:) + 1654 in libswiftRuntime.so
 2 [ra]           0x00007f2082afea5e DwarfReader.readDieAttributes(at:unit:abbrevInfo:shouldFetchIndirect:) + 365 in libswiftRuntime.so
 3 [ra]           0x00007f2082b04907 DwarfReader._scanUnits() + 646 in libswiftRuntime.so
 4 [ra]           0x00007f2082b0068f DwarfReader._lazyInfo.getter + 46 in libswiftRuntime.so
 5 [ra]           0x00007f2082b00b11 DwarfReader.lookupInlineCallSites(at:) + 32 in libswiftRuntime.so
 6 [ra]           0x00007f2082b0f501 ElfImage.inlineCallSites(at:) + 48 in libswiftRuntime.so
```

owenv (approved):
I guess this isn't doing too much harm

PR #1311, opened 2026-04-15
Add build setting to control swiftinterface verification for installed headers

sepy97:
Gate the -no-verify-emitted-module-interface flag behind a new internal build setting SWIFT_SKIP_INSTALLED_HEADER_INTERFACE_VERIFICATION, defaulting to YES to preserve existing behavior. Setting it to NO allows .swiftinterface verification to run for frameworks that install a generated ObjC compatibility header.

rdar://174872349


sepy97:
@swift-ci please test

nkcsgexi (approved):

owenv (approved):
makes sense to me, this seems like the right approach to staging in the verification

PR #1312, opened 2026-04-16
Optimize SourcesTaskProducer by using direct identifier lookups

shrudge:
## Motivation

When querying the sourcecode.swift spec in `SourcesTaskProducer`, the `context.lookupFileType(fileName: "sourcecode.swift")` overload was being used. This incorrectly triggers an O(N) linear scan across all registered `FileTypeSpec` objects to match the .swift extension. This only succeeded by coincidence because the spec identifier's suffix matches the valid file extension.

## Modifications:

Updated 5 call sites in `SourcesTaskProducer.swift `to use `context.lookupFileType(identifier: "sourcecode.swift")` instead of fileName:.

## Result:

Replaces the O(N) file extension scan with a direct O(1) proxy dictionary lookup, eliminating redundant iterations during task construction.


jakepetroules (approved):
Great catch, thanks!

jakepetroules:
@swift-ci test

owenv (approved):

owenv:
@swift-ci test

PR #1315, opened 2026-04-16
Ensure stdout and stderr are drained concurrently to avoid deadlock

rconnell9:
The process that runs clang to check its version writes verbose (-v) output to stderr.  On some systems this can exceed the kernel buffer size.

Before this change the parent drained stdout and stderr sequentially, if the child filled the stderr pipe the child stalled waiting for the parent to drain it but the parent is waiting for stdout which would never arrive.

### DIFF
--- a/Sources/SWBUtil/Process+Async.swift
+++ b/Sources/SWBUtil/Process+Async.swift
@@ -45,21 +45,10 @@ extension Process {
     ///
     /// - note: This method will mutate the `standardOutput` or `standardError` property of the Process object, replacing any existing `Pipe` or `FileHandle` which may be set. It must be called before the process is started.
     @available(macOS 15.0, iOS 18.0, tvOS 18.0, watchOS 11.0, visionOS 2.0, *)
-    public func makeStream(for keyPath: ReferenceWritableKeyPath<Process, Pipe?>, using pipe: Pipe) -> AsyncThrowingStream<SWBDispatchData, any Error> {        precondition(!isRunning) // the pipe setters will raise `NSInvalidArgumentException` anyways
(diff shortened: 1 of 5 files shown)


rconnell9:
@swift-ci test

Tests/SWBUtilTests/ProcessTests.swift
```diff
@@ -144,6 +144,18 @@ fileprivate struct ProcessTests {
         }
     }
 
+#if os(macOS)
```
jakepetroules: Can this be `#if !os(Windows)`? I think Linux and FreeBSD should both have `dd` by default.

Sources/SWBUtil/Process+Async.swift
```diff
@@ -45,10 +45,21 @@ extension Process {
     ///
     /// - note: This method will mutate the `standardOutput` or `standardError` property of the Process object, replacing any existing `Pipe` or `FileHandle` which may be set. It must be called before the process is started.
     @available(macOS 15.0, iOS 18.0, tvOS 18.0, watchOS 11.0, visionOS 2.0, *)
-    public func makeStream(for keyPath: ReferenceWritableKeyPath<Process, Pipe?>, using pipe: Pipe) -> some AsyncSequence<SWBDispatchData, any Error> {
```
jakepetroules, line 48: I don't think you need the changes to this function. I understand why it deadlocks: the AsyncThrowingStream returned by bytes() (which calls into dataStream()) enumerates the stream in the context of the caller, not in the background, so the collect() calls are not concurrent in any way.

However, adding `async let` to the collect calls should be sufficient to ensure they run concurrently, we shouldn't need to double-background it with the extra unstructured Task here.

rconnell9: Makes sense, updated to restore the body and made the minor adjustments to supporting functions.

jakepetroules (approved):

rconnell9:
@swift-ci test

PR #1316, opened 2026-04-16
Infer -library-level ipi for SKIP_INSTALL=YES targets

sepy97:
Add SWIFT_ENABLE_IPI_LIBRARY_LEVEL build setting (default NO). When enabled, targets with SKIP_INSTALL=YES automatically get -library-level ipi passed to the compiler, enabling IPI import diagnostics. An explicit SWIFT_LIBRARY_LEVEL still takes precedence.

rdar://174947092

### DIFF
--- a/Sources/SWBCore/Settings/BuiltinMacros.swift
+++ b/Sources/SWBCore/Settings/BuiltinMacros.swift
@@ -1053,6 +1053,7 @@ public final class BuiltinMacros {
     public static let SWIFT_ENABLE_BATCH_MODE = BuiltinMacros.declareBooleanMacro("SWIFT_ENABLE_BATCH_MODE")
     public static let SWIFT_ENABLE_INCREMENTAL_COMPILATION = BuiltinMacros.declareBooleanMacro("SWIFT_ENABLE_INCREMENTAL_COMPILATION")
     public static let SWIFT_ENABLE_INCREMENTAL_SCAN = BuiltinMacros.declareBooleanMacro("SWIFT_ENABLE_INCREMENTAL_SCAN")
+    public static let SWIFT_ENABLE_IPI_LIBRARY_LEVEL = BuiltinMacros.declareBooleanMacro("SWIFT_ENABLE_IPI_LIBRARY_LEVEL")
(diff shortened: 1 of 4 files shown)


sepy97:
@swift-ci please test

Sources/SWBCore/Settings/BuiltinMacros.swift
```diff
@@ -2295,6 +2296,7 @@ public final class BuiltinMacros {
         SWIFT_ENABLE_BATCH_MODE,
         SWIFT_ENABLE_INCREMENTAL_COMPILATION,
         SWIFT_ENABLE_INCREMENTAL_SCAN,
+        SWIFT_ENABLE_IPI_LIBRARY_LEVEL,
```
owenv, line 2299: Instead of adding a new setting for this, could we add a new entry to the compiler's features.json file and check for that?

sepy97: The idea is to stage this feature and run side builds since it could be a code breaking change

jakepetroules: I know with certainty this will break a few things, since some projects set SKIP_INSTALL=YES on targets that are embedded in other targets with SKIP_INSTALL=NO via copy files build phases, and are still intended to be part of the API surface.

sepy97: With this setting set to NO by default no breaking for either SKIP_INSTALL setting are expected. I see that in general we might want to shift from using SKIP_INSTALL as the source of truth for IPI library level, does that make any sense?

nkcsgexi: hey @owenv , I think configuring library level as IPI and detecting leakage from the compiler is a nice to have feature that users might want a knob to disable. Having this behavior hard-coded in the compiler side in the `features.json` file seems to me to be overly course grained. WDYT?

owenv: ok, that sounds fine

owenv (approved):

PR #1317, opened 2026-04-16
PlatformFilter support for SwiftPM prebuilts.

dschaefer2:
SwiftPM Prebuilts are only supported currently for the host. For other platforms we need to be able to build from the source of the prebuilts. To model this in the PIF, we need to be able to add platform filters on dependencies to select the prebuilt targets for host and the source targets for not host.

This adds an exclude flag to the PlatformFilter to signal to filter out dependencies for the given platform, as opposed to filtering in for the given platform which is the current semantics.

We'll reach a point with SwiftPM that we'll need a more advanced expression model for platform filtering, but this should enable us for now with the prebuilt requirements and add support for host tools along with our current support for macros.

The matching SwiftPM PR is here. https://github.com/swiftlang/swift-package-manager/pull/9905


Sources/SWBCore/ProjectModel/PlatformFilter.swift
```diff
@@ -52,7 +57,11 @@ public final class PlatformFilter: ProjectModelItem, Hashable, Codable {
     }
 
     public static func ==(lhs: PlatformFilter, rhs: PlatformFilter) -> Bool {
```
jakepetroules, line 60: This should just be `lhs.platform == rhs.platform && lhs.environment == rhs.environment && lhs.exclude == rhs.exclude`.

Comparison of the exclude flag should be in external business logic instead of in the == operator

jakepetroules:
Can you add some tests for this?

dschaefer2:
Definitely. That's the [WIP] part in the title :). I need to do some more manual testing as well.

Sources/SWBCore/ProjectModel/PlatformFilter.swift
dschaefer2: Yeah, I was surprised that the == operator seemed to do all the heavy lifting in the filtering. But yes, I'm not very comfortable with this either.

Sources/SwiftBuild/ProjectModel/PlatformFilter.swift
```diff
@@ -33,19 +35,24 @@ extension ProjectModel.PlatformFilter: Codable {
...
         let container = try decoder.container(keyedBy: CodingKeys.self)
         self.platform = try container.decode(String.self, forKey: .platform)
+        self.exclude = try container.decodeIfPresent(Bool.self, forKey: .exclude) ?? false
         self.environment = try container.decodeIfPresent(String.self, forKey: .environment) ?? ""
     }
 
     public func encode(to encoder: any Encoder) throws {
         var container = encoder.container(keyedBy: CodingKeys.self)
         try container.encode(self.platform, forKey: .platform)
+        if let exclude {
```
owenv: Can this directly encode a Bool without the string conversion?

dschaefer2: I was hoping. But I didn't see anything on the decode side for Bool.

jakepetroules: IIRC it's a long standing thing that bools are strings over the wire in PIF JSON, but this actually _is_ decoding as a bool on the other side so I think this code path can be all bool-based

Sources/SWBCore/ProjectModel/PlatformFilter.swift
dschaefer2: Moved this up to the match function. Will look for other cases that I need to take into account.

I think at the end of the day the PlatformFilters needs a bit of a redo as I'm really stretching it's capabilities here.

Sources/SwiftBuild/ProjectModel/PlatformFilter.swift
dschaefer2: Added a decoder for plBool that still defaults back to the StringEnum if it doesn't match.

dschaefer2:
@swift-ci please test

Sources/SWBCore/ProjectModel/PlatformFilter.swift
```diff
@@ -62,19 +68,21 @@ extension PlatformFilter: Comparable {
...
-        return platform + (!environment.isEmpty ? "-\(environment)" : "")
+        return (exclude ? "!" : "") + platform + (!environment.isEmpty ? "-\(environment)" : "")
     }
 }
 
 extension PlatformFilter {
     static func fromBuildConditionParameterString(_ string: String) -> Set<PlatformFilter> {
         return Set(string.components(separatedBy: ";").compactMap {
             let parameters = $0.components(separatedBy: "-")
+            let exclude = parameters[0].first == "!"
```
jakepetroules, line 79: This will trap if `parameters` is empty.

Sources/SWBCore/ProjectModel/ProjectModelItem.swift
```diff
@@ -110,6 +110,19 @@ extension ProjectModelItem
         return try require(key) { try parseOptionalValueForKeyAsStringEnum(key, pifDict: pifDict) }
     }
 
+    public static func parseValueForKeyAsBoolValue(_ key: String, pifDict: ProjectModelItemPIF, defaultValue: Bool = false) throws -> Bool {
```
jakepetroules: Please don't add this, there is already a standard mechanism for parsing PIF bools directly below this, I don't want multiple mechanisms.

Note that this is completed unrelated to the Codable encoding/decoding, which does use real bools and is used for an internal protocol between the tests and the internal project model, unrelated to PIF.

Yes, there are essentially two encoding mechanisms for "PIF". No, this is not great. :)

dschaefer2: @owenv asked this to be encoded as a Bool. The existing mechanism assumes the field is encoded as a String. Which one do I follow?

jakepetroules: We should use string for _this_ decode path in order to be consistent with all of the other bools - there are multiple decode paths which use different bool formattings.

owenv: Jake has a better idea of how this is supposed to work than I do

Sources/SWBCore/ProjectModel/PlatformFilter.swift
dschaefer2: True.

Sources/SWBCore/ProjectModel/ProjectModelItem.swift
dschaefer2: OK, I've reverted back to use the string.

Sources/SWBCore/ProjectModel/PlatformFilter.swift
dschaefer2: Actually, not true. String.components always returns at least one item.

dschaefer2:
One more set of tests incoming. I'm working on some tests to make sure the exclusions affect the build graph like I need for prebuilts.

dschaefer2:
@swift-ci please test

dschaefer2:
One more tweak to the test and it'll be ready.

dschaefer2:
This is ready for review.

Tests/SWBTaskConstructionTests/PrebuiltsTaskConstructionTests.swift
```diff
@@ -0,0 +1,277 @@
...
+import struct SWBProtocol.PlatformFilter
+import SWBTestSupport
+@_spi(Testing) import SWBUtil
+
+import SWBTaskConstruction
+import Foundation
+
+/// Task construction tests related to prebuilts from SwiftPM.
+@Suite fileprivate struct PrebuiltsTaskConstructionTests: CoreBasedTests {
+    @Test func prebuiltsAreHostOnly() async throws {
```
owenv: This should specify .requireSDKs(.host)

Tests/SWBTaskConstructionTests/PrebuiltsTaskConstructionTests.swift
```diff
@@ -0,0 +1,277 @@
...
+        let prebuiltsDir = Path.root.join("tmp").join("Test").join("prebuiltsProject").join("build").join("prebuilts")
```
owenv, line 28: We prefer to use withTemporaryDirectory for tests instead of /tmp so it gets cleaned up afterwards

owenv: ah, you're using PseudoFS, this should be alright

owenv (approved):
lgtm

Tests/SWBTaskConstructionTests/PrebuiltsTaskConstructionTests.swift
dschaefer2: I was wondering if I should. I'll do that.

Tests/SWBTaskConstructionTests/PrebuiltsTaskConstructionTests.swift
dschaefer2: Yup. I was actually disappointed I had to use a temporary directory for the Swift SDK manifest. SwiftPM has FileSystem plumbed throughout and am able to use the InMemoryFileSystem for pretty much everything. Apparently Core does not.

dschaefer2:
@swift-ci please test

PR #1320, opened 2026-04-17
Fix single file compiles when source files have same base name.

rjmansfield:
computeOutputPaths used ref.guid as the uniquing suffix for object file disambiguation, but task construction uses a path based hash. The mismatch caused "missing input and no rule to build it" errors.

rdar://144018175


rjmansfield:
@swift-ci please test

owenv (approved):

PR #1321, opened 2026-04-17
Windows: normalize Path at init instead of every property access

roman-bcny:
On Windows, `Path.str` called `FilePath(_str).string` on every access, making dictionary lookups, hashing, and property accesses like `basename` and `fileExtension` each pay for a FilePath round-trip.

Moves the normalization into `init(_ str: String)` so `str` is always pre-normalized, and funnels all other inits through it. Also makes `useLegacyImplementation` static since it's a compile-time constant.

Since `str` is now just a trivial stored property, the private `_str` backing field and computed getter are replaced with a public `let str`.

Noticed while profiling `swift build` on Windows — `evaluateWorkload` was spending ~70% of its time in Dictionary hash lookups and Path string operations due to the repeated normalization. After this change, idle gaps between compilation processes in the build timeline are visibly reduced.

### DIFF
--- a/Sources/SWBUtil/Path.swift
+++ b/Sources/SWBUtil/Path.swift
@@ -52,16 +52,12 @@ public struct Path: Serializable, Sendable {
 
     private init(_ impl: FilePath) {
         // When switching to the new implementation, this store the instance directly
-        _str = impl.string
+        str = impl.string
         precondition(!Self.useLegacyImplementation)
     }
 
-    private let _str: String // normalized at init
-
-    /// The path's file system representation as a string.
-    public var str: String {
-        return _str
-    }
+    /// The path's file system representation as a string, normalized at init.
+    public let str: String
 
     /// The system path separator.
     #if os(Windows)
@@ -129,9 +125,9 @@ public struct Path: Serializable, Sendable {
 
     public init(_ str: String) {
         if Self.useLegacyImplementation {
-            self._str = str
+            self.str = str
         } else {
-            self._str = FilePath(str).string
+            self.str = FilePath(str).string
         }
     }
 
--- a/Sources/SWBUtil/Path.swift
+++ b/Sources/SWBUtil/Path.swift
@@ -58,16 +58,6 @@ public struct Path: Serializable, Sendable {
 
     private let _str: String // normalized at init
 
-    /// Single storage init — normalizes via FilePath on Windows.
-    @inline(__always)
-    private init(storage str: String) {
-        #if os(Windows)
-        self._str = FilePath(str).string
-        #else
-        self._str = str
-        #endif
-    }
-
     /// The path's file system representation as a string.
     public var str: String {
         return _str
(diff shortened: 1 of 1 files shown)


Sources/SWBUtil/Path.swift
```diff
@@ -56,14 +56,20 @@ public struct Path: Serializable, Sendable {
         precondition(!useLegacyImplementation)
     }
 
-    private let _str: String
+    private let _str: String // normalized at init
+
+    @inline(__always)
+    private static func filePathNormalized(_ str: String) -> String {
```
owenv: Can you change this to pass through and check `useLegacyImplementation` instead of #if? It makes it easier to test changing the behavior on other platforms

owenv (commented):
I think this makes sense until we're able to migrate to FilePath-backed paths everywhere and use it as the underlying storage. Just had one comment

owenv:
@swift-ci test

Sources/SWBUtil/Path.swift
roman-bcny: @compnerd pointed out we can just funnel everything into a normalizing init, which does look neater and avoids the naming issue. However I can't use `useLegacyImplementation` there as it's not static. I could make it static if you like but that'll affect a bunch of lines and be more verbose `Self.useLegacyImplementation`. LMK if you still want it.

owenv: I'd prefer to make it static if you don't mind making the change, should be pretty mechanical. There's really no reason for it to currently be an instance property AFAIK

owenv: But yeah, I do like the change to funnel everything through the one `init`

roman-bcny: sure, makes the PR messier but it's a separate commit hopefully that works. Thanks!

Sources/SWBUtil/Path.swift
```diff
@@ -46,24 +46,21 @@ public struct Path: Serializable, Sendable {
...
 
     private init(_ impl: FilePath) {
         // When switching to the new implementation, this store the instance directly
         _str = impl.string
-        precondition(!useLegacyImplementation)
+        precondition(!Self.useLegacyImplementation)
     }
 
-    private let _str: String
+    private let _str: String // normalized at init
```
owenv: nit: I guess there's no reason to have the separate `str` computed property anymore, we could just rename this instead. I can always do that in a follow up

owenv (approved):

owenv:
@swift-ci test

Sources/SWBUtil/Path.swift
roman-bcny: right thanks, done

roman-bcny:
I can't tell what's up with CI, the test passed locally

owenv:
status got cleared, I'll take a look when the tests re-run

@swift-ci test

jakepetroules (approved):
Nice, thanks!

roman-bcny:
Can't tell how to interpret CI failures, can anyone help?

Things like Android:
```
/home/runner/.config/swiftpm/swift-sdks/swift-DEVELOPMENT-SNAPSHOT-2026-04-01-a_android.artifactbundle/swift-android/swift-resources/usr/lib/swift-aarch64/clang/include/module.modulemap:10:8: error: redefinition of module '_Builtin_intrinsics'
```
and Linux:
```
error: Unable to resolve build file: BuildFile<PACKAGE-PRODUCT:targetmismatch_Sample.Sample::BUILDPHASE_0::0> (The workspace has a reference to a missing target with GUID 'PACKAGE-TARGET:Sample')
error: Build failed
```

Are these "expected failures"? Where do I look to understand what's actually failing?

owenv:
@roman-bcny  These are expected in the sense we're not done bringing up these jobs yet, from a CI perspective this looks good to go. Let me know if you need me to merge for you

roman-bcny:
yeah this should be good to go. I don't have a button to merge though. Thanks!

PR #1322, opened 2026-04-17
Mark some Android tests as requiring a host SDK

owenv:
These technically require a host SDK to initialize the build system for the test

rdar://174991207


owenv:
@swift-ci test

mhrawdon (approved):

jakepetroules (approved):

PR #1323, opened 2026-04-19
Don't filter args following -no_exported_symbols for preview dylibs

owenv:
-no_exported_symbols takes no argument, but there was some accidentally copy-pasted code here which assumed it did and attempted to filter it.


owenv:
@swift-ci test

jakepetroules (approved):

PR #1318, opened 2026-04-17
Respect the run destination's choice of specific Windows SDKs

owenv:
Part 1 of addressing https://github.com/swiftlang/swift-package-manager/issues/9939

- Ensure we load the full set of SDKs in the Windows platform
- Disambiguate the clashing canonical names of WindowsExperimental.sdk and Windows.sdk
- Ensure specialization of package targets doesn't replace the Windows SDK in the run destination with the default windows sdk

The specialization changes in particular deserve some scrutiny because they impact all platforms, not just windows.

PR #1326, opened 2026-04-20
Use Swift SDK manifest path as imposed SDKROOT during specialization

### LINKED ISSUE #1325: `unable to find sdk '<platform>'` when building Swift package against a Swift SDK with target-platform specialization
When SwiftPM (`swiftbuild` build system) builds a multi-target Swift package against a Wasm Swift SDK, the build fails during dependency resolution / settings construction with:

```
error: unable to find sdk 'webassembly'
warning: ONLY_ACTIVE_ARCH=YES requested with multiple ARCHS and no active architecture could be computed; building for all applicable architectures
error: Build failed
```

The same root cause should affect any platform that has only a Swift SDK installed (no built-in or platform-extension-registered SDK with the platform's canonical name) — Wasm is the most exposed today because `SWBWebAssemblyPlatform` does not register a fallback SDK the way `SWBAndroidPlatform`'s `AndroidSDKRegistryExtension` does.

This is the follow-up to #1267 / #1266: `addRunDestinationSettingsPlatformSDK` is one of two sites that push the platform-name canonical-SDK string for Swift-SDK-backed builds.

## Reproduction

`https://github.com/swiftwasm/JavaScriptKit`, commit `ff90c3e`:

```bash
TOOLCHAINS=<recent-main-snapshot> \
  swift package --swift-sdk swift-DEVELOPMENT-SNAPSHOT_wasm \
  --traits Tracing --disable-sandbox \
  js test --build-only --prelude ./Tests/prelude.mjs
```

```
Building for debugging...
Computing dependencies
Provisioning 19 / 60
error: unable to find sdk 'webassembly'
warning: ONLY_ACTIVE_ARCH=YES requested with multiple ARCHS and no active architecture could be computed; building for all applicable architectures
error: Build failed
```

(Workaround: `--build-system native` succeeds — the legacy build system is unaffected.)

MaxDesiatov:
Synthesized Swift SDKs (registered via `SDKRegistry.synthesizedSDK`) use the SDK's manifest path as their canonical name and not the platform name. When imposing `SDKROOT` for specialization, falling back to `platform.sdkCanonicalName` e.g. `"webassembly"`) leaves the SDK lookup with no match and the build fails with `error: unable to find sdk '<platform>'`. When the active run destination is a Swift SDK targeting this same platform, we instead use its identifier (`runDestination.sdk` = the manifest path) so the lookup hits the synthesized SDK directly.

Resolves https://github.com/swiftlang/swift-build/issues/1325.

### DIFF
--- a/Sources/SWBCore/DependencyResolution.swift
+++ b/Sources/SWBCore/DependencyResolution.swift
@@ -192,32 +192,6 @@ struct SpecializationParameters: Hashable, CustomStringConvertible {
         }
     }
 
-    /// Resolve the SDKROOT value to impose on dependencies, given the build's parameters.
-    ///
-    /// Synthesized Swift SDKs (registered via `SDKRegistry.synthesizedSDK`) use the
-    /// SDK's manifest path as their canonical name — NOT the platform name. So when
-    /// imposing SDKROOT for specialization, falling back to `platform.sdkCanonicalName`
-    /// (e.g. `"webassembly"`) leaves the SDK lookup with no match and the build fails
-    /// with `error: unable to find sdk '<platform>'`. When the active run destination
-    /// is a Swift SDK targeting this same platform, we instead use its identifier
-    /// (`runDestination.sdk` = the manifest path) so the lookup hits the synthesized
-    /// SDK directly.
-    private func resolvedSdkRoot(for parameters: BuildParameters) -> String? {
-        guard let platform else { return nil }
-
-        if let runDestination = parameters.activeRunDestination,
-           case .swiftSDK = runDestination.buildTarget,
-           runDestination.platform == platform.name {
-            let identifier = runDestination.sdk
-            if let canonicalNameSuffix, !canonicalNameSuffix.isEmpty {
-                return "\(identifier).\(canonicalNameSuffix)"
-            }
-            return identifier
-        }
-
-        return sdkRoot
-    }
-
     /// Check if a configured target can be used when this specialization is required.
     func isCompatible(with configuredTarget: ConfiguredTarget, settings: Settings, workspaceContext: WorkspaceContext) -> Bool {
         let toolchain = effectiveToolchainOverride(originalParameters: configuredTarget.parameters, workspaceContext: workspaceContext)
--- a/Sources/SWBCore/DependencyResolution.swift
+++ b/Sources/SWBCore/DependencyResolution.swift
@@ -192,6 +192,32 @@ struct SpecializationParameters: Hashable, CustomStringConvertible {
         }
     }
 
+    /// Resolve the SDKROOT value to impose on dependencies, given the build's parameters.
+    ///
+    /// Synthesized Swift SDKs (registered via `SDKRegistry.synthesizedSDK`) use the
+    /// SDK's manifest path as their canonical name — NOT the platform name. So when
+    /// imposing SDKROOT for specialization, falling back to `platform.sdkCanonicalName`
+    /// (e.g. `"webassembly"`) leaves the SDK lookup with no match and the build fails
+    /// with `error: unable to find sdk '<platform>'`. When the active run destination
+    /// is a Swift SDK targeting this same platform, we instead use its identifier
+    /// (`runDestination.sdk` = the manifest path) so the lookup hits the synthesized
+    /// SDK directly.
+    private func resolvedSdkRoot(for parameters: BuildParameters) -> String? {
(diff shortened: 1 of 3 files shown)


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

owenv:
This might be addressedI think this is likely addressed by https://github.com/swiftlang/swift-build/pull/1318, which is not really windows-specific, do you want to try rebasing the test on that?

Tests/SWBWebAssemblyPlatformTests/SWBWebAssemblyPlatformTests.swift
```diff
@@ -126,4 +126,178 @@ fileprivate struct SWBWebAssemblyPlatformTests: CoreBasedTests {
...
+    ///
+    /// Two code paths in swift-build push `SDKROOT = platform.sdkCanonicalName` (= `"webassembly"`)
+    /// for Swift-SDK-backed builds, neither of which has a matching SDK in the registry when the
+    /// only wasm sysroot comes from a synthesized Swift SDK (whose canonical name is the manifest
+    /// path, not the platform name):
+    ///
+    /// 1. `SpecializationParameters.imposed(on:workspaceContext:)`
+    ///    in `swift-build/Sources/SWBCore/DependencyResolution.swift` — exercised by `MyLibrary`,
+    ///    which is reachable via specialization from `MyApp`.
+    /// 2. `addRunDestinationSettingsPlatformSDK` in
```
owenv, line 141: What package triggers this case?

owenv: The expectation is package targets should always have PIF with SDKROOT: auto

Sources/SWBCore/DependencyResolution.swift
```diff
@@ -192,6 +192,32 @@ struct SpecializationParameters: Hashable, CustomStringConvertible {
...
+    ///
+    /// Synthesized Swift SDKs (registered via `SDKRegistry.synthesizedSDK`) use the
+    /// SDK's manifest path as their canonical name — NOT the platform name. So when
+    /// imposing SDKROOT for specialization, falling back to `platform.sdkCanonicalName`
+    /// (e.g. `"webassembly"`) leaves the SDK lookup with no match and the build fails
+    /// with `error: unable to find sdk '<platform>'`. When the active run destination
+    /// is a Swift SDK targeting this same platform, we instead use its identifier
+    /// (`runDestination.sdk` = the manifest path) so the lookup hits the synthesized
+    /// SDK directly.
+    private func resolvedSdkRoot(for parameters: BuildParameters) -> String? {
```
owenv: looks like this is unused now

Sources/SWBCore/DependencyResolution.swift
```diff
@@ -192,6 +192,32 @@ struct SpecializationParameters: Hashable, CustomStringConvertible {
...
+            let identifier = runDestination.sdk
+            if let canonicalNameSuffix, !canonicalNameSuffix.isEmpty {
+                return "\(identifier).\(canonicalNameSuffix)"
+            }
+            return identifier
+        }
+
+        return sdkRoot
+    }
+
```
MaxDesiatov:
```suggestion
```

Tests/SWBWebAssemblyPlatformTests/SWBWebAssemblyPlatformTests.swift
MaxDesiatov: The repro steps are given in the linked issue, but essentially all packages targeting Wasm and depending on JavaScriptKit (which is arguably the vast majority of SwiftPM packages for Wasm) are broken and not buildable with SwiftBuild at this point https://github.com/swiftlang/swift-build/issues/1325

MaxDesiatov:
@swift-ci test

Tests/SWBWebAssemblyPlatformTests/SWBWebAssemblyPlatformTests.swift
owenv: Those packages don't produce a non-auto SDKROOT though, so the change to the else branch of addRunDestinationSettingsPlatformSDK should be a no-op

owenv (approved):
After discussing offline this LGTM but the test needs `.requireSDKs(.macOS)` before CI will pass

PR #1328, opened 2026-04-20
Add an option which can be used to allow index preparation to enable optimizations

owenv:
Historically index build disabled Swift optimizations, but SourceKit-LSP expects to receive compiler args with -O when setup to use the release configuration. Add a build setting to control the override and adopt it in SWBBuildServer.


owenv:
@swift-ci test

jakepetroules (approved):

PR #1329, opened 2026-04-20
Skip SwiftPM test not compatible with GitHub Actions env

owenv:


owenv:
@swift-ci test

jakepetroules (approved):

jakepetroules:
I expect you had a reason but ideally that skip would go on the test itself...

PR #1330, opened 2026-04-21
Fix task ordering where stub executor does not depend on debug dylib link task (rdar://171933514)

jonathanpenn:
Long standing race bug where the stub executor link task is not properly ordered after the debug dylib link task. That means we've been getting lucky to have them scheduled in the right order most of the time.

The fix makes the stub executor depend on the lipo'd debug dylib.

Added single arch and multi-arch tests to verify the ordering invariant. Without the fix, the tests fail with this assertion:

    task 'Ld .../AppTarget normal' has no edge forcing it to follow 'Ld .../AppTarget.debug.dylib normal'

We've seen some intermittent code sign issues as well, so added extra assertions to validate that the code signing follows the debug dylib and stub executor link tasks, too.

### DIFF
--- a/Sources/SWBTaskConstruction/TaskProducers/BuildPhaseTaskProducers/SourcesTaskProducer.swift
+++ b/Sources/SWBTaskConstruction/TaskProducers/BuildPhaseTaskProducers/SourcesTaskProducer.swift
@@ -1253,7 +1253,7 @@ package final class SourcesTaskProducer: FilesBasedBuildPhaseTaskProducerBase, F
                                     scope: scope,
                                     inputs: linkerInputs,
                                     output: output,
-                                    commandOrderingInputs: additionalLinkerOrderingInputs,
+                                    commandOrderingInputs: additionalLinkerOrderingInputs + [linkedBinaryPreviewDylibNode].compactMap { $0 },
                                     commandOrderingOutputs: commandOrderingOutputs
                                 ),
                                 delegate,
--- a/Tests/SWBTaskConstructionTests/PreviewsTaskConstructionTests.swift
+++ b/Tests/SWBTaskConstructionTests/PreviewsTaskConstructionTests.swift
@@ -1222,4 +1222,181 @@ fileprivate struct PreviewsTaskConstructionTests: CoreBasedTests {
         }
     }
 
+    /// rdar://171933514 — The stub executor link task must be ordered after the debug.dylib link task.
+    /// Without this, llbuild can schedule them concurrently, causing "no such file or directory" for
+    /// the debug.dylib when the stub executor link starts before the debug.dylib link completes.
+    @Test(.requireSDKs(.iOS))
+    func previewsDylibStubExecutorDependsOnDebugDylib() async throws {
+        let core = try await getCore()
+
+        let testProject = try await TestProject(
+            "ProjectName",
+            groupTree: TestGroup(
+                "Sources", path: "Sources",
+                children: [
+                    TestFile("File.swift"),
+                    TestFile("Info.plist"),
+                ]),
+            targets: [
+                TestStandardTarget(
+                    "AppTarget",
+                    type: .application,
+                    buildConfigurations: [
+                        TestBuildConfiguration("Debug", buildSettings: [
+                            "SDKROOT": "iphoneos",
+                            "PRODUCT_NAME": "$(TARGET_NAME)",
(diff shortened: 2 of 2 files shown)


Sources/SWBTaskConstruction/TaskProducers/BuildPhaseTaskProducers/SourcesTaskProducer.swift
```diff
@@ -1253,7 +1253,7 @@ package final class SourcesTaskProducer: FilesBasedBuildPhaseTaskProducerBase, F
                                     scope: scope,
                                     inputs: linkerInputs,
                                     output: output,
-                                    commandOrderingInputs: additionalLinkerOrderingInputs,
+                                    commandOrderingInputs: additionalLinkerOrderingInputs + commandOrderingOutputsPreviewDylib,
```
owenv: It looks like this will be a no-op in multi-architecture builds

jonathanpenn: Yeah, the empty array for `commandOrderingOutputsPreviewDylib` was pre-existing.

Is it safe to change the the code above this to unconditionally do:

```
commandOrderingOutputsPreviewDylib = [linkedBinaryPreviewDylibNode].compactMap { $0 }
commandOrderingOutputsBlankInjectionDylib = [linkedBinaryPreviewBlankInjectionDylibNode].compactMap { $0 }
```

...regardless of arch count?

jonathanpenn: ah, no it seems like there's more going on here. Let me keep digging.

jonathanpenn: @owenv I amended the commit and added a multi-arch test that reproduces the same way. I resolved it in the multi-arch case by creating virtual nodes for the command ordering outputs. From what I can tell they get bound to the command ordering inputs when constructing the linker task in `appendGeneratedTasks`. This does resolve the test failure. Let me know if this isn't the right approach.

Sources/SWBTaskConstruction/TaskProducers/BuildPhaseTaskProducers/SourcesTaskProducer.swift
```diff
@@ -1070,8 +1070,14 @@ package final class SourcesTaskProducer: FilesBasedBuildPhaseTaskProducerBase, F
 
                         outputPreviewDylib = binaryPreviewDylibOutput.map { scope.evaluate(BuiltinMacros.PER_ARCH_OBJECT_FILE_DIR).join("Binary").join($0.basename) }
                         outputPreviewBlankInjectionDylib = binaryPreviewBlankInjectionDylibOutput.map { scope.evaluate(BuiltinMacros.PER_ARCH_OBJECT_FILE_DIR).join("Binary").join($0.basename) }
-                        commandOrderingOutputsPreviewDylib = []
-                        commandOrderingOutputsBlankInjectionDylib = []
+                        // These need per-arch virtual nodes so the stub executor link
+                        // is ordered after the per-arch preview dylib links
+                        commandOrderingOutputsPreviewDylib = outputPreviewDylib.map {
```
owenv: Why do these need to be virtual nodes, are we not modeling the output files of the preceding task correctly?

jonathanpenn: I don't have a good historical understanding of the use of virtual nodes here, but I'm just following the pattern as I can see from the single-arch case. Virtual nodes were already used here for setting up the edges. The multi-arch case just didn't have those edges.

owenv: I think the executor should have a dependency on the lipo-d debug dylib rather than the per-arch versions

jonathanpenn: Ah, thanks! That's the insight I was missing. Depending on the lipo-d debug dylib makes the tests pass. I updated the commit.

owenv (approved):

jonathanpenn:
@swift-ci test

PR #1331, opened 2026-04-21
Add DSYMUTIL_EMBED_RESOURCES build setting for embedding files into dSYM bundles

rjmansfield:
Wires up dsymutil's new --embed-resource option to a new build setting. Resource source paths are tracked as directory tree inputs so that modifying an input triggers dsymutil to rerun.

rdar://50633614

### DIFF
--- a/Sources/SWBCore/SpecImplementations/Tools/DsymutilTool.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/DsymutilTool.swift
@@ -56,7 +56,12 @@ public final class DsymutilToolSpec : GenericCommandLineToolSpec, SpecIdentifier
         // Create a virtual output node so any strip task can be ordered after this.
         let orderingOutputNode = delegate.createVirtualNode("GenerateDSYMFile \(output.str)")
 
-        let inputs: [any PlannedNode] = cbc.inputs.map({ delegate.createNode($0.absolutePath) }) + cbc.commandOrderingInputs
+        let embedResources = cbc.scope.evaluate(BuiltinMacros.DSYMUTIL_EMBED_RESOURCES)
+        let inputs: [any PlannedNode] = cbc.inputs.map({ delegate.createNode($0.absolutePath) }) + cbc.commandOrderingInputs + embedResources.compactMap { entry in
+            guard let src = entry.split(separator: "=", maxSplits: 1).first else { return nil }
+            let path = Path(String(src))
+            return delegate.createDirectoryTreeNode(path.isAbsolute ? path : cbc.scope.evaluate(BuiltinMacros.PROJECT_DIR).join(path))
+        }
         let outputs: [any PlannedNode] = [delegate.createNode(output), orderingOutputNode] + cbc.commandOrderingOutputs
 
         var builder = PlannedTaskBuilder(type: self, ruleInfo: ruleInfo, commandLine: commandLine.map { .literal(ByteString(encodingAsUTF8: $0)) }, environment: environmentFromSpec(cbc, delegate), enableSandboxing: enableSandboxing)
--- a/Sources/SWBUniversalPlatform/Specs/BuiltInCompilers.xcspec
+++ b/Sources/SWBUniversalPlatform/Specs/BuiltInCompilers.xcspec
@@ -68,6 +68,13 @@
                 DefaultValue = NO;
                 CommandLineFlag = "-q";
             },
+            {   Name = DSYMUTIL_EMBED_RESOURCES;
+                Type = StringList;
+                DefaultValue = "";
+                Category = BuildOptions;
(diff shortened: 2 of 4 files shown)


rjmansfield:
@swift-ci test

Sources/SWBCore/SpecImplementations/Tools/DsymutilTool.swift
```diff
@@ -56,7 +58,12 @@ public final class DsymutilToolSpec : GenericCommandLineToolSpec, SpecIdentifier
         // Create a virtual output node so any strip task can be ordered after this.
         let orderingOutputNode = delegate.createVirtualNode("GenerateDSYMFile \(output.str)")
 
-        let inputs: [any PlannedNode] = cbc.inputs.map({ delegate.createNode($0.absolutePath) }) + cbc.commandOrderingInputs
+        let inputs: [any PlannedNode] = cbc.inputs.map({ delegate.createNode($0.absolutePath) }) + cbc.commandOrderingInputs + embedResources.compactMap { entry in
+            // Each entry is "<src-path>=<dst-path>", track the source as an input for incremental builds.
+            guard let src = entry.split(separator: "=", maxSplits: 1).first else { return nil }
```
owenv, line 61: In practice paths sometimes contain '=' which makes this splitting hard to do totally correctly. In practice, do we think there's a use case. Maybe there should be one setting which just copies all the inputs into /Contents/Resources and another setting for /Contents/Resources/Python? I'm not sure there's a use case for copying something anywhere else.

Sources/SWBCore/SpecImplementations/Tools/DsymutilTool.swift
```diff
@@ -46,6 +46,8 @@ public final class DsymutilToolSpec : GenericCommandLineToolSpec, SpecIdentifier
                 } else {
                     return nil
                 }
+            case BuiltinMacros.DSYMUTIL_EMBED_RESOURCES:
```
owenv: I don't think this should be necessary, because embedResources should already be what DSYMUTIL_EMBED_RESOURCES evaluates to (unless we make the change to drop the '=' from the setting)

Sources/SWBUniversalPlatform/Specs/BuiltInCompilers.xcspec
```diff
@@ -68,6 +68,11 @@
                 DefaultValue = NO;
                 CommandLineFlag = "-q";
             },
+            {   Name = DSYMUTIL_EMBED_RESOURCES;
```
owenv, line 71: We should document this with a `Category` and `Description` field since it's user facing

owenv (commented):
mostly lgtm with some small comments

jakepetroules:
Nice, this resolves a long standing issue.

Sources/SWBCore/SpecImplementations/Tools/DsymutilTool.swift
rjmansfield: Well I added `--embed-resource src=dst` to dsymutil because it followed the convention of other options, like dsymutil's --object-prefix-map, or clang's -fdebug-prefix-map, --object-prefix-map options, and IIUC, they split on the first `=` and all suffer from the same problem of parsing paths with =. That said, we don't have to expose this in the build settings.

TBH, I'm not sure what the right path to take on this build setting and I don't have strong opinions. I think right now the main usage is Python scripts but I don't want to paint us into a corner.  

There's a couple options 

1. Use the base name directory as the dest directory. e.g

`DSYMUTIL_EMBED_RESOURCES = $(SRCROOT)/Python $(SRCROOT)/LLDB` which becomes 
`-embed-resource .../Python=Python --embed-resource .../LLDB=LLDB`

2. Introduce yet another setting to specify where the files should be copied to inside of Resources/

```
DSYMUTIL_EMBED_RESOURCES = $(SRCROOT)/scripts/lldb_helpers.py
DSYMUTIL_EMBED_RESOURCES_PATH = Python
```

but then the path setting limits it to one dest dir in the bundle.

3. Live with the `=` limitation.

owenv: I don't love any of these. How about we live with the = for now

rjmansfield:
@swift-ci test

Tests/SWBTaskConstructionTests/DebugInformationTests.swift
```diff
@@ -418,4 +418,44 @@ fileprivate struct DebugInformationTests: CoreBasedTests {
             results.checkNoDiagnostics()
         }
     }
+
+    /// Test that DSYMUTIL_EMBED_RESOURCES generates --embed-resource flags and tracks sources as inputs.
+    @Test(.requireSDKs(.host), .skipHostOS(.windows))
```
owenv: This should probably require a macOS host since we won't generate dSYMs elsewhere. Should resolve the remaining CI failures

owenv (approved):

rjmansfield:
@swift-ci test

PR #1334, opened 2026-04-21
Add internal macro conditions for host and destination platforms

owenv:
These will be used by SwiftPM internally to correctly modify extra flags forwarded via -Xswiftc, command plugins, etc. These are not meant to be user facing, and only work correctly if a run destination is used (which is always the case when driving the build system via SwiftPM / when using packages.


owenv:
@swift-ci test

jakepetroules (approved):

PR #1336, opened 2026-04-22
Introduce `SWIFT_DISABLE_HEADERMAPS` build setting

akyrtzi:
This allows opting-in for disabling headermaps for Swift targets only, as an easier transition path compared to disabling headermaps for the whole project.


akyrtzi:
@swift-ci Please test

owenv (approved):

PR #1337, opened 2026-04-22
Drop version from environment when constructing platform filters for Android triples

owenv:
Android triples may contain version numbers in the environment field, which must be dropped for platform filter matching to work

### DIFF
--- a/Sources/SWBCore/PlatformFiltering.swift
+++ b/Sources/SWBCore/PlatformFiltering.swift
@@ -18,11 +18,22 @@ extension PlatformFilter {
         let platformName = scope.evaluate(BuiltinMacros.PLATFORM_NAME)
         let os = (!scope.evaluate(BuiltinMacros.__USE_PLATFORM_NAME_FOR_FILTERS) ? scope.evaluate(BuiltinMacros.SWIFT_PLATFORM_TARGET_PREFIX).nilIfEmpty : nil) ?? platformName
 
-        // We always want developers to set platform filters for a device platform *and* its simulator counterpart as a *single* inseparable unit.
-        // To implicitly enforce this behavior (and avoid the need for Swift Build clients to replicate it individually) for any target platform
-        // whose *environment* is "simulator" we simply omit it, effectively treating the target the same as the corresponding device platform.
+        // FIXME: We should consider moving to directly using the triple for the whole computation here once we have more tests in place.
+        let triple = try? LLVMTriple(scope.evaluate(BuiltinMacros.SWIFT_TARGET_TRIPLE))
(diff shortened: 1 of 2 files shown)


owenv:
@swift-ci test

Sources/SWBCore/PlatformFiltering.swift
```diff
@@ -18,11 +18,22 @@ extension PlatformFilter {
         let platformName = scope.evaluate(BuiltinMacros.PLATFORM_NAME)
         let os = (!scope.evaluate(BuiltinMacros.__USE_PLATFORM_NAME_FOR_FILTERS) ? scope.evaluate(BuiltinMacros.SWIFT_PLATFORM_TARGET_PREFIX).nilIfEmpty : nil) ?? platformName
 
-        // We always want developers to set platform filters for a device platform *and* its simulator counterpart as a *single* inseparable unit.
-        // To implicitly enforce this behavior (and avoid the need for Swift Build clients to replicate it individually) for any target platform
-        // whose *environment* is "simulator" we simply omit it, effectively treating the target the same as the corresponding device platform.
+        // We should consider moving to directly using the triple for the whole computation here once we have more tests in place.
```
jakepetroules: This seems more of a FIXME level comment

Sources/SWBCore/PlatformFiltering.swift
```diff
@@ -18,11 +18,22 @@ extension PlatformFilter {
...
         let targetTripleSuffix = scope.evaluate(BuiltinMacros.LLVM_TARGET_TRIPLE_SUFFIX)
-        let env = targetTripleSuffix == "-simulator" ? "" : targetTripleSuffix
+        let env: String
+
+        if targetTripleSuffix == "-simulator" {
+            // We always want developers to set platform filters for a device platform *and* its simulator counterpart as a *single* inseparable unit.
+            // To implicitly enforce this behavior (and avoid the need for Swift Build clients to replicate it individually) for any target platform
+            // whose *environment* is "simulator" we simply omit it, effectively treating the target the same as the corresponding device platform.
+            env = ""
+        } else if let androidEnv = triple?.unversioned.environment, ["android", "androideabi"].contains(androidEnv) {
```
jakepetroules: also `qnx`

maybe you could do `else if let triple, triple.unversioned.environment != triple.environment` to make it more generic?

owenv: environment is already normalizing away the version, so I changed this to compare .environment to .environmentComponent

owenv:
@swift-ci test

mhrawdon (approved):

Sources/SWBCore/PlatformFiltering.swift
jakepetroules: Cool. Then that'll work for QNX too.

PR #1338, opened 2026-04-22
When imposing an sdk during specialization, ensure we don't apply a s…

owenv:
…pecialized canonical name suffix twice

Small fix on top of https://github.com/swiftlang/swift-build/pull/1318/changes#diff-01206525e7218e8476d47b5585093d407ef6418e1e6aac147d0c44eb801f10a4. Ensure that if the run destination SDK matches the imposed platform, we are imposing an SDK suffix, and the run destination SDK already has that suffix, we don't append it a second time

### DIFF
--- a/Sources/SWBCore/DependencyResolution.swift
+++ b/Sources/SWBCore/DependencyResolution.swift
@@ -170,7 +170,7 @@ struct SpecializationParameters: Hashable, CustomStringConvertible {
             } else {
                 sdkNameBase = platform.sdkCanonicalName
             }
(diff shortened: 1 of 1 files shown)


owenv:
@swift-ci test

jakepetroules (approved):

jakepetroules:
This is subtle and might warrant a test...

owenv:
I haven't figured out a good way to mock an SDK to create one yet

owenv:
@swift-ci test windows

PR #1339, opened 2026-04-22
Specify -language-mode when building test libraries

owenv:
We want to enforce that swiftinterfaces always specify a language mode. Update some test code which wasn't doing this


owenv:
@swift-ci test

jakepetroules (approved):

owenv:
@swift-ci test

PR #1340, opened 2026-04-22
Fix the regular expressions used for locating module errors.

cwakamo:
Some of these were inconsistently usign curly quotes, but the compiler only emits straight quotes. This fixes the regular expressions so they match what the compiler produces.

This addresses <rdar://problem/175287268>.


cwakamo:
@swift-ci please test

jakepetroules (approved):

PR #1341, opened 2026-04-23
Do not attempt to consume additional linker args when ebm is disabled due to Swift 4 usage

owenv:
The code producing and consuming -linker-args.resp files used inconsistent criteria. When using a compiler which supported explicit module based debug info, but where explicit modules were disabled due to use of Swift 4 mode or a blocklist entry, we could end up attempting to consume a file which would not be produced. Centralize the logic for this in the Swift spec to ensure it's consistent, and add a basic regression test for compiles using g  the 4.2 language mode


owenv:
@swift-ci test

jakepetroules (approved):

owenv:
@swift-ci test

PR #1343, opened 2026-04-23
Stabilize custom task hash suffix calculation

cmcgee1024:
The suffix of the custom task hash is based on the current environment variables and their values. This is done to force re-running the custom task whenever there is a change in the environment in case the change might cause different results in the custom task.

The md5 hash suffix for the custom task is calculated by iterating over the environment bindings dictionary. However, Swift dictionary iteration is intentionally randomized, and ordering affects the hash calculation. The same environment might produce different results.

Change the hashing of the environment to use a repeatable ordering to maintain hash stability.


cmcgee1024:
@swift-ci please test

PR #1345, opened 2026-04-24
Add EmitFrontendCommandLines user default

benlangmuir:
Introduce a new user default that sets EMIT_FRONTEND_COMMAND_LINES persistently and works better with swift package builds outside Xcode. This is a lighter-weight alternative to EnableDebugActivityLogs for users who only want frontend command lines logged.

Why not just use EnableDebugActivityLogs? It's too verbose for my use case where I want to leave this enabled permanently. Debug logs are ~4x larger in storage size than simply emitting frontend lines, and more importantly they are ~15x more lines, which I find makes them very challenging to read as a human (the discrepancy between 4x and 15x is because many of the extra lines have few characters on them while the frontend commands themselves are very long lines). That was measured building swift-build itself, but it matches my general impression from other projects.

### DIFF
--- a/Sources/SWBBuildService/Messages.swift
+++ b/Sources/SWBBuildService/Messages.swift
@@ -312,7 +312,8 @@ private struct SetSessionUserPreferencesMsg: MessageHandler {
             enableBuildSystemCaching: message.enableBuildSystemCaching,
             activityTextShorteningLevel: message.activityTextShorteningLevel,
             usePerConfigurationBuildLocations: message.usePerConfigurationBuildLocations,
-            allowsExternalToolExecution: message.allowsExternalToolExecution ?? UserPreferences.allowsExternalToolExecutionDefaultValue)
+            allowsExternalToolExecution: message.allowsExternalToolExecution ?? UserPreferences.allowsExternalToolExecutionDefaultValue,
+            emitFrontendCommandLines: message.emitFrontendCommandLines ?? false)
         )
 
         return VoidResponse()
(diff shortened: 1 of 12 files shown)


benlangmuir:
@swift-ci test

Sources/SWBCore/Settings/Settings.swift
```diff
@@ -2115,6 +2115,10 @@ private class SettingsBuilder: ProjectMatchLookup {
             pushTable(.exported) { $0.push(BuiltinMacros.COMPILATION_CACHE_CAS_PATH, BuiltinMacros.namespace.parseString("$(\(compileCacheDirMacro))/CompilationCache.noindex")) }
         }
 
+        if UserDefaults.emitFrontendCommandLines {
```
jakepetroules: We try to avoid reading UserDefaults from static globals like this for new settings. Take a look at UserPreferences instead

jakepetroules (changes requested):

Sources/SWBCore/Settings/Settings.swift
benlangmuir: Do `setUserPreferences` or `SetSessionUserPreferencesRequest` have API stability or other compatibility requirements? I see we already have one deprecated setUserPreferences that lets you skip one value, which is what made me wonder about this.

jakepetroules: IPC requests like SetSessionUserPreferencesRequest or public APIs in the SwiftBuild module should try to maintain temporary compatibility on the decode side for approximately one release cycle. Other than that, no.

benlangmuir:
@swift-ci test

PR #1346, opened 2026-04-25
Make swiftCompilerTag optional

jakepetroules:
This is technically not guaranteed to be present. In these cases the relevant tests will be cleanly skipped rather than failing.


jakepetroules:
@swift-ci test

3405691582 (approved):

owenv (approved):

PR #1348, opened 2026-04-28
Fix test target compilation for WebAssembly

owenv:
- Build test targets as static archives when targeting WebAssembly
- Ensure test runners use --whole-archive when incorporating static archives so we don't drop swift-testing tests
- Refactor handling of search paths vs direct inputs when constructing archiver arguments
- Fix an issue where the generic unix test product type wasn't setting ENABLE_TESTING_SEARCH_PATHS correctly


owenv:
@swift-ci test

owenv:
@swift-ci test

jakepetroules (approved):

PR #1349, opened 2026-04-28
Fix a trivial unused variable warning

owenv:
The last use of this was dropped by a recent bugfix

### DIFF
--- a/Sources/SWBCore/BuildRequestContext.swift
+++ b/Sources/SWBCore/BuildRequestContext.swift
@@ -193,7 +193,7 @@ extension BuildRequestContext {
                 }
                 let (outputDir, outputSuffix) = computeOutputParameters(for: input, command: command, settings: settings, lookup: lookup)
                 let uniquingSuffix: String
-                if let ref, sourceCodeBasenames.filter({ $0 == file.basenameWithoutSuffix.lowercased() }).count > 1 && Self.fileTypesWhichUseUniquing.contains(input.fileType.identifier) {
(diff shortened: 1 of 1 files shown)


owenv:
@swift-ci test

owenv:
@swift-ci test macOS

Sources/SWBCore/BuildRequestContext.swift
```diff
@@ -193,7 +193,7 @@ extension BuildRequestContext {
                 }
                 let (outputDir, outputSuffix) = computeOutputParameters(for: input, command: command, settings: settings, lookup: lookup)
                 let uniquingSuffix: String
-                if let ref, sourceCodeBasenames.filter({ $0 == file.basenameWithoutSuffix.lowercased() }).count > 1 && Self.fileTypesWhichUseUniquing.contains(input.fileType.identifier) {
+                if sourceCodeBasenames.filter({ $0 == file.basenameWithoutSuffix.lowercased() }).count > 1 && Self.fileTypesWhichUseUniquing.contains(input.fileType.identifier) {
```
jakepetroules, line 240: Isn't this a behavior change, and merely resolving the warning would be changing to `ref != nil`, not removing?

owenv: AFAICT it should be fine to pass this check if ref is nil, so long as there's a base name match

jakepetroules (approved):

owenv:
@swift-ci test

owenv:
@swift-ci test

PR #1350, opened 2026-04-28
Ensure embed-in-code accessor is regenerated when resources change

owenv:
Previously the accessor content was being generated at build planning-time and written via an auxiliary file task action. Instead, add a new spec & task action which generate the accessor at task execution-time so we can properly invalidate it when the resources change.

### DIFF
--- a/Sources/SWBTaskExecution/TaskActions/GenerateEmbedInCodeAccessorTaskAction.swift
+++ b/Sources/SWBTaskExecution/TaskActions/GenerateEmbedInCodeAccessorTaskAction.swift
@@ -1,0 +1,75 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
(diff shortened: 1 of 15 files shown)


owenv:
@swift-ci test

Sources/SWBTaskExecution/TaskActions/GenerateEmbedInCodeAccessorTaskAction.swift
```diff
@@ -0,0 +1,75 @@
...
+import ArgumentParser
+import Foundation
+
+/// Generates the `embedded_resources.swift` accessor for resources marked `embedInCode`.
+public final class GenerateEmbedInCodeAccessorTaskAction: TaskAction {
+    public override class var toolIdentifier: String {
+        return "generate-embed-in-code-accessor"
+    }
+
+    private struct Options: ParsableArguments {
```
jakepetroules, line 25: ❤️

jakepetroules (approved):

owenv:
@swift-ci test

PR #1355, opened 2026-04-29
Fixup libfuzzer linker flags

owenv:
Ensure ENABLE_LIBFUZZER works with clang or swiftc as the linker driver

Fixes: #1231


owenv:
@swift-ci test

jakepetroules (approved):

PR #1356, opened 2026-04-29
Fix free space test skip in PBXCPTests.largerFile

owenv:
This has been failing intermittently for a little while in CI. Move the return out of the withKnownIssue closure to make sure we correctly early-exit the test when skipping it

### DIFF
--- a/Tests/SWBTaskExecutionTests/PBXCpTests.swift
+++ b/Tests/SWBTaskExecutionTests/PBXCpTests.swift
@@ -453,8 +453,8 @@ fileprivate struct PBXCpTests: CoreBasedTests {
             if let freeSpace = try fs.getFreeDiskSpace(tmp), freeSpace < requiredSpace {
                 withKnownIssue {
                     Issue.record("There is not enough free disk space to run this test (required: \(requiredSpace), free: \(freeSpace))")
-                    return
                 }
+                return
             }
 
             // Test copying a large file.


owenv:
@swift-ci test

Tests/SWBTaskExecutionTests/PBXCpTests.swift
```diff
@@ -453,8 +453,8 @@ fileprivate struct PBXCpTests: CoreBasedTests {
             if let freeSpace = try fs.getFreeDiskSpace(tmp), freeSpace < requiredSpace {
                 withKnownIssue {
                     Issue.record("There is not enough free disk space to run this test (required: \(requiredSpace), free: \(freeSpace))")
```
jakepetroules, line 455: We should be able to use 6.3 APIs unconditionally now. How about using `Test.cancel` instead of withKnownIssue + Issue.record? That's actually the semantically correct action here.

owenv: Some dependents are still building with 6.2 AFAIK

jakepetroules: Oh? Which ones?

jakepetroules (approved):

Tests/SWBTaskExecutionTests/PBXCpTests.swift
owenv: SwiftPM, e.g. https://ci.swift.org/job/swift-package-manager-with-xcode-self-hosted-PR-osx-release-toolchain/4660/

jakepetroules: Hah and that job is even labeled 6.1 despite running with 6.2.

PR #1359, opened 2026-04-29
Using .unsafeFlags(["-weak_framework", "X"]) linker setting results in a wrong linker argument: unknown argument: '-weak_framework'

rconnell9:
swiftc does not understand raw ld flags such as -weak_framework and requires them to be wrapped with -Xlinker.

### DIFF
--- a/Sources/SWBCore/SpecImplementations/Tools/LinkerTools.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/LinkerTools.swift
@@ -683,52 +683,57 @@ public final class LdLinkerSpec : GenericLinkerSpec, SpecIdentifierType, @unchec
                 // We need to keep otherwise unused stub executor library symbols present so
                 // PreviewsInjection can call them when doing the XOJIT handshake.
                 return cbc.scope.namespace.parseLiteralString("YES")
-            case BuiltinMacros.OTHER_LDFLAGS where isPreviewDylib:
-                let ldFlagsToEvaluate: [String]
-                if dyldEnvDiagnosticBehavior == .warning {
-                    ldFlagsToEvaluate = filterLinkerFlagsWhenUnderPreviewsDylib(originalLdFlags)
-                }
-                else {
-                    ldFlagsToEvaluate = originalLdFlags
-                }
+            case BuiltinMacros.OTHER_LDFLAGS where isPreviewDylib || resolvedLinkerDriver == .swiftc:
+                var ldFlagsToEvaluate = originalLdFlags
+                if isPreviewDylib {
+                    if dyldEnvDiagnosticBehavior == .warning {
+                        ldFlagsToEvaluate = filterLinkerFlagsWhenUnderPreviewsDylib(originalLdFlags)
+                    }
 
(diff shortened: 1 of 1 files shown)


mhrawdon (approved):

rconnell9:
@swift-ci test

rjmansfield (approved):

Sources/SWBCore/SpecImplementations/Tools/LinkerTools.swift
```diff
@@ -2020,6 +2025,23 @@ fileprivate func enumerateLinkerCommandLine(arguments: [String], handleWl: Bool
...
+    var result: [String] = []
+    var it = flags.makeIterator()
+    while let arg = it.next() {
+        if arg == "-Xlinker" {
+            result.append(arg)
+            if let next = it.next() {
+                result.append(next)
+            }
+        } else {
+            result.append("-Xlinker")
```
owenv, line 2044: Unconditionally adding the -Xlinker prefix will break for flags which are accepted by clang when driving the linker, but rejected by the underlying linker. I think this will need to look for a specific set of known flags

Sources/SWBCore/SpecImplementations/Tools/LinkerTools.swift
```diff
@@ -683,52 +683,57 @@ public final class LdLinkerSpec : GenericLinkerSpec, SpecIdentifierType, @unchec
...
-                // executor can find something, even if the symbol would have been hidden by
-                // `-fvisibility=hidden`.
-                // rdar://122928395 ("Could not find entry point 'main'" in preview dylib)
-                let entryPoint = cbc.scope.evaluate(BuiltinMacros.LD_ENTRY_POINT)
-                let ldFlagsForEntryPointAlias = [
-                    "-Xlinker", "-alias",
-                    "-Xlinker", entryPoint.nilIfEmpty ?? "_main",
-                    "-Xlinker", "___debug_main_executable_dylib_entry_point",
-                ]
+            case BuiltinMacros.OTHER_LDFLAGS where isPreviewDylib || resolvedLinkerDriver == .swiftc:
```
owenv: nit: Maybe we should move the isPreviewDylib || resolvedLinkerDriver == .swiftc condition inside the switch case and break them out into separate branches of an `if` to make this more readable

owenv (changes requested):

Sources/SWBCore/SpecImplementations/Tools/LinkerTools.swift
owenv: Looks like the swiftDriverRawOutputGetsNewlines failure in https://github.com/swiftlang/swift-build/actions/runs/25126428043/job/73640696424?pr=1359 is an example of this happening

rconnell9: Ah, I was curious about that.  I'll update to look for specific flags

rconnell9:
@swift-ci test

owenv (approved):

PR #1360, opened 2026-04-30
Respect INDEX_ENABLE_DATA_STORE=NO for Swift, Clang, and Metal index store options

daveinglis:
The *_INDEX_STORE_ENABLE xcspec settings use INDEX_ENABLE_DATA_STORE as their DefaultValue, but if a PIF explicitly sets them to YES (as SwiftPM's new PIF builder does for package targets), the default is bypassed and INDEX_ENABLE_DATA_STORE=NO on the command line has no effect.

Add INDEX_ENABLE_DATA_STORE as a leading guard on the Condition for each setting, making it authoritative regardless of what the PIF provides.

rdar://174859173

### DIFF
--- a/Sources/SwiftBuild/ProjectModel/BuildSettings.swift
+++ b/Sources/SwiftBuild/ProjectModel/BuildSettings.swift
@@ -42,6 +42,7 @@ extension ProjectModel {
             case GCC_C_LANGUAGE_STANDARD
             case GCC_OPTIMIZATION_LEVEL
             case GENERATE_PRELINK_OBJECT_FILE
+            case INDEX_ENABLE_DATA_STORE
             case INFOPLIST_FILE
             case IPHONEOS_DEPLOYMENT_TARGET
             case KEEP_PRIVATE_EXTERNS


daveinglis:
@swift-ci test

Sources/SWBApplePlatform/Specs/MetalCompiler.xcspec
```diff
@@ -227,7 +227,7 @@
                 Name = "MTL_ENABLE_INDEX_STORE";
                 Type = Boolean;
                 DefaultValue = "$(INDEX_ENABLE_DATA_STORE)";
-                Condition = "$(COMPILER_INDEX_STORE_ENABLE) == YES || ( $(COMPILER_INDEX_STORE_ENABLE) == Default && $(MTL_ENABLE_DEBUG_INFO) != NO )";
```
owenv, line 230: Instead of changing this here, should we consider updating the SwiftPM PIFBuilder to set INDEX_ENABLE_DATA_STORE=YES instead of SWIFT_INDEX_STORE_ENABLE=YES? There are a few rare cases where we rely on the fact that these settings bypass the global toggle. (The linux test failures on this PR are because test targets on non-Apple platforms rely on enabling it in support of XCTest discovery)

daveinglis:
@swift-ci test

owenv (approved):

PR #1361, opened 2026-04-30
[6.4] Allow constructing a run destination from an in-memory representation of a Swift SDK

owenv:
I'd like to migrate SwiftPM over to this API, and only directly read Swift SDK metadata from disk for testing.

1. This allows SwiftPM to overlay options set via `swift sdk configure`
2. This allows SwiftPM to handle converting the various versions of the Swift SDK metadata format into a unified representation for the underlying build system.

### DIFF
--- a/Sources/SwiftBuild/SWBBuildParameters.swift
+++ b/Sources/SwiftBuild/SWBBuildParameters.swift
@@ -46,6 +46,10 @@ public struct SWBRunDestinationInfo: Codable, Sendable {
         public static func swiftSDK(sdkManifestPath: String, triple: String) -> SWBBuildTarget {
             SWBBuildTarget(_internalBuildTarget: .swiftSDK(sdkManifestPath: sdkManifestPath, triple: triple))
         }
+
+        public static func swiftSDK(swiftSDK: SWBSwiftSDK, triple: String) -> SWBBuildTarget {
+            SWBBuildTarget(_internalBuildTarget: .inMemorySwiftSDK(swiftSDK: swiftSDK, triple: triple))
+        }
     }
 
     public var buildTarget: SWBBuildTarget
@@ -94,7 +98,7 @@ public struct SWBRunDestinationInfo: Codable, Sendable {
             try container.encode(platform, forKey: .platform)
             try container.encode(sdk, forKey: .sdk)
             try container.encode(sdkVariant, forKey: .sdkVariant)
-        case .swiftSDK:
+        case .swiftSDK, .inMemorySwiftSDK:
             try container.encode(buildTarget, forKey: .buildTarget)
         }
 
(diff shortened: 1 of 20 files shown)


owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test

Sources/SwiftBuild/SWBBuildParameters.swift
```diff
@@ -46,6 +46,10 @@ public struct SWBRunDestinationInfo: Codable, Sendable {
         public static func swiftSDK(sdkManifestPath: String, triple: String) -> SWBBuildTarget {
             SWBBuildTarget(_internalBuildTarget: .swiftSDK(sdkManifestPath: sdkManifestPath, triple: triple))
         }
+
+        public static func inMemorySwiftSDK(swiftSDK: SWBSwiftSDK, triple: String) -> SWBBuildTarget {
```
jakepetroules: nit: I think this would be better as just `swiftSDK` and let overloading do its job. We could deprecate the other overload later.

jakepetroules (approved):

Sources/SwiftBuild/SWBSwiftSDK.swift
```diff
@@ -0,0 +1,64 @@
...
+public struct SWBSwiftSDK: Codable, Sendable {
+    public struct TripleProperties: Codable, Sendable {
+        public var sdkRootPath: String
+        public var swiftResourcesPath: String?
+        public var swiftStaticResourcesPath: String?
+        public var includeSearchPaths: [String]?
+        public var librarySearchPaths: [String]?
+        public var toolsetPaths: [String]?
+
+        public init(sdkRootPath: String, swiftResourcesPath: String?, swiftStaticResourcesPath: String?, includeSearchPaths: [String]?, librarySearchPaths: [String]?, toolsetPaths: [String]?) {
```
jakepetroules: sdkRootPath needs to be nullable because of Android

jakepetroules (changes requested):

Sources/SwiftBuild/SWBSwiftSDK.swift
finagolfin: Not just Android, any platform SDK that is proprietary and cannot be included in the Swift SDK bundle.

owenv:
@swift-ci test

finagolfin:
I tried this with swiftlang/swift-package-manager#10013 and it works to set most of the paths when compiling, with two exceptions:

1. The `--sdk-root-path` path is ignored for Android, as discussed in #1279
2. The `--toolset-path` flag is not actually used to set the internal toolset flags, as is the case with the native build-system too and that I fixed separately for that llbuild-based build system in swiftlang/swift-package-manager#10016

I will fix 1. separately in a subsequent pull, and we should hammer out how to fix 2. too.

owenv:
@swift-ci test

owenv:
@swift-ci test

jakepetroules (approved):

PR #1362, opened 2026-05-01
Promise: Resume the continuations outside the lock to avoid potential deadlock if invoked in a cancellation handler

jakepetroules:
This was already done for the Failure==Never case, but not for the Failure==any Error case


jakepetroules:
@swift-ci test

owenv (approved):

PR #1368, opened 2026-05-04
Remove FIXMEs for rdar://24564960

rjmansfield:
The module map file contents are no longer ingested at task construction time so the workarounds are no longer needed.


rjmansfield:
@swift-ci please test

owenv (approved):

PR #1373, opened 2026-05-06
Ensure toolset linker flag escaping matches SwiftPM behavior

owenv:
Align the escaping behavior for linker flags in toolsets with the SwiftPM native build system. This fixes the WebAssembly build of SwiftSyntax's tests when used with a toolset which adjusts the default stack size


owenv:
@swift-ci test

owenv:
@swift-ci test

jakepetroules (approved):

PR #1377, opened 2026-05-06
Add INVOKE_SSAF and EXTRACT_SUMMARIES build settings for Clang

t-rasmud:
Introduces two new build settings:
  - INVOKE_SSAF (Boolean): when enabled alongside clang's `invoke-ssaf` feature flag, passes `--ssaf-extract-summaries` and `--ssaf-tu-summary-file` to the compiler, writing a per-TU JSON summary co-located with the object file.
  - EXTRACT_SUMMARIES (String): controls which summaries to extract; its value is forwarded verbatim to `--ssaf-extract-summaries`.

SWBTaskConstructionTests cover the YES/NO flag combinations and that the EXTRACT_SUMMARIES value is passed through. SwiftBuildOperationTests verify that the summary json file generated for CallGraph analysis on a test file has the correct summary.

### DIFF
--- a/Sources/SWBCore/Settings/BuiltinMacros.swift
+++ b/Sources/SWBCore/Settings/BuiltinMacros.swift
@@ -1016,6 +1016,8 @@ public final class BuiltinMacros {
     public static let REZ_PREFIX_FILE = BuiltinMacros.declarePathMacro("REZ_PREFIX_FILE")
     public static let REZ_SEARCH_PATHS = BuiltinMacros.declarePathListMacro("REZ_SEARCH_PATHS")
     public static let RUN_CLANG_STATIC_ANALYZER = BuiltinMacros.declareBooleanMacro("RUN_CLANG_STATIC_ANALYZER")
+    public static let INVOKE_SSAF = BuiltinMacros.declareBooleanMacro("INVOKE_SSAF")
+    public static let EXTRACT_SUMMARIES = BuiltinMacros.declareStringMacro("EXTRACT_SUMMARIES")
     public static let SWIFT_API_DIGESTER_MODE = BuiltinMacros.declareEnumMacro("SWIFT_API_DIGESTER_MODE") as EnumMacroDeclaration<SwiftAPIDigesterMode>
     public static let RUN_SWIFT_ABI_CHECKER_TOOL = BuiltinMacros.declareBooleanMacro("RUN_SWIFT_ABI_CHECKER_TOOL")
     public static let RUN_SWIFT_ABI_CHECKER_TOOL_DRIVER = BuiltinMacros.declareBooleanMacro("RUN_SWIFT_ABI_CHECKER_TOOL_DRIVER")
@@ -2226,6 +2228,8 @@ public final class BuiltinMacros {
         REZ_PREFIX_FILE,
         REZ_SEARCH_PATHS,
         RUN_CLANG_STATIC_ANALYZER,
+        INVOKE_SSAF,
+        EXTRACT_SUMMARIES,
         RUN_DOCUMENTATION_COMPILER,
         SKIP_BUILDING_DOCUMENTATION,
         RUN_SYMBOL_GRAPH_EXTRACT,
--- a/Sources/SWBCore/SpecImplementations/Tools/CCompiler.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/CCompiler.swift
@@ -1353,6 +1353,11 @@ public class ClangCompilerSpec : CompilerSpec, SpecIdentifierType, GCCCompatible
             commandLine += ["-mllvm", "-cas-friendly-debug-info"]
         }
 
+        if cbc.scope.evaluate(BuiltinMacros.INVOKE_SSAF), clangInfo?.toolFeatures.has(.invokeSsaf) == true {
+            let extractSummariesValue = cbc.scope.evaluate(BuiltinMacros.EXTRACT_SUMMARIES)
+            commandLine += ["--ssaf-extract-summaries=\(extractSummariesValue)", "--ssaf-tu-summary-file=\(outputNode.path.dirname.join(outputNode.path.basenameWithoutSuffix + ".json").str)"]
+        }
+
         // If explicit modules are enabled we need to create the scan task first
         let extraInputs: [any PlannedNode]
         if action != nil {
--- a/Sources/SWBCore/SpecImplementations/Tools/CCompiler.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/CCompiler.swift
@@ -1355,7 +1355,9 @@ public class ClangCompilerSpec : CompilerSpec, SpecIdentifierType, GCCCompatible
 
         if cbc.scope.evaluate(BuiltinMacros.INVOKE_SSAF), clangInfo?.toolFeatures.has(.invokeSsaf) == true {
             let extractSummariesValue = cbc.scope.evaluate(BuiltinMacros.EXTRACT_SUMMARIES)
-            commandLine += ["--ssaf-extract-summaries=\(extractSummariesValue)", "--ssaf-tu-summary-file=\(outputNode.path.dirname.join(outputNode.path.basenameWithoutSuffix + ".json").str)"]
+            let jsonPath = outputNode.path.dirname.join(outputNode.path.basenameWithoutSuffix + ".json")
+            commandLine += ["--ssaf-extract-summaries=\(extractSummariesValue)", "--ssaf-tu-summary-file=\(jsonPath.str)"]
+            extraOutputs.append(delegate.createNode(jsonPath))
         }
 
         // If explicit modules are enabled we need to create the scan task first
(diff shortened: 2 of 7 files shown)


t-rasmud:
@swift-ci test

t-rasmud:
@swift-ci please test macOS Platform

Tests/SWBBuildSystemTests/SwiftBuildOperationTests.swift
```diff
@@ -346,6 +348,178 @@ fileprivate struct SwiftBuildOperationTests: CoreBasedTests {
...
+            try await tester.checkBuild(runDestination: .host) { results in
+                try results.checkTask(.matchRuleType("CompileC")) { task throws in
+                    let objectPath = try #require(task.outputPaths.first { $0.str.hasSuffix(".o") })
+                    let expectedJsonPath = objectPath.dirname.join(objectPath.basenameWithoutSuffix + ".json").str
+                    task.checkCommandLineContains(["--ssaf-extract-summaries=CallGraph"])
+                    task.checkCommandLineContains(["--ssaf-tu-summary-file=\(expectedJsonPath)"])
+
+                    let jsonBytes = try tester.fs.read(Path(expectedJsonPath))
+                    let jsonData = Data(jsonBytes.bytes)
+                    let parsed = try JSONSerialization.jsonObject(with: jsonData) as! [String: Any]
```
jakepetroules: Please do not use JSONSerialization, use PropertyListItem instead in order to keep type safety and eliminate usage of Any

Tests/SWBBuildSystemTests/SwiftBuildOperationTests.swift
```diff
@@ -346,6 +346,183 @@ fileprivate struct SwiftBuildOperationTests: CoreBasedTests {
...
+
+                    // Build id -> USR lookup; entity IDs are assigned non-deterministically
+                    let idTable = try #require(parsed["id_table"]?.arrayValue)
+                    var idToUSR: [Int: String] = [:]
+                    for entry in idTable {
+                        let entryDict = try #require(entry.dictValue)
+                        let id = try #require(entryDict["id"]?.intValue)
+                        let nameInfo = try #require(entryDict["name"]?.dictValue)
+                        idToUSR[id] = try #require(nameInfo["usr"]?.stringValue)
+                    }
```
steakhal: I'm worried a bit about the fragility of this test. If we change the format of the JSON, then things will break here.

I'm not sure if you check the `clang-ssaf-linker` output, but that output is unstable.
Consequently, a similar hard-coded expectations wouldn't work without some normalisation.
This means that we would need to relax the tests anyway. Now, I'm thinking of what could be the least we should test here. The presence of some non-empty but valid JSON maybe?

Remember: the correctness of each component is checked in the relevant component. This means that here we should test the integration, and not the internals of the CallGraph extractor, format, or the linker.

Tests/SWBBuildSystemTests/SwiftBuildOperationTests.swift
```diff
@@ -346,6 +346,183 @@ fileprivate struct SwiftBuildOperationTests: CoreBasedTests {
...
+                    #expect(defLineByUSR["c:@F@foo#"] == 1)
+                    #expect(calleesByUSR["c:@F@foo#"] == [])
+
+                    #expect(prettyNameByUSR["c:@F@bar#"] == "bar()")
+                    #expect(defLineByUSR["c:@F@bar#"] == 3)
+                    #expect(calleesByUSR["c:@F@bar#"] == ["c:@F@foo#"])
+
+                    #expect(prettyNameByUSR["c:@F@baz#"] == "baz()")
+                    #expect(defLineByUSR["c:@F@baz#"] == 7)
+                    #expect(calleesByUSR["c:@F@baz#"] == [])
```
steakhal: IMO all of these are deeply implementation-specific. The USR, the pretty-name, even the fact that we mention the file/line/col for each definition. Ideally, that location metadata would be separate from the CallGraph, because it's essentially a property of an EntityId, thus if any other summary would refer to this EntityId, they would be also interested in its location. So in that sense, it was just a stop-gap solution to have remotely useful CallGraph summary upstream.

Sources/SWBUniversalPlatform/Specs/en.lproj/Apple Clang.strings
```diff
@@ -965,6 +965,9 @@ The restrictions on `offsetof` may be relaxed in a future version of the C++ sta
 "[RUN_CLANG_STATIC_ANALYZER]-name" = "Analyze During 'Build'";
 "[RUN_CLANG_STATIC_ANALYZER]-description" = "Activating this setting will cause Xcode to run the `Clang` static analysis tool on qualifying source files during every build.";
 
+"[INVOKE_SSAF]-name" = "Invoke Static Analyzer Framework";
+"[INVOKE_SSAF]-description" = "Activating this setting will cause the `Clang` static analysis tool to run on qualifying source files, equivalent to enabling `RUN_CLANG_STATIC_ANALYZER`.";
+
```
steakhal, line 973: From this diff, it wasn't clear to me what `static analysis tool` it refers to.
How is this equivalent to `RUN_CLANG_STATIC_ANALYZER`?

Is it in the sense of interposing the build? In other words, it would do its thing part of a regular build?

Sources/SWBCore/Settings/BuiltinMacros.swift
```diff
@@ -2226,6 +2228,8 @@ public final class BuiltinMacros {
         REZ_PREFIX_FILE,
         REZ_SEARCH_PATHS,
         RUN_CLANG_STATIC_ANALYZER,
+        INVOKE_SSAF,
+        EXTRACT_SUMMARIES,
```
steakhal, line 2232: I was expecting the `EXTRACT_SUMMARIES` mentioned in the docs.

Sources/SWBCore/SpecImplementations/Tools/CCompiler.swift
```diff
@@ -1353,6 +1353,11 @@ public class ClangCompilerSpec : CompilerSpec, SpecIdentifierType, GCCCompatible
             commandLine += ["-mllvm", "-cas-friendly-debug-info"]
         }
 
+        if cbc.scope.evaluate(BuiltinMacros.INVOKE_SSAF), clangInfo?.toolFeatures.has(.invokeSsaf) == true {
+            let extractSummariesValue = cbc.scope.evaluate(BuiltinMacros.EXTRACT_SUMMARIES)
+            commandLine += ["--ssaf-extract-summaries=\(extractSummariesValue)", "--ssaf-tu-summary-file=\(outputNode.path.dirname.join(outputNode.path.basenameWithoutSuffix + ".json").str)"]
```
owenv: The JSON file should also be tracked as an output of the task below

Tests/SWBBuildSystemTests/SwiftBuildOperationTests.swift
t-rasmud: We modified this test to only verify that the file isn't empty based on @steakhal 's comment below.

Tests/SWBBuildSystemTests/SwiftBuildOperationTests.swift
t-rasmud: Modified the test to only check for a non-empty JSON file.

Sources/SWBUniversalPlatform/Specs/en.lproj/Apple Clang.strings
t-rasmud: Fixed with correct documentation.

t-rasmud:
Failing test `appHostedTestTargetRebuild` is unrelated to current changes.

t-rasmud:
@jakepetroules @owenv Thanks for your earlier comments. Please take a look at the latest changes and provide your feedback.

Sources/SWBCore/Settings/BuiltinMacros.swift
```diff
@@ -1016,6 +1016,8 @@ public final class BuiltinMacros {
     public static let REZ_PREFIX_FILE = BuiltinMacros.declarePathMacro("REZ_PREFIX_FILE")
     public static let REZ_SEARCH_PATHS = BuiltinMacros.declarePathListMacro("REZ_SEARCH_PATHS")
     public static let RUN_CLANG_STATIC_ANALYZER = BuiltinMacros.declareBooleanMacro("RUN_CLANG_STATIC_ANALYZER")
+    public static let INVOKE_SSAF = BuiltinMacros.declareBooleanMacro("INVOKE_SSAF")
+    public static let EXTRACT_SUMMARIES = BuiltinMacros.declareStringMacro("EXTRACT_SUMMARIES")
```
owenv, line 1020: Should this be a string list that then gets concatenated before it's passed to clang, or is the format of the arg more complicated than that?

owenv (approved):

Sources/SWBCore/Settings/BuiltinMacros.swift
t-rasmud: The value of `EXTRACT_SUMMARIES` is passed through verbatim to `--ssaf-extract-summaries` when `INVOKE_SSAF ` is set to `TRUE`. User can pass a comma separated list of analyses to `EXTRACT_SUMMARIES`, for ex: `CallGraph,UnsafeBufferUsage`.

steakhal (approved):
LGTM

owenv:
@swift-ci test

t-rasmud:
@swift-ci test

t-rasmud:
Thanks for the reviews. Merging this PR. I'm happy to address any post-commit feedback.

PR #1378, opened 2026-05-08
Skip DSTROOT during clean when DEPLOYMENT_LOCATION is off

rjmansfield:
Aligns clean with CreateBuildDirectoryTaskProducer, which already skips DSTROOT when DEPLOYMENT_LOCATION is off.

### DIFF
--- a/Sources/SWBBuildSystem/CleanOperation.swift
+++ b/Sources/SWBBuildSystem/CleanOperation.swift
@@ -392,7 +392,14 @@ package final class CleanOperation: BuildSystemOperation, TargetDependencyResolv
 
 extension WorkspaceContext {
     func buildDirectories(settings: Settings) -> [Path] {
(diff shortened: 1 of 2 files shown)


rjmansfield:
@swift-ci test

rjmansfield:
I should mention when DEPLOYMENT_LOCATION is NO, swift-build does not use DSTROOT for build output, but if the directory happened to exist on disk (maybe pre-existing or a script phase) the clean operation would error. e.g. 

```
error: Could not delete `/tmp/MyProject.dst` because it was not created by the build system                                                                                              
  ▎     note: To mark this directory as deletable by the build system, run `xattr -w com.apple.xcode.CreatedByBuildSystem true /tmp/MyProject.dst` when it is created.
```

rjmansfield:
@swift-ci test

owenv (approved):

Issue #1313, opened 2026-04-16
ObjectLibraryAssembler duplicate detection is case-sensitive, fails on NTFS


roman-bcny:
ddbb57de added duplicate basename detection to ObjectLibraryAssembler, but the detection uses exact (case-sensitive) string comparison on basenames. On NTFS and default macOS APFS (both case-insensitive), files like `Repeat.o` and `repeat.o` pass the duplicate check but collide on disk. The copy fails with `Win32Error(code: 80)` / `NSCocoaErrorDomain Code=516`.

Reproduces with `swift build` (swiftbuild backend) on swift-argument-parser on Windows. The `Examples/repeat/` target produces both `Repeat.o` (from `Repeat.swift`) and `repeat.o` (module emit for module `repeat`). Same issue on `color`, `math`, `default-as-flag` examples.

roman-bcny:
cc @daveinglis

jakepetroules:
Hmm, we'll probably need to add some suffix/disambiguator.

roman-bcny:
The suffix was already added by the change I linked:
https://github.com/swiftlang/swift-build/blob/070cdf8c434385b10ca16e8a6e7b7c9440c4d63f/Sources/SWBTaskExecution/TaskActions/ObjectLibraryAssemblerTaskAction.swift#L55
```
                    // Duplicate detected, add suffix before extension
                    let nameWithoutSuffix = input.withoutSuffix
                    let suffix = input.fileSuffix  // Includes the dot
                    destinationName = "\(Path(nameWithoutSuffix).basename)-\(count)\(suffix)"
```

The files are named {base}-1.o, {base}-2.o etc - when collision is detected. The issue is that collision is _not_ detected because "Repeat" != "repeat" but paths are still colliding on disk.

ps. forgot to cc @compnerd :)

PR #1379, opened 2026-05-09
[6.4] Ensure object library contents don't collide on case-insensitive file systems

owenv:
Resolves an issue building swift-argument-parser on windows, where object libraries build from source files would collide with the module wrap output in a batch mode compile.

Closes https://github.com/swiftlang/swift-build/issues/1313 / https://github.com/apple/swift-argument-parser/issues/896


owenv:
@swift-ci test

jakepetroules (approved):

PR #1380, opened 2026-05-11
Improve makefile dependency parser in verify-swift-dependency-scanner mode

rjmansfield:
The parser used to verify scanner completeness against .d files was only reading the first line of the file. Since swiftc emits multi-line .d files with backslash-newline continuations, the verification was missing dependencies.


rjmansfield:
@swift-ci test

owenv (approved):

PR #1381, opened 2026-05-11
Fixup task construction tests which depend on Swift/Linker version

owenv:
Fixup some task construction tests which depend on the linker in the CI Xcode and/or whether the compiler in use supports the debugInfoExplicitDependency feature


owenv:
@swift-ci test

jakepetroules (approved):

PR #1384, opened 2026-05-11
Copy experimentalWindowsDLL artifact bundle variants to the target build directory

daveinglis:
Artifact bundles with `experimentalWindowsDLL` entries now have their matching variant copied to `$(TARGET_BUILD_DIR)` alongside the built product, using a workspace-level producer that guarantees deterministic, duplicate-free task generation regardless of how many targets transitively reference the same bundle.

Architecture (mirrors XCFrameworkTaskProducer):

- WindowsDLLCopyContext (new) — holds a [destinationPath: CopyRequirement] map on GlobalProductPlan. First registration per destination wins.

- WindowsDLLCopyTaskProducer (new) — workspace-level producer registered alongside XCFrameworkTaskProducer in ProductPlanner. Its prepare() iterates target contexts *sequentially*, triple-matches DLL variants per arch, and registers (source, destination) pairs. Because iteration is sequential the winner is always the same target across incremental builds, eliminating the stale-file-removal warnings that a concurrent first-come-first-served approach would produce. generateTasks() emits one Copy task per registered destination.

Tests:
- artifactBundleWithWindowsDLLs: correct variant copied, non-matching skipped
- artifactBundleWithWindowsDLLsNoMatchWarning: warning + no copy when no variant matches

### DIFF
--- a/Sources/SWBTaskConstruction/CMakeLists.txt
+++ b/Sources/SWBTaskConstruction/CMakeLists.txt
@@ -56,8 +56,10 @@ add_library(SWBTaskConstruction
   TaskProducers/WorkspaceTaskProducers/HeadermapVFSTaskProducer.swift
   TaskProducers/WorkspaceTaskProducers/IndexBuildVFSDirectoryRemapTaskProducer.swift
   TaskProducers/WorkspaceTaskProducers/PCHModuleMapTaskProducer.swift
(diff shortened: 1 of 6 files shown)


daveinglis:
@swift-ci test

daveinglis:
doesn't look like we need this in 6.4

PR #1385, opened 2026-05-11
Unregister app bundles from LaunchServices during clean

rjmansfield:
During builds, record app bundles registered with LaunchServices to a registered-launchservices.txt file. During clean, read the file and call lsregister -u to unregister them.

rdar://142580463

### DIFF
--- a/Sources/SWBBuildSystem/CleanOperation.swift
+++ b/Sources/SWBBuildSystem/CleanOperation.swift
@@ -16,6 +16,7 @@ package import SWBProtocol
 package import SWBUtil
 
 package import class Foundation.FileManager
+package import class Foundation.Process
 package import var Foundation.NSCocoaErrorDomain
 package import var Foundation.NSFileNoSuchFileError
 package import var Foundation.NSLocalizedDescriptionKey
@@ -78,7 +79,7 @@ package final class CleanOperation: BuildSystemOperation, TargetDependencyResolv
     }
 
     package func buildDataDirectory() throws -> Path {
-        return try BuildDescriptionManager.cacheDirectory(buildRequest, buildRequestContext: buildRequestContext, workspaceContext: workspaceContext).join("XCBuildData")
+        return try buildRequestContext.cacheDirectory(for: buildRequest).join("XCBuildData")
     }
 
     package func build() async -> BuildOperationEnded.Status {
@@ -146,6 +147,7 @@ package final class CleanOperation: BuildSystemOperation, TargetDependencyResolv
             }
         }
 
+        await unregisterFromLaunchServices(buildDataDirectory: buildDataDirectory)
         clean(folders: Set(foldersToClean), buildOutputDelegate: buildOutputDelegate)
 
         return delegate.buildComplete(self, status: nil, delegate: buildOutputDelegate, metrics: nil)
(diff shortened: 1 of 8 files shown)


rjmansfield:
@swift-ci test

Sources/SWBBuildSystem/CleanOperation.swift
```diff
@@ -341,6 +344,22 @@ package final class CleanOperation: BuildSystemOperation, TargetDependencyResolv
...
+    private func unregisterFromLaunchServices(buildDataDirectory: Path) {
+        guard workspaceContext.core.hostOperatingSystem == .macOS else { return }
+
+        let recordPath = buildDataDirectory.join("registered-launchservices.txt")
+        guard let content = try? workspaceContext.fs.read(recordPath).asString else { return }
+
+        let paths = content.split(separator: "\n", omittingEmptySubsequences: true).map(String.init)
+        guard !paths.isEmpty else { return }
+
+        _Concurrency.Task.detached {
```
jakepetroules: Let's please stick these in a task group instead of using detached tasks, we really want to maintain structured concurrency, and this as it is is subject to race conditions

jakepetroules (changes requested):

Sources/SWBBuildSystem/BuildOperation.swift
```diff
@@ -304,6 +304,32 @@ package final class BuildOperation: BuildSystemOperation {
         }
     }
 
+    /// Record app bundles registered with LaunchServices so clean can unregister them.
+    private func recordRegisteredLaunchServicesApps() {
```
owenv: Instead of doing this here, could we maybe make it part of the LSRegisterURLTaskAction.performTaskAction implementation instead?

rjmansfield: That's what I had originally tried to do, but AFAICT there's no way for the task actions access the build data directory without protocol changes to DynamicTaskExecutionDelegate, and adding to it didn't look right to me. What would be the right way to do this?

owenv: The path could be added to the command line when creating the task, and then parsed along with the (1) other command line arg inside the task action. Some of the other task actions use ArgumentParser for this, GenerateTestEntryPointTaskAction is a good example

rjmansfield: OK I see how I can pass the directory as an arg, but then the next hiccup is that the task doesn't have access to cacheDirectory either (unlike in the build description). It looks like most cases, it's OBJROOT/XCBuildData but corner cases like when OBJROOT isn't absolute or BUILD_DESCRIPTION_CACHE_DIR

owenv: ah, yeah that might be tricky, I'll have to think about it

rjmansfield: Hmm, what if I move cacheDirectory from BuildDescriptionManager to BuildRequestContext? It already holds workspaceContext and exposes getCachedSettings, and then ProductPostprocessingTaskProducer and CleanOperation could both use it

owenv: I think that may work

owenv (approved):

jakepetroules (approved):

rjmansfield:
@swift-ci test

PR #1386, opened 2026-05-11
BUILD_ONLY_KNOWN_LOCALIZATIONS hits single input assertion when there are multiple Asset Catalogs

matthewseaman:
`cbc.input` has a precondition that there is only one input. That's not guaranteed to be the case with Asset Catalogs.

To fix, stop using `cbc.input` and handle multiple asset catalog paths.


matthewseaman:
@swift-ci test

Tantalum73 (approved):

owenv (approved):

owenv:
This is small and might be worth taking on the 6.4 branch

matthewseaman:
> This is small and might be worth taking on the 6.4 branch

@owenv https://github.com/swiftlang/swift-build/pull/1396

PR #1387, opened 2026-05-11
Update platform versions in Package.swift

jakepetroules:
We need to raise the minimum deployment target so that we can begin using macOS 15 APIs like Atomic and Mutex, and unblock SwiftSubprocess from potentially raising its minimum deployment target to macOS 15 accordingly, as we'll soon be taking it as a dependency.

This follows the sourcekit-lsp update in swiftlang/sourcekit-lsp#2636 and the swift-package-manager update in swiftlang/swift-package-manager#10043

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -75,9 +75,9 @@ let package = Package(
     name: "SwiftBuild",
     defaultLocalization: "en",
     platforms: [
-        .macOS(.v14),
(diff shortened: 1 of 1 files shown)


jakepetroules:
@swift-ci test

owenv (approved):

Issue #1363, opened 2026-05-01
[SwiftBuild] Build output folder no longer includes the target triple


marcprux:
With the "native" (soon to be legacy) build system, we could build for multiple target triples:

```
$ swiftly run swift build --show-bin-path -c debug --swift-sdk x86_64-unknown-linux-android28 +6.3.1
.build/x86_64-unknown-linux-android28/debug

$ swiftly run swift build --swift-sdk aarch64-unknown-linux-android28 +6.3.1
$ swiftly run swift build --swift-sdk armv7-unknown-linux-android28 +6.3.1
$ swiftly run swift build --swift-sdk x86_64-unknown-linux-android28 +6.3.1
```

and then grab the resulting shared objects from `.build/aarch64-unknown-linux-android28/debug/`, `.build/armv7-unknown-linux-android28/debug/`, and `.build/x86_64-unknown-linux-android28/debug/` to bundle into our Android .apk for multi-architecture support.

But now with swift-build:

```
$ swiftly run swift build --build-system native --show-bin-path -c debug --swift-sdk x86_64-unknown-linux-android28 +swift-DEVELOPMENT-SNAPSHOT-2026-04-30-a
.build/x86_64-unknown-linux-android28/debug
warning: '--build-system native' has been deprecated and will be removed in a future release; please report an issue at https://github.com/swiftlang/swift-package-manager/issues if you are unable to adopt the default build system.

$ swiftly run swift build --show-bin-path -c debug --swift-sdk x86_64-unknown-linux-android28 +swift-DEVELOPMENT-SNAPSHOT-2026-04-30-a
.build/out/Products/Debug-android

$ swiftly run swift build --swift-sdk aarch64-unknown-linux-android28 +swift-DEVELOPMENT-SNAPSHOT-2026-04-30-a
$ swiftly run swift build --swift-sdk armv7-unknown-linux-android28 +swift-DEVELOPMENT-SNAPSHOT-2026-04-30-a
$ swiftly run swift build --swift-sdk x86_64-unknown-linux-android28 +swift-DEVELOPMENT-SNAPSHOT-2026-04-30-a
```

everything is dropped into `.build/out/Products/Debug-android/`, without the triple included in the output path, and clobbers the previous triple's build. Is this by design? Is there any possibility we might get the triple included in the output folder?

This might also affect other SDKs that build for multiple triples (e.g., `x86_64-swift-linux-musl` and `aarch64-swift-linux-musl`).

[...]

PR #1388, opened 2026-05-12
[6.4] Add buildTargetInfo API which exposes the effective platform name to clients

owenv:
Expose API on build target info for EFFECTIVE_PLATFORM_NAME / the build products directory suffix. This ensures SwiftPM doesn't need to duplicate the same logic, allowing us to more easily change it to better accommodate platforms which don't support universal binaries (https://github.com/swiftlang/swift-build/issues/1363)


owenv:
@swift-ci test

jakepetroules (approved):

PR #1393, opened 2026-05-13
Fix IPI library-level derivation

sepy97:
In corner cases, when module is IPI (due to SKIP_INSTALL setting), but is still treated as SPI (while IPI library-level is not enabled), swift-build in the current state detects (correctly) the IPI module and just does not emit any flag. Instead, for backward-compatibility, it should skip IPI detection completely and proceed to SPI/API check

rdar://176209013


sepy97:
@swift-ci please test

nkcsgexi (approved):

sepy97:
@swift-ci please test

owenv (approved):

PR #1397, opened 2026-05-13
Update build progress message to include build action count

bripeticca:
Add extra properties to BuildOperationProgressUpdated that track the number of commands to execute + completed, as well as a condensed status message that omits the fraction usually appended to the status property.

### DIFF
--- a/Sources/SWBProtocol/BuildOperationMessages.swift
+++ b/Sources/SWBProtocol/BuildOperationMessages.swift
@@ -809,27 +809,53 @@ public struct BuildOperationProgressUpdated: Message, Equatable {
     /// Whether or not to create a corresponding entry in the build log.
     public let showInLog: Bool
 
-    public init(targetName: String? = nil, statusMessage: String, percentComplete: Double, showInLog: Bool) {
+    /// The number of commands that have been started.
+    public let numCommandsStarted: Int?
+
+    /// The number of build actions to complete.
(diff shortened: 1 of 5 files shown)


Sources/SWBProtocol/BuildOperationMessages.swift
```diff
@@ -809,27 +809,53 @@ public struct BuildOperationProgressUpdated: Message, Equatable {
...
 
-    public init(targetName: String? = nil, statusMessage: String, percentComplete: Double, showInLog: Bool) {
+    /// The number of commands that have been started.
+    public let numCommandsStarted: Int?
+
+    /// The number of build actions to complete.
+    public let numPossibleMaxExecutedCommands: Int?
+
+    /// A condensed status message.
+    public let condensedStatusMessage: String?
```
jakepetroules, line 819: What's the idea behind having `statusMessage` _and_ `condensedStatusMessage`?

bripeticca: The current `statusMessage` contains the `<completed> / <to be executed commands>` fraction, which we don't want on the SwiftPM side. I wanted to make an additive change to this model since there are other possible consumers that could rely on the current derivation of the `statusMessage`, but at the same time we need something different for SwiftPM to construct its own progress logs.

bripeticca:
@swift-ci test

bripeticca:
@swift-ci test

PR #1399, opened 2026-05-13
[6.4] Replace all references to LockedValue with SWBMutex

jakepetroules:
On Darwin this is a no-op since they are typealiased. On other platforms this moves clients to sit on top of the stdlib implementation rather than Swift Build's polyfill. This preps the codebase for removing the polyfill and standardizing on Mutex on all platforms in the main branch targeting the next Swift release.


jakepetroules:
@swift-ci test

owenv (approved):

PR #1402, opened 2026-05-14
[6.4] Update build progress message to include build action count

bripeticca:
Add extra properties to BuildOperationProgressUpdated that track the number of commands to execute + completed, as well as a condensed status message that omits the fraction usually appended to the status property.


bripeticca:
@swift-ci test

owenv (approved):

PR #1404, opened 2026-05-14
Pass IPI Clang module names to the Swift compiler

sepy97:
When a Swift target is compiled, swift-build now computes which Clang modules in its transitive dependency closure should be treated as IPI (project-internal) and passes their names via frontend flags.

The SKIP_INSTALL condition covers target-produced modules including auto-generated modulemaps. The SRCROOT condition covers explicit modulemaps committed to a project's source tree regardless of SKIP_INSTALL.

rdar://177010683

### DIFF
--- a/Sources/SWBTaskConstruction/ProductPlanning/ProductPlan.swift
+++ b/Sources/SWBTaskConstruction/ProductPlanning/ProductPlan.swift
@@ -100,6 +100,11 @@ package final class GlobalProductPlan: GlobalTargetInfoProvider
     /// Maps each target to its artifactbundles (direct and transitive).
     package private(set) var artifactBundlesByTarget: [ConfiguredTarget: [ArtifactBundleInfo]]
 
+    /// For each consumer target, the names of Clang modules in its transitive deps that should
+    /// be treated as IPI — a dep with `SKIP_INSTALL=YES` and a `MODULEMAP_FILE` resolving to a
+    /// path under any `SRCROOT` in the closure.
+    package private(set) var ipiClangModuleNamesByTarget: [ConfiguredTarget: [String]]
+
     /// All targets in the product plan.
     /// - remark: This property is preferred over the `TargetBuildGraph` in the `BuildPlanRequest` as it performs additional computations for Swift packages.
     package private(set) var allTargets: [ConfiguredTarget] = []
@@ -241,6 +246,7 @@ package final class GlobalProductPlan: GlobalTargetInfoProvider
         // Perform post-processing analysis of the build graph
         self.clientsOfBundlesByTarget = Self.computeBundleClients(buildGraph: planRequest.buildGraph, buildRequestContext: planRequest.buildRequestContext)
         self.artifactBundlesByTarget = await Self.computeArtifactBundleInfo(buildGraph: planRequest.buildGraph, provisioningInputs: planRequest.provisioningInputs, buildRequest: planRequest.buildRequest, buildRequestContext: planRequest.buildRequestContext, workspaceContext: planRequest.workspaceContext, getLinkageGraph: getLinkageGraph, metadataCache: self.artifactBundleMetadataCache, delegate: delegate)
+        self.ipiClangModuleNamesByTarget = Self.computeIPIClangInfo(buildGraph: planRequest.buildGraph, buildRequestContext: planRequest.buildRequestContext, workspaceContext: planRequest.workspaceContext)
         let directlyLinkedDependenciesByTarget: [ConfiguredTarget: OrderedSet<LinkedDependency>]
         (self.impartedBuildPropertiesByTarget, directlyLinkedDependenciesByTarget) = await Self.computeImpartedBuildProperties(planRequest: planRequest, getLinkageGraph: getLinkageGraph, delegate: delegate)
         self.mergeableTargetsToMergingTargets = Self.computeMergeableLibraries(buildGraph: planRequest.buildGraph, provisioningInputs: planRequest.provisioningInputs, buildRequestContext: planRequest.buildRequestContext)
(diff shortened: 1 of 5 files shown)


sepy97:
@swift-ci please test

sepy97:
@swift-ci please test Windows

sepy97:
@swift-ci please test

sepy97:
@swift-ci please test macOS

sepy97:
@swift-ci please test Linux

sepy97:
@swift-ci please test

sepy97:
@swift-ci please test

Sources/SWBTaskConstruction/ProductPlanning/ProductPlan.swift
```diff
@@ -100,6 +97,11 @@ package final class GlobalProductPlan: GlobalTargetInfoProvider
     /// Maps each target to its artifactbundles (direct and transitive).
     package private(set) var artifactBundlesByTarget: [ConfiguredTarget: [ArtifactBundleInfo]]
 
+    /// For each consumer target, the names of Clang modules in its transitive deps that should
+    /// be treated as IPI — a dep with `SKIP_INSTALL=YES` and a `MODULEMAP_FILE` resolving to a
+    /// path under any `SRCROOT` in the closure.
+    package private(set) var ipiClangModuleNamesByTarget: [ConfiguredTarget: [String]]
```
owenv, line 106: Instead of exposing this via settings and adding it to all the cache keys, can we access it from the task producer directly via the globalProductPlan?

Sources/SWBTaskConstruction/ProductPlanning/ProductPlan.swift
```diff
@@ -589,6 +592,59 @@ package final class GlobalProductPlan: GlobalTargetInfoProvider
...
+                    // Resolve symlinks via the parent directory — the modulemap file itself may not
+                    // exist yet during planning, but its parent directory should.
+                    let resolvedModulemapDir = (try? localFS.realpath(modulemapPath.dirname)) ?? modulemapPath.dirname
+                    return srcroots.contains(where: {
+                        ((try? localFS.realpath($0)) ?? $0).isAncestorOrEqual(of: resolvedModulemapDir)
+                    })
+                }()
+                guard skipInstall || modulemapUnderSRCROOT else { continue }
+                let name = scope.evaluate(BuiltinMacros.PRODUCT_MODULE_NAME)
+                if !name.isEmpty {
```
owenv: It might be an ok heuristic to use, but if MODULEMAP_FILE or MODULEMAP_FILE_CONTENTS is set the module(s) defined by the modulemap might not actually be named $(PRODUCT_MODULE_NAME)

Tests/SWBTaskConstructionTests/SwiftTaskConstructionTests.swift
```diff
@@ -4081,6 +4008,661 @@ fileprivate struct SwiftTaskConstructionTests: CoreBasedTests {
         }
     }
 
+    // Test -ipi-clang-module emission for a Clang IPI (SKIP_INSTALL=YES + MODULEMAP_FILE under SRCROOT).
+    @Test(.requireSDKs(.macOS))
+    func ipiClangModuleBasic() async throws {
+        if try await !swiftFeatures.has(.ipiClangModule) {
```
owenv: you can use requireSwiftFeatures(.foo) to skip the test entirely instead

Sources/SWBTaskConstruction/ProductPlanning/ProductPlan.swift
sepy97: Great catch! This change should also support manually written modulemaps. I was thinking of using `knownClangModuleNames` via `computeModuleInfo()`, but this case is there under FIXME...

sepy97:
@swift-ci please test

Sources/SWBTaskConstruction/ProductPlanning/ProductPlan.swift
```diff
@@ -589,6 +592,74 @@ package final class GlobalProductPlan: GlobalTargetInfoProvider
...
+    /// `MODULEMAP_FILE` whose resolved path lies under some `SRCROOT` in the consumer's
+    /// transitive dependency closure (including the consumer's own `SRCROOT`).
+    private static func computeIPIClangInfo(buildGraph: TargetBuildGraph, buildRequestContext: BuildRequestContext, workspaceContext: WorkspaceContext) -> [ConfiguredTarget: [String]] {
+        var namesByTarget: [ConfiguredTarget: [String]] = [:]
+        for configuredTarget in buildGraph.allTargets {
+            let deps = transitiveClosure([configuredTarget], successors: buildGraph.dependencies(of:)).0
+            let allTargets = [configuredTarget] + deps
+            // Collect the set of project-internal roots for this consumer.
+            var srcroots: Set<Path> = []
+            for target in allTargets {
```
owenv: This will repeat the module name calculation for a target every time it appears in the transitive closure of another. Can we have a top level loop over all the targets which first computes the module name, then do the transitive closure computation later?

sepy97: Makes sense, thanks for catching it!

sepy97:
@swift-ci please test

Sources/SWBTaskConstruction/ProductPlanning/ProductPlan.swift
```diff
@@ -589,6 +592,100 @@ package final class GlobalProductPlan: GlobalTargetInfoProvider
...
+            // Collect the set of project-internal roots for this consumer
+            var srcroots: Set<Path> = []
+            for target in allTargets {
+                if let root = ipiInfoByTarget[target]?.resolvedSrcroot {
+                    srcroots.insert(root)
+                }
+            }
+            var names: Set<String> = []
+            for target in allTargets {
+                guard let info = ipiInfoByTarget[target], info.producesClangModule else { continue }
```
owenv: This inner loop is still pretty expensive, can we build the sets up incrementally while doing a postorder DFS of the graph?

sepy97:
@swift-ci please test

owenv (approved):

PR #1407, opened 2026-05-18
[6.4] Add resource search paths to CoreCacheKey

owenv:
If 2 cores are initialized with different resource search paths, they should not share a cache key.


owenv:
@swift-ci test

jakepetroules (approved):

PR #1408, opened 2026-05-18
[6.4.x] Fix a missing deserialization case for PathOrderedSet macro declarations

owenv:
This is only used for pre-existing declarations today, but avoid future deserialization issues by handling this case properly.


owenv:
@swift-ci test

mhrawdon (approved):

PR #1410, opened 2026-05-18
[6.4] Fix unbalanced parens in settings contruction macro refs

owenv:
These happened to have the right behavior when parsed, but we really shouldn't rely on the parser to fix them up


owenv:
@swift-ci test

mhrawdon (approved):

PR #1413, opened 2026-05-19
[6.4 🍒] Fix IPI library-level derivation

sepy97:
Explanation: In corner cases, when module is IPI (due to SKIP_INSTALL setting), but is still treated as SPI (while IPI library-level is not enabled), swift-build in the current state detects (correctly) the IPI module and just does not emit any flag. Instead, for backward-compatibility, it should skip IPI detection completely and proceed to SPI/API check
Scope: Fixes typechecking errors when module is IPI due to SKIP_INSTALL setting but is still treated as SPI
Risk: Low
Original PR: https://github.com/swiftlang/swift-build/pull/1393
Testing: Added test to test suite and validated
Reviewers: @owenv


sepy97:
@swift-ci please test

mhrawdon (approved):

PR #1414, opened 2026-05-19
Default PCM pruning to the target's toolchain libclang

rjmansfield:
Previously CLANG_EXPLICIT_MODULES_LIBCLANG_PATH had to be set explicitly for pruning to run. Fall back to the target's toolchain library search paths so pruning works if the toolchain libclang supports it.


rjmansfield:
@swift-ci please test

owenv (approved):

owenv:
We may want to include this on release/6.4.x as well

rjmansfield:
> We may want to include this on release/6.4.x as well

OK, I'll post a 6.4 PR after this merges. I just had to amend to consume some PruneExplicitPrecompiledModules activity, when the toolchain supports the libclang API (not yet in the CI toolchain)

rjmansfield:
@swift-ci please test

PR #1419, opened 2026-05-20
Ensure that the UILaunchScreen plist entry is an empty dictionary

rbenegal:
Previously there was a nested dictionary for `UILaunchScreen` which seems to be incorrect. Ensure we only create a flat, empty dictionary and add a test for this.

After this change, instead of having a nested Info plist like:

```
    UILaunchScreen =     {
        UILaunchScreen =         {
        };
    };
```

It will be a flat list like:

```
UILaunchScreen = {};
```

Related: rdar://173729182


mhrawdon (approved):

rbenegal:
Re-targetting this to the `swiftlang:release/6.4.x` branch.

rbenegal:
@swift-ci test

rconnell9:
@swift-ci test windows

PR #1420, opened 2026-05-21
Support for extracting symbols graphs for clang targets

cmcgee1024:
Add test for TAPI API output override

Follow pretty print setting

Correct build settings enum


cmcgee1024:
@swift-ci please test

owenv (approved):

PR #1427, opened 2026-05-25
Add build configuration filters to build files and target dependencies (SE-0273 follow-up)

dmhts:
This is the swift-build half of finishing the work proposed in [SE-0273][se273], following up on [my forum thread][forum]. SE-0273 originally introduced platform-based conditional dependencies (`.when(platforms:)`). Configuration-based conditionals were part of the discussion at the time but never landed. This PR is the first of two steps to finally ship them. Step two will be the matching SwiftPM adoption.

When I posted on the forums, @owenv pointed out that mirroring the existing platform-filter mechanism would be the most scalable approach, so that's what this does. `BuildConfigurationFilter` is a sibling of `PlatformFilter` end-to-end: the filter type itself, the PIF wire format, the ProjectModel surfaces, every task-producer call site, the diagnostics.

[se273]: https://github.com/swiftlang/swift-evolution/blob/main/proposals/0273-swiftpm-conditional-target-dependencies.md
[forum]: https://forums.swift.org/t/se-0273-configuration-conditionals-for-target-dependencies-time-to-finish-the-job/85669

### What lands here

- New `BuildConfigurationFilter` type in `SWBProtocol`, `SWBCore`, `SwiftBuild.ProjectModel`, and the PIF builder in `SWBProjectModel` (one per layer, matching `PlatformFilter`).
- `buildConfigurationFilters` property on `BuildFile` and `TargetDependency`, with PIF serialization. The new `"buildConfigurationFilters"` key is optional in the loader, so existing PIFs still load fine.
- We check the filter everywhere we already check the platform filter: the filtering logic itself, dependency resolution, linkage, the header / TAPI file list / Swift package auto-embed paths, and the build-dependency-info that goes out to IDE clients.
- A new `.buildConfigurationFilter` exclusion reason with its own "Skipping … because its build configuration filter (…) does not match …" debug note, mirroring the platform-filter equivalent.

### Scope rationale

The end goal of this work is per-configuration target dependencies in SwiftPM (the SE-0273 follow-up). swift-build's existing platform-filter mechanism already covers every build-file kind, not just target dependencies, so this PR mirrors it in full.

### Next step

A follow-up PR in `swiftlang/swift-package-manager` will add `.when(configuration:)` to SwiftPM's condition surface, translate it into the filter set, and route it through the existing `addDependency(..., linkProduct: true)` path. Once both PRs land we'll have parity with `.when(platforms:)` for build configurations.


Sources/SWBCore/ProjectModel/BuildConfigurationFilter.swift
```diff
@@ -0,0 +1,57 @@
...
+public final class BuildConfigurationFilter: ProjectModelItem, Hashable, Codable {
+
+    /// The name of the build configuration
+    public let buildConfiguration: String
+
+    public init(buildConfiguration: String) {
+        self.buildConfiguration = buildConfiguration
+    }
+
+    convenience init(_ model: SWBProtocol.BuildConfigurationFilter, _ pifLoader: PIFLoader) {
```
owenv: The pifLoader arg seems unnecessary

Sources/SWBCore/ProjectModel/BuildConfigurationFilter.swift
```diff
@@ -0,0 +1,57 @@
...
+        self.init(buildConfiguration: model.buildConfiguration)
+    }
+
+    convenience init(fromDictionary pifDict: ProjectModelItemPIF, withPIFLoader pifLoader: PIFLoader) throws {
```
owenv: same here

Sources/SWBCore/ProjectModel/BuildConfigurationFilter.swift
```diff
@@ -0,0 +1,57 @@
...
+        try self.init(
+            buildConfiguration: Self.parseValueForKeyAsString("buildConfiguration", pifDict: pifDict)
+        )
+    }
+
+    public var description: String {
+        return "\(type(of: self)) <buildConfiguration:\(buildConfiguration)>"
+    }
+
+    public func hash(into hasher: inout Hasher) {
```
owenv, line 40: Why not let these be synthesized?

Sources/SWBCore/ProjectModel/BuildConfigurationFilter.swift
```diff
@@ -0,0 +1,57 @@
...
+        return lhs.buildConfiguration == rhs.buildConfiguration
+    }
+}
+
+extension BuildConfigurationFilter: Comparable {
+    public static func <(lhs: BuildConfigurationFilter, rhs: BuildConfigurationFilter) -> Bool {
+        return lhs.buildConfiguration < rhs.buildConfiguration
+    }
+
+    public var comparisonString: String {
```
owenv: buildConfiguration is already public, is this really needed? The name makes the purpose fairly unclear IMO

Sources/SWBCore/BuildConfigurationFiltering.swift
```diff
@@ -0,0 +1,55 @@
...
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+public import SWBMacro
+import SWBUtil
+
+extension BuildConfigurationFilter {
+    public convenience init(_ scope: MacroEvaluationScope) {
+        let buildConfiguration = scope.evaluate(BuiltinMacros.CONFIGURATION)
```
owenv: This will bind to the effective configuration name, which won't necessarily match the configuration of a package if a package config override was applied. I think this potentially needs a bit of design work to determine the right matching strategy

Sources/SWBProtocol/ProjectModel/BuildConfigurationFilter.swift
```diff
@@ -0,0 +1,46 @@
...
+        self.buildConfiguration = buildConfiguration
+    }
+}
+
+extension BuildConfigurationFilter: Comparable {
+    public static func < (lhs: BuildConfigurationFilter, rhs: BuildConfigurationFilter) -> Bool {
+        return lhs.comparisonString < rhs.comparisonString
+    }
+
+    fileprivate var comparisonString: String {
```
owenv: Same here, this comparisonString property appears to be redundant

Sources/SWBProtocol/ProjectModel/BuildConfigurationFilter.swift
```diff
@@ -0,0 +1,46 @@
...
+        return buildConfiguration
+    }
+}
+
+// MARK: SerializableCodable
+
+extension BuildConfigurationFilter: PendingSerializableCodable {
```
owenv: Why is this only PendingSerializableCodable?

Sources/SWBTestSupport/BuildConfigurationFilter.swift
```diff
@@ -0,0 +1,207 @@
...
+//===----------------------------------------------------------------------===//
+
+package import Testing
+
+import SWBUtil
+package import SWBCore
+package import struct SWBProtocol.BuildConfigurationFilter
+
+extension CoreBasedTests {
+    package func testBuildConfigurationFiltering(_ targetType: TestStandardTarget.TargetType = .application, runDestination: RunDestinationInfo, buildConfiguration: String, buildConfigurationFilters: Set<SWBTestSupport.BuildConfigurationFilter> = [], expectFiltered: Bool, sourceLocation: SourceLocation = #_sourceLocation) async throws {
```
owenv: These tests are platform agnostic so all these helpers can live alongside the corresponding tests

Tests/SWBTaskConstructionTests/BuildConfigurationFilteringHeaderTaskConstructionTests.swift
```diff
@@ -0,0 +1,101 @@
...
+
+import Testing
+
+import SWBCore
+import SWBTestSupport
+import SWBUtil
+import struct SWBProtocol.BuildConfigurationFilter
+
+@Suite
+fileprivate struct BuildConfigurationFilteringHeaderTaskConstructionTests: CoreBasedTests {
```
owenv: Most of these test files seem like they could be combined, they're all pretty short

owenv (changes requested):
Thanks,This approach looks fairly reasonable except for one major design decision to be made around package config overrides. note that the unimplemented parts of SE-0273 have expired since it's been >1 year since it was accepted though, so the changes will need to be re-proposed before it officially lands in SwiftPM.

Sources/SWBCore/BuildConfigurationFiltering.swift
dmhts: Thanks for getting to this so quickly! I felt there must've been some catch here. 
From what I see, `Settings.targetConfiguration` already holds exactly the `BuildConfiguration` we'd want, since its resolution already takes the package config override into account. 
So the natural design that comes to mind would be pushing a sibling macro inside `addTargetSettings` carrying `config.name`, and reading that from the filter init instead of `CONFIGURATION`. Something like `_RESOLVED_CONFIGURATION`, or whatever name you think fits better. 
`addIndexBuildOverrides` wouldn't touch the new macro, so filtering during index builds would also be correct.
Does this line up with the matching strategy you had in mind?

Sources/SWBCore/ProjectModel/BuildConfigurationFilter.swift
dmhts: I'd love to, but automatic synthesis of `Hashable` is not supported for class declarations. And the custom `description` mirrors `PlatformFilter`, overriding the less descriptive default in `ProjectModelItem`.

Sources/SWBCore/ProjectModel/BuildConfigurationFilter.swift
dmhts: You're right, not needed. It was a leftover from mirroring `PlatformFilter`, which has the same unused `pifLoader` arg btw. Dropped from both inits [here](https://github.com/swiftlang/swift-build/pull/1427/changes/a690534c0b7126c85aec3704e8c9d48113ebbfca).

Sources/SWBCore/ProjectModel/BuildConfigurationFilter.swift
dmhts: You're right, it's redundant here. `PlatformFilter` has it because it composes multiple fields into a sortable string, but with just `buildConfiguration` there's nothing to compose. Removed [here](https://github.com/swiftlang/swift-build/pull/1427/changes/a690534c0b7126c85aec3704e8c9d48113ebbfca).

Sources/SWBProtocol/ProjectModel/BuildConfigurationFilter.swift
dmhts: Yeah, no migration to worry about since this is new in the PR (unlike `PlatformFilter`). [Switched to `SerializableCodable`](https://github.com/swiftlang/swift-build/pull/1427/changes/f39de32ad8201767d1bf2232b791af7b6fbaef90).

Sources/SWBTestSupport/BuildConfigurationFilter.swift
dmhts: Makes sense. [Done](https://github.com/swiftlang/swift-build/pull/1427/changes/641284ed387edad07b3bd17a5dbffb961a768fae). I kept the convenience constants in SWBTestSupport since they're used across three test targets.

Tests/SWBTaskConstructionTests/BuildConfigurationFilteringHeaderTaskConstructionTests.swift
dmhts: [Done](https://github.com/swiftlang/swift-build/pull/1427/changes/b47d6aa8049d35b15e53c6a4016846ed7db3b37a). Merged the header and package auto-embed tests in. Left BuildFileFilteringInteractionTests separate since it covers combined platform + build-config interaction.

Sources/SWBCore/BuildConfigurationFiltering.swift
owenv: that sounds like a reasonable approach to me

dmhts: [Done](https://github.com/swiftlang/swift-build/pull/1427/commits/471678b8). Small correction to the above: ended up pushing in `addProjectSettings` rather than `addTargetSettings`. The target-level config can fall back to a config the target happens to declare when it doesn't have the requested one, so `effectiveProjectConfig` (which resolves the package override) is the right source.

Sources/SWBCore/ProjectModel/BuildConfigurationFilter.swift
```diff
@@ -0,0 +1,53 @@
+//===----------------------------------------------------------------------===//
```
owenv, line 1: This file will need to be added to CMakeLists.txt

Sources/SWBCore/BuildConfigurationFiltering.swift
```diff
@@ -0,0 +1,55 @@
+//===----------------------------------------------------------------------===//
```
owenv, line 1: Same here, this file will need to be added to CMakeLists.txt

Sources/SWBProtocol/ProjectModel/BuildConfigurationFilter.swift
```diff
@@ -0,0 +1,31 @@
+//===----------------------------------------------------------------------===//
```
owenv, line 1: Same comment about CMakeLists.txt

Sources/SwiftBuild/ProjectModel/BuildConfigurationFilter.swift
```diff
@@ -0,0 +1,44 @@
+//===----------------------------------------------------------------------===//
```
owenv, line 1: Same note about CMakeLists.txt

Tests/SWBTaskConstructionTests/BuildConfigurationFilteringTaskConstructionTests.swift
```diff
@@ -0,0 +1,423 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import Testing
+import SWBTestSupport
+import SWBCore
+import struct SWBProtocol.BuildConfigurationFilter
+import SWBUtil
+
+@Suite
```
owenv, line 19: Most of these tests look like they could just require a host SDK and run on linux/windows as well

owenv:
@swift-ci test

Sources/SWBCore/ProjectModel/BuildConfigurationFilter.swift
dmhts: Thanks, fixed. Added the new sources to the SWBCore, SWBProtocol, and SwiftBuild CMakeLists. The SWBTestSupport one isn't included since that module has no CMake build.

Tests/SWBTaskConstructionTests/BuildConfigurationFilteringTaskConstructionTests.swift
dmhts: Good point. [Adapted them to the host SDK](https://github.com/swiftlang/swift-build/pull/1427/changes/22ee5dee58e7e89b0cc249ae365af5d3f2403240), except the TAPI one. The package auto-embed test is the one I'm less sure about cross-platform, since framework embedding is fairly Apple-specific, so I'll keep an eye on it in CI and gate it back to macOS if it doesn't hold up.

Sources/SwiftBuild/ProjectModel/Targets.swift
```diff
@@ -293,8 +293,8 @@ extension ProjectModel {
 
 
         /// Adds a dependency on another target.  It is the caller's responsibility to avoid creating dependency cycles.  A dependency of one target on another ensures that the other target is built first. If `linkProduct` is true, the receiver will also be configured to link against the product produced by the other target (this presumes that the product type is one that can be linked against).
-        @discardableResult public mutating func addDependency(on targetId: GUID, platformFilters: Set<PlatformFilter>, linkProduct: Bool = false) -> TargetDependency {
-            let dependency = TargetDependency(targetId: targetId, platformFilters: platformFilters)
+        @discardableResult public mutating func addDependency(on targetId: GUID, platformFilters: Set<PlatformFilter>, buildConfigurationFilters: Set<BuildConfigurationFilter> = [], linkProduct: Bool = false) -> TargetDependency {
```
dmhts, line 296: Defaulted this to [] so existing callers like SwiftPM keep compiling. This is the fallback I mentioned in the description, and it should fix the failing [job](https://github.com/swiftlang/swift-build/actions/runs/26976730210/job/79644622863).

dmhts:
@owenv could you please trigger another CI run?

owenv:
@swift-ci test

dmhts:
@owenv I brought the branch up to date with upstream: resolved a conflict in `SettingsTests` and adopted the new `builtProductsDirSuffix(core:)` signature it introduced, which caused the earlier test failures. Could you kick off another CI run when you get a chance?

owenv:
@swift-ci test

Tests/SWBTaskConstructionTests/BuildConfigurationFilteringTaskConstructionTests.swift
dmhts: @owenv CI confirmed the concern about the auto-embed test, so I gated it back to macOS as planned. The source-filtering ones failed too, but for a fixable reason: under the host SDK they used `.application`, whose product type (`com.apple.product-type.application`) doesn't resolve outside Apple platforms. I switched those to `.commandLineTool`, which resolves everywhere, matching what `platformFiltering_Linux` does, and verified them passing in a Linux container, not just the macOS host. The same change should clear the Windows failures too.

So the final split:

- `buildConfigurationFiltering_Debug/_Release`: cross-platform on the host SDK via `.commandLineTool`.
- `filteredPackageFrameworkExcludedFromAutoEmbed`: macOS-only. On Linux the package framework builds to a .so and there's no bundle to embed into.
- `filteredPublicHeaderExcludedFromTAPIFileList`: macOS-only, since TAPI is Apple-only.

Could you kick off another CI run when you get a chance? Hopefully one of the last.

dmhts: Unrelated: while verifying this on Linux I noticed there's no doc on running the tests in a container beyond the .devcontainer config. Happy to add a short "running on Linux" section to the README as a separate PR if that'd be useful.

dmhts:
@owenv Could you please trigger another CI run when you have a minute? Hopefully one of the last.

owenv:
@swift-ci test

dmhts:
@owenv The 3 Windows nightly-main failures were mine and are now fixed. My test asserted hard-coded POSIX paths, which do not match how Windows spells them, so I switched to matching the note tail instead.

The other failures seem unrelated to this change:
- 3 are FreeBSD, Test SwiftPM Linux, and Test SwiftPM macOS, the same known failures that were red on a recently merged PR.
- 3 are the 6.4.x Windows jobs hitting the broken` _FoundationCShims` header in that SDK.
- 1 is the macOS diagnosingMissingTargetDependencies failure, which looks like a pre-existing flaky test. It relies on a sleep to order two parallel targets, and it does not touch build configuration filters at all.

Could you give it another CI run please? I think we are basically there.

owenv:
@swift-ci test

dmhts:
@owenv Good news first, I do not see any build configuration filtering failures anymore.

I pulled the CI logs from [the ignore compiler verbosity in task signatures PR](https://github.com/swiftlang/swift-build/actions/runs/30193157285/job/90429578664) that merged about 20 hours ago and compared them to [my latest run](https://github.com/swiftlang/swift-build/actions/runs/30392280998/job/90430424799?pr=1427). They fail the exact same set of checks, 8 each:

- The Build Verdicts summaries are identical.
- `3` are the 6.4.x Windows jobs hitting the broken `_FoundationCShims` header in that SDK.
- `2` are Test SwiftPM Linux and macOS, the same failing tests (lldbRun, libFuzzerSupport, migrateCommand, enableDisableTestabilityDisabled).
- `1` is FreeBSD `dynamicLibraryConsumingObjectLibraryWithCodeCoverage`.
- `1` is a Windows nightly-main file-lock race during the build. On the merged PR it was the WebAssembly bundle, on mine it is the test bundle. Different file each run, before any test runs.

To me this looks good to do. All the remaining reds show up on a PR that already landed, unless you still have any concerns?

Sources/SWBProjectModel/PIFGenerationModel.swift
```diff
@@ -242,11 +242,13 @@ public enum PIF {
...
             /// The platform filters for this target dependency.
             public let platformFilters: Set<PlatformFilter>
+            /// The build configuration filters for this target dependency.
+            public let buildConfigurationFilters: Set<BuildConfigurationFilter>
         }
 
         /// Adds a dependency on another target.  It is the caller's responsibility to avoid creating dependency cycles.  A dependency of one target on another ensures that the other target is built first.
-        public func addDependency(on targetId: String, platformFilters: Set<PlatformFilter>) {
-            dependencies.append(TargetDependency(targetId: targetId, platformFilters: platformFilters))
+        public func addDependency(on targetId: String, platformFilters: Set<PlatformFilter>, buildConfigurationFilters: Set<BuildConfigurationFilter>) {
```
owenv, line 250: because this is part of the public ABI, we should keep the old version of this which doesn't accept build config filters around and just forward to the new version

Sources/SWBProjectModel/PIFGenerationModel.swift
```diff
@@ -427,16 +430,16 @@ public enum PIF {
         }
 
         /// Adds a dependency on another target.  It is the caller's responsibility to avoid creating dependency cycles.  A dependency of one target on another ensures that the other target is built first.  If `linkProduct` is true, the receiver will also be configured to link against the product produced by the other target (this presumes that the product type is one that can be linked against).
-        public func addDependency(on targetId: String, platformFilters: Set<PlatformFilter>, linkProduct: Bool) {
-            super.addDependency(on: targetId, platformFilters: platformFilters)
+        public func addDependency(on targetId: String, platformFilters: Set<PlatformFilter>, buildConfigurationFilters: Set<BuildConfigurationFilter>, linkProduct: Bool) {
```
owenv, line 437: Same comment about preserving the old version of the function for the time being and forwarding to the new one

Sources/SWBProjectModel/PIFGenerationModel.swift
```diff
@@ -457,30 +460,30 @@ public enum PIF {
         }
 
         /// Convenience function to add a file reference to the Frameworks build phase, after creating it if needed.
-        @discardableResult public func addLibrary(ref: FileReference, platformFilters: Set<PlatformFilter>, codeSignOnCopy: Bool = false, removeHeadersOnCopy: Bool = false) -> BuildFile {
+        @discardableResult public func addLibrary(ref: FileReference, platformFilters: Set<PlatformFilter>, buildConfigurationFilters: Set<BuildConfigurationFilter>, codeSignOnCopy: Bool = false, removeHeadersOnCopy: Bool = false) -> BuildFile {
```
owenv, line 471: same compatibility comment

Sources/SWBProjectModel/PIFGenerationModel.swift
```diff
@@ -457,30 +460,30 @@ public enum PIF {
...
-            self.addLibrary(ref: ref, platformFilters: [], codeSignOnCopy: codeSignOnCopy, removeHeadersOnCopy: removeHeadersOnCopy)
+            self.addLibrary(ref: ref, platformFilters: [], buildConfigurationFilters: [], codeSignOnCopy: codeSignOnCopy, removeHeadersOnCopy: removeHeadersOnCopy)
         }
 
         @_disfavoredOverload @discardableResult public func addLibrary(ref: FileReference) -> BuildFile {
             self.addLibrary(ref: ref, codeSignOnCopy: false)
         }
 
-        @discardableResult public func addResourceFile(ref: FileReference, platformFilters: Set<PlatformFilter>, resourceRule: BuildFile.ResourceRule? = nil) -> BuildFile {
+        @discardableResult public func addResourceFile(ref: FileReference, platformFilters: Set<PlatformFilter>, buildConfigurationFilters: Set<BuildConfigurationFilter>, resourceRule: BuildFile.ResourceRule? = nil) -> BuildFile {
```
owenv, line 488: same here

Sources/SWBProjectModel/PIFGenerationModel.swift
```diff
@@ -457,30 +460,30 @@ public enum PIF {
...
             let resourcesPhase = buildPhases.first{ $0 is CopyBundleResourcesBuildPhase } ?? addCopyBundleResourcesBuildPhase()
-            return resourcesPhase.addBuildFile(fileRef: ref, platformFilters: platformFilters, resourceRule: resourceRule)
+            return resourcesPhase.addBuildFile(fileRef: ref, platformFilters: platformFilters, buildConfigurationFilters: buildConfigurationFilters, resourceRule: resourceRule)
         }
 
-        @_disfavoredOverload @discardableResult public func addResourceFile(ref: FileReference, platformFilters: Set<PlatformFilter>) -> BuildFile {
-            self.addResourceFile(ref: ref, platformFilters: platformFilters, resourceRule: nil)
+        @_disfavoredOverload @discardableResult public func addResourceFile(ref: FileReference, platformFilters: Set<PlatformFilter>, buildConfigurationFilters: Set<BuildConfigurationFilter>) -> BuildFile {
```
owenv: same here

Sources/SWBProjectModel/PIFGenerationModel.swift
```diff
@@ -545,39 +549,39 @@ public enum PIF {
         }
 
         /// Adds a new build file that refers to `fileRef`.
-        @discardableResult public func addBuildFile(fileRef: FileReference, platformFilters: Set<PlatformFilter> = [], codeSignOnCopy: Bool = false, removeHeadersOnCopy: Bool = false, generatedCodeVisibility: BuildFile.GeneratedCodeVisibility? = nil, resourceRule: BuildFile.ResourceRule? = nil) -> BuildFile {
-            let buildFile = BuildFile(id: nextBuildFileId, reference: fileRef, platformFilters: platformFilters, codeSignOnCopy: codeSignOnCopy, removeHeadersOnCopy: removeHeadersOnCopy)
+        @discardableResult public func addBuildFile(fileRef: FileReference, platformFilters: Set<PlatformFilter> = [], buildConfigurationFilters: Set<BuildConfigurationFilter> = [], codeSignOnCopy: Bool = false, removeHeadersOnCopy: Bool = false, generatedCodeVisibility: BuildFile.GeneratedCodeVisibility? = nil, resourceRule: BuildFile.ResourceRule? = nil) -> BuildFile {
```
owenv, line 568: same here

Sources/SWBProjectModel/PIFGenerationModel.swift
```diff
@@ -545,39 +549,39 @@ public enum PIF {
...
             buildFile.generatedCodeVisibility = generatedCodeVisibility
             buildFile.resourceRule = resourceRule
             files.append(buildFile)
             return buildFile
         }
 
-        @_disfavoredOverload @discardableResult public func addBuildFile(fileRef: FileReference, platformFilters: Set<PlatformFilter> = [], codeSignOnCopy: Bool = false, removeHeadersOnCopy: Bool = false, generatedCodeVisibility: BuildFile.GeneratedCodeVisibility? = nil) -> BuildFile {
-            return addBuildFile(fileRef: fileRef, platformFilters: platformFilters, codeSignOnCopy: codeSignOnCopy, removeHeadersOnCopy: removeHeadersOnCopy, generatedCodeVisibility: generatedCodeVisibility, resourceRule: nil)
+        @_disfavoredOverload @discardableResult public func addBuildFile(fileRef: FileReference, platformFilters: Set<PlatformFilter> = [], buildConfigurationFilters: Set<BuildConfigurationFilter> = [], codeSignOnCopy: Bool = false, removeHeadersOnCopy: Bool = false, generatedCodeVisibility: BuildFile.GeneratedCodeVisibility? = nil) -> BuildFile {
```
owenv: same here

Sources/SWBProjectModel/PIFGenerationModel.swift
```diff
@@ -801,31 +819,33 @@ public enum PIF {
...
             self.id = id
             self.ref = .reference(id: reference.id)
             self.platformFilters = platformFilters
+            self.buildConfigurationFilters = buildConfigurationFilters
             self.codeSignOnCopy = codeSignOnCopy
             self.removeHeadersOnCopy = removeHeadersOnCopy
         }
-        @_disfavoredOverload public convenience init(id: String, reference: FileReference, platformFilters: Set<PlatformFilter>) {
-            self.init(id: id, reference: reference, platformFilters: platformFilters, codeSignOnCopy: false)
+        @_disfavoredOverload public convenience init(id: String, reference: FileReference, platformFilters: Set<PlatformFilter>, buildConfigurationFilters: Set<BuildConfigurationFilter>) {
```
owenv: This initializer should also be preserved and can forward to the new one

Sources/SWBProtocol/ProjectModel/BuildFile.swift
```diff
@@ -103,11 +105,12 @@ extension BuildFile: PendingSerializableCodable {
         self.shouldLinkWeakly = try deserializer.deserialize()
         self.assetTags = try deserializer.deserialize()
         self.platformFilters = try deserializer.deserialize()
+        self.buildConfigurationFilters = try deserializer.deserialize()
```
owenv: This code needs to be able to deserialize a BuildFile which may have been serialized by an older client-side framework. This is usually accomplished by adding the new field at the end and checking the deserializer's field count before trying to decode it

Sources/SWBProtocol/ProjectModel/TargetDependency.swift
```diff
@@ -16,29 +16,33 @@ public struct TargetDependency: Sendable {
...
+        self.buildConfigurationFilters = buildConfigurationFilters
     }
 }
 
 // MARK: SerializableCodable
 
 extension TargetDependency: PendingSerializableCodable {
     public func legacySerialize<T: Serializer>(to serializer: T) {
-        serializer.serializeAggregate(3) {
+        serializer.serializeAggregate(4) {
```
owenv, line 33: same note about serialization compatibility

Tests/SWBBuildServiceTests/BuildDependencyInfoTests.swift
```diff
@@ -558,6 +559,102 @@ import Foundation
         }
     }
 
+    @Test(.requireSDKs(.iOS)) func filteredFrameworkExcludedFromDependencyInfo() async throws {
```
owenv: I'm not sure this test is providing a lot of value, the new filter type doesn't meaningfully interact with this feature

Tests/SWBTaskConstructionTests/BuildFileFilteringInteractionTests.swift
```diff
@@ -0,0 +1,149 @@
...
+                results.checkNote("Skipping '/tmp/Test/aProject/Filtered.m' because its build configuration filter (\(buildConfigurationFiltersString)) does not match the build configuration filter of the current context (\(buildConfiguration)). (in target 'AppTarget' from project 'aProject')")
+            }
+
+            results.checkNoTask(sourceLocation: sourceLocation)
+            results.checkNoDiagnostics()
+        }
+    }
+
+    /// Both filters match the current context. The file is included.
+    @Test(.requireSDKs(.macOS))
```
owenv, line 111: It would be nice to run this with any host SDK, but since it's not trivial to do that I think it' alright to leave as is for now

owenv (commented):
Thanks, this all looks good to me. I've left one last set of comments about ensuring we're not removing public API, they should all be mechanical changes. Once those are addressed I think we can go ahead and land this.

Because SE-0273 is (much) older than the 1 year expiry date for unimplemented components of proposals, I think the SwiftPM changes will need to go through the evolution process before landing. Because the changes in this PR are only touching internal implementations, I'm ok merging this one before going through evolution though.

Tests/SWBTaskConstructionTests/BuildFileFilteringInteractionTests.swift
dmhts: Yeah, I did look into it though. The blocker is that these need a platform filter matching whichever host is running, and test support only has macOS and Linux helpers today. So it would mean adding windows and freebsd ones plus some kind of host helper. Hard to verify any of that locally, so it would probably need a bit more CI dance to get right. Happy to do it as a follow up if you think it is worth having.

Tests/SWBBuildServiceTests/BuildDependencyInfoTests.swift
dmhts: Agreed, [removed](https://github.com/swiftlang/swift-build/pull/1427/changes/5b89e5c0ba2ce527f4a23755727854d0982faa6f). Checking it properly, the wiring here is compiler enforced anyway.

Sources/SWBProtocol/ProjectModel/BuildFile.swift
dmhts: [Both fixed](https://github.com/swiftlang/swift-build/pull/1427/changes/494ce1af1b149fc9ea1ea1a4b8544f67270217da). There were no tests covering legacy deserialization before, so I added some. They build an old style payload and deserialize it, and I verified they would have caught this.

Sources/SWBProjectModel/PIFGenerationModel.swift
dmhts: Done. Every prior signature is kept and forwards to the new `buildConfigurationFilters` version, and I applied that to all the changed overloads and initializers rather than just the flagged ones. That is 8 new `buildConfigurationFilters` overloads in total (6 methods and 2 initializers), each keeping its prior signature.

owenv (approved):

owenv:
@swift-ci test

dmhts:
Thanks for your patience and the thorough review, Owen, it's been fun learning some of the swift-build internals along the way.

Makes sense on SE-0273. I've already started writing up a follow-up proposal for the SwiftPM part. Thanks for being ok to land this one first, it should help get that through a bit quicker.

PR #1431, opened 2026-05-26
Ensure linux platform filters match linux-musl triples

owenv:
This ensures the right platform conditions are active when building with the static linux SDK


owenv:
@swift-ci test

jakepetroules (approved):
I wonder if we'll want a way to differentiate at some point, but I guess this seems correct for now.

owenv:
@swift-ci test

PR #1428, opened 2026-05-26
Argument Parser 1.8.0 now does parsing async

finagolfin:
I've had to patch this locally for the last couple months of using the main branch of Argument Parser, which we'll need merged now that the toolchain is being updated to use Argument Parser 1.8.0, swiftlang/swift#89404, with this change most likely tied to apple/swift-argument-parser#855.

PR #1432, opened 2026-05-27
fix: add missing `await`

Brett-Best:
Fixes:
```
SwiftCompile normal arm64 /Users/brettbest/Developer/swift-build/Sources/SWBUtil/PbxCp.swift (in target 'SWBUtil' from project 'SwiftBuild')
    cd /Users/brettbest/Developer/swift-build/.swiftpm/xcode
    

/Users/brettbest/Developer/swift-build/Sources/SWBUtil/PbxCp.swift:571:19: error: expression is 'async' but is not marked with 'await'
        options = try CopyOptions.parse(argv)
                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~
                      await 
/Users/brettbest/Developer/swift-build/Sources/SWBUtil/PbxCp.swift:571:23: note: call is 'async'
        options = try CopyOptions.parse(argv)
```

I believe this was caused by https://github.com/apple/swift-argument-parser/commit/d6f4e7a371cbd322a89305ae831d36a47d0e3357 and the dependency constraint for it is `from: "1.0.3"` meaning `1.8.0` satisfies this - hence it breaking.

If a lower version is checked out now a warning of: `No 'async' operations occur within 'await' expression` is generated.

I thought the below might work:
```swift
        #if canImport(ArgumentParser, _version: 1.8.0)
        options = try await CopyOptions.parse(argv)
        #else
        options = try CopyOptions.parse(argv)
        #endif
```
But I get: https://docs.swift.org/compiler/documentation/diagnostics/module-version-missing/

I thought I saw SwiftPM would set this but perhaps that hasn't shipped yet or I imagined it :D

### DIFF
--- a/Sources/SWBUtil/PbxCp.swift
+++ b/Sources/SWBUtil/PbxCp.swift
@@ -568,7 +568,7 @@ public func pbxcp(_ argv: [String], cwd: Path) async -> (success: Bool, output: 
     do {
         var argv = argv
         argv.removeFirst()
-        options = try CopyOptions.parse(argv)
+        options = try await CopyOptions.parse(argv)
     } catch {
         return (false, CopyOptions.message(for: error))
     }


Brett-Best:
Closing in favour of https://github.com/swiftlang/swift-build/pull/1428 which I missed when looking through the list before making this 🥲

PR #1433, opened 2026-05-27
Update Xcode version to 26.4 in GitHub Actions workflows

owenv:
Runners no longer have 26.0 installed, so this should fix a handful of the failures seen in CI


owenv:
@swift-ci test

mhrawdon (approved):

PR #1434, opened 2026-05-28
Skip swiftAstPathForStaticLibrariesMultiPlatform if using EBM-based debug info

owenv:
Skip this test if the relevant feature is enabled because we don't expect to be using -add_ast_path in that case


owenv:
@swift-ci test

jakepetroules (approved):

owenv:
@swift-ci test windows

PR #1441, opened 2026-06-03
Strip shell session env vars from BuildDescription signature.

rjmansfield:
OLDPWD and TERM_SESSION_ID change across terminal/shell sessions but don't affect build output. Including them in the signature can invalidate the on-disk BuildDescription cache for cosmetic differences, producing redundant bundle directories per terminal session.

### DIFF
--- a/Sources/SWBTaskExecution/BuildDescriptionSignature.swift
+++ b/Sources/SWBTaskExecution/BuildDescriptionSignature.swift
@@ -93,7 +93,7 @@ package struct BuildDescriptionSignatureComponents: Codable, Hashable, Sendable 
                 macroConfigSignature: request.buildRequestContext.getCachedSettings(request.buildRequest.parameters, project: $0).inputPathsAffectingSettingsSignature)
         }
         systemInfo = request.workspaceContext.systemInfo
-        userInfo = request.workspaceContext.userInfo
+        userInfo = request.workspaceContext.userInfo?.filteredUserInfoForSignature
         developerPath = request.workspaceContext.core.developerPath.path
         xcodeVersionString = request.workspaceContext.core.xcodeVersionString
         xcodeProductBuildVersionString = request.workspaceContext.core.xcodeProductBuildVersionString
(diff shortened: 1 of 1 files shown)


rjmansfield:
@swift-ci please test

owenv (approved):
LGTM, would be good to get a second opinion though in case anyone can think of a better way to maintain the list

Sources/SWBTaskExecution/BuildDescriptionSignature.swift
```diff
@@ -106,10 +106,31 @@ package struct BuildDescriptionSignatureComponents: Codable, Hashable, Sendable
...
+
 extension BuildDescriptionSignatureComponents {
     var humanReadableString: ByteString {
         get throws {
-            try ByteString(JSONEncoder(outputFormatting: [.prettyPrinted, .sortedKeys, .withoutEscapingSlashes]).encode(self))
+            var components = self
+            if let ui = components.userInfo {
+                components.userInfo = UserInfo(
+                    user: ui.user, group: ui.group, uid: ui.uid, gid: ui.gid, home: ui.home,
+                    processEnvironment: Self.filteredEnvironmentForSignature(ui.processEnvironment),
```
jakepetroules: I think you need to do this in the initializer, not after the fact. Because you want the hashable and == to be based on the filtered view, not the original. The purpose of BuildDescriptionSignatureComponents is that it encodes _exactly_ what the key is that will be used for comparison.

jakepetroules (changes requested):
> LGTM, would be good to get a second opinion though in case anyone can think of a better way to maintain the list

I can't. Concerns on the implementation though.

Sources/SWBTaskExecution/BuildDescriptionSignature.swift
```diff
@@ -107,6 +107,24 @@ package struct BuildDescriptionSignatureComponents: Codable, Hashable, Sendable
...
 extension BuildDescriptionSignatureComponents {
+    /// Env vars stripped from `UserInfo` before hashing into the signature.
+    /// These change across terminal/shell sessions but don't affect build output
+    /// and including them can invalidate the on-disk cache for cosmetic differences.
+    fileprivate static let environmentDenylist: Set<String> = [
+        "OLDPWD",
+        "TERM_SESSION_ID",
+    ]
+
+    fileprivate static func filteredUserInfoForSignature(_ ui: UserInfo) -> UserInfo {
```
jakepetroules: nit: would be shorter and more future proof as an extension on UserInfo (no need to update this call site if the constructor changes in future):

```swift
extension UserInfo {
    fileprivate var filteredUserInfoForSignature: Self {
        var ui = self
        for key in environmentDenylist {
           ui.processEnvironment.removeValue(forKey: key)
           ui.buildSystemEnvironment.removeValue(forKey: key)
        }
        return ui
    }
}
```

rjmansfield:
@swift-ci please test

jakepetroules (approved):

PR #1447, opened 2026-06-06
Use access level on import in resource_bundle_accessor.swift

### LINKED ISSUE #10163: Ambiguous implicit access level for Foundation import in `resource_bundle_accessor.swift`
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

When a Swift package specifies a resource as part of a build target, SwiftPM synthesizes a `resource_bundle_accessor.swift` to provide runtime hooks to access the resource. With Swift 6.1, if the target uses access levels on import (internal/private/public import), then the target fails to build with:

```
error: ambiguous implicit access level for import of 'Foundation'
```

Once any other file in the same module declares an access-qualified import, the compiler requires every import in the module (including this generated one) to be explicit. As users cannot modify this file, the only workaround for this issue is to make the Foundation import in the package public.

### Expected behavior

The generated `resource_bundle_accessor.swift` declares an access level for its imports, and allows the target to build successfully.

### Actual behavior

The build fails with:

```
error: ambiguous implicit access level for import of 'Foundation'
```

### Steps to reproduce

1. Create a package with `swift-tools-version: 6.1`.
2. Add a target with `resources: [.copy("SomeFolder")]` so SwiftPM synthesizes `resource_bundle_accessor.swift`.
3. In one of that target's source files, add `internal import Foundation`.
4. Run `swift build`.

snprajwal:
When a Swift package specifies a resource as part of a build target, SwiftPM synthesizes a resource_bundle_accessor.swift to provide runtime hooks to access the resource. This generated file imports Foundation but does not specify an access level. With Swift 6.0, access-level-on-import is part of the language mode, meaning that if a target specified a resource and contained a non-public import of Foundation, the build would fail with:

```
error: ambiguous implicit access level for import of 'Foundation'
```

This patch updates the generation logic to include an explicit `public` access level modifier for the Foundation import in `bundle_resource_accessor.swift` if the Swift version is >= 6.0.

Fixes swiftlang/swift-package-manager#10163

## Additional testing

1. Create a package with `swift-tools-version: 6.0`.
2. Add a target with resources: [.copy("SomeFolder")] so SwiftPM synthesizes resource_bundle_accessor.swift.
3. In one of that target's source files, add `internal import Foundation`.
4. Run `swift build --build-system swiftbuild`.

Without this patch, the build fails with the above specified error. With this patch, the build succeeds.

### DIFF
--- a/Sources/SWBTaskConstruction/TaskProducers/BuildPhaseTaskProducers/SourcesTaskProducer.swift
+++ b/Sources/SWBTaskConstruction/TaskProducers/BuildPhaseTaskProducers/SourcesTaskProducer.swift
@@ -2021,8 +2021,18 @@ package final class SourcesTaskProducer: FilesBasedBuildPhaseTaskProducerBase, F
     private func generatePackageTargetBundleAccessorForSwift(_ scope: MacroEvaluationScope, bundleName: String) async -> GeneratedSourceCodeResult {
         let filePath = scope.evaluate(BuiltinMacros.DERIVED_SOURCES_DIR).join("resource_bundle_accessor.swift")
 
+        // Emit an explicit access level on imports if access-level-on-import is part of the language mode (>= 6.0).
+        // Without this, targets that use access levels on imports will fail to compile.
+        let importPrefix: String = {
+            guard let swiftVersion = try? Version(scope.evaluate(BuiltinMacros.SWIFT_VERSION)),
+                  swiftVersion >= Version(6) else {
+                return ""
+            }
+            return "public "
+        }()
+
         let contents = bundleName.isEmpty ? """
-            import class Foundation.Bundle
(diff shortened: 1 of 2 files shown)


snprajwal:
@swift-ci please test

owenv (changes requested):
Determining the access level based on the language mode alone would be a breaking change. Realistically, I think this will probably need to become configurable via the package manifest and go through the evolution process

snprajwal:
That makes sense, thanks. I'll pitch a better fix for this on the forums, and take it forward from there. Closing this PR for now.

PR #1449, opened 2026-06-09
[6.4] macOS 27.0 architecture updates

owenv:
macOS 27.0 no longer supports x86_64. Mark the architecture deprecated so we only build an x86 slice by default when using a deployment target macOS 26 or earlier. This PR also updates a number of tests to account for this change.

### DIFF
--- a/Sources/SWBApplePlatform/Specs/DriverKit.xcspec
+++ b/Sources/SWBApplePlatform/Specs/DriverKit.xcspec
@@ -82,6 +82,7 @@
         Name = "Intel 64-bit";
         Description = "64-bit Intel";
         PerArchBuildSettingName = "Intel 64-bit";
+        DeploymentTargetRange = ( "0", "27" );
     },
     {
         _Domain = driverkit;
--- a/Sources/SWBApplePlatform/Specs/macOSArchitectures.xcspec
+++ b/Sources/SWBApplePlatform/Specs/macOSArchitectures.xcspec
@@ -79,6 +79,7 @@
 		PerArchBuildSettingName = "Intel 64-bit";
 		ListInEnum = NO;
 		SortNumber = 106;
+        DeploymentTargetRange = ( "10.4.4", "27" );
 	},
     {
         _Domain = macosx;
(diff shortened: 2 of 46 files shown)


owenv:
Starting as a draft while I confirm all the test updates are correct

owenv:
@swift-ci test

Sources/SWBApplePlatform/Specs/DriverKit.xcspec
```diff
@@ -82,6 +82,7 @@
         Name = "Intel 64-bit";
         Description = "64-bit Intel";
         PerArchBuildSettingName = "Intel 64-bit";
+        DeploymentTargetRange = ( "0", "27" );
```
jakepetroules, line 85: Super pedantic but... since DriverKit 26 doesn't exist and is aliased to DriverKit 27, should the upper bound be set to 26 so that erroneously specified deployment targets don't try to build for x86_64?

"that's enough of an edge case that we can ignore it" is also an acceptable answer.

jakepetroules (approved):

mhrawdon (approved):

owenv:
@swift-ci test

Sources/SWBApplePlatform/Specs/DriverKit.xcspec
owenv: That feels a bit too subtle for me, I think I'd almost prefer diagnosing any attempt to use the nonexistent version

jakepetroules: Agree. Just wanted to call it out.

owenv:
@swift-ci test

owenv:
@swift-ci test macOS

owenv:
@swift-ci test linux

owenv:
@swift-ci test

owenv:
@swift-ci test windows

owenv:
@swift-ci test

owenv:
@swift-ci test macOS

owenv:
@swift-ci test linux

PR #1450, opened 2026-06-10
[6.4]Add support for new screen-capture background mode in Info.plist on Apple platforms

owenv:
This is modeled exactly the same as the other existing background modes


owenv:
@swift-ci test

jakepetroules (approved):

PR #1464, opened 2026-06-17
[6.4] Tag tests which require specific platforms to be loaded when cross-compiling

owenv:
Fix some test failures which occur when using XCODE_ONLY_PLATFORM_FAMILY_NAMES to prevent plugins from registering new platforms.

rdar://175558887


owenv:
@swift-ci test

jakepetroules (approved):

jakepetroules:
would be nice if we could collapse requireSDK and requirePlatform somehow

PR #1465, opened 2026-06-17
Generate XOJIT preview thunks for source files reached through a symlink

pmattos:
### Context

When SwiftUI Previews run in XOJIT mode, the previews client asks Swift Build — per source file — for the compile command that builds that file's "preview thunk", via `generatePreviewInfo(.thunkInfo)`, identifying the file by path. In `SwiftCompilerSpec.generatePreviewInfo`, Swift Build locates that path among the target's compile inputs and hands it to `LibSwiftDriver.frontendCommandLine` as the primary input, which produces the `swift-frontend` invocation for the thunk.

That lookup compares paths by exact string. If the source file was recorded in the build description under a different — but equivalent — spelling than the one in the request, nothing matches: `frontendCommandLine` returns no command line, `generatePreviewInfo` returns an empty result, and the client reports `noPreviewInfos`, so the preview never renders.

This happens for standalone Swift packages whose sources live under a symlinked directory. On macOS, temporary directories such as `/tmp` are symlinks to `/private/tmp`, and SwiftPM's new package PIF builder symlink-resolves source paths (to match what the index store records). The compile command then holds `/private/tmp/.../View.swift` while the preview request still carries the unresolved `/tmp/.../View.swift` — the same file, a different spelling, no match.

### Fix

In the XOJIT branch of `SwiftCompilerSpec.generatePreviewInfo`, match the requested source file against the command's inputs by resolved path (ie, `FSProxy.realpath`), and use the spelling that actually appears in the command as the primary input. The output-file-map and the VFS overlay are keyed on that same spelling so they line up with what the driver resolves. When the requested and recorded paths already agree — e.g. the legacy package PIF builder, which doesn't resolve symlinks — this is a no-op.

### Testing

Adds `previewXOJITThunkInfoResolvesSymlinkedSourcePath` to `PreviewsBuildOperationTests`. Rather than standing up a real package, it reproduces the spelling mismatch with a single symlink: it builds a one-file app under a real directory and then requests the preview thunk through a symlink to that directory, so the requested path and the build's input path resolve to the same file but differ as strings. It asserts that exactly one preview info comes back and that the returned compile command references the real (build) spelling, never the symlinked request path.

Without the fix the test returns zero preview infos (ie, the `noPreviewInfos` condition); with it one, as expected.

Fixes rdar://176386125 (and related to rdar://179887248).

### DIFF
--- a/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
@@ -3237,7 +3237,7 @@ public final class SwiftCompilerSpec : CompilerSpec, SpecIdentifierType, SwiftDi
             // See the related `previewXOJITThunkInfoResolvesSymlinkedSourcePath` test.
             func resolvedPath(_ path: Path) -> Path { (try? fs.realpath(path)) ?? path }
             let realSourceFile = resolvedPath(sourceFile)
-            selectedInputPath = originalInputs.first { resolvedPath($0) == realSourceFile } ?? sourceFile
+            selectedInputPath = originalInputs.only { resolvedPath($0) == realSourceFile } ?? sourceFile
 
             if let driverPayload = payload.driverPayload {
                 do {
--- a/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
@@ -3224,13 +3224,29 @@ public final class SwiftCompilerSpec : CompilerSpec, SpecIdentifierType, SwiftDi
         if payload.previewStyle == .xojit {
             // Also pass the auxiliary Swift files.
             commandLine.append(contentsOf: originalInputs.map(\.str))
-            selectedInputPath = sourceFile
+
+            // Fixes rdar://176386125. The package PIF builder may symlink-resolve source paths in
+            // the build description (eg, /tmp -> /private/tmp) so they match what the index service stores.
+            // The previews client, however, asks us to thunk the *unresolved* path, so a plain `sourceFile` won't
+            // match the command's inputs and libSwiftDriver returns no command line -> `noPreviewInfos`.
+            //
+            // Use the spelling that actually appears in the command's inputs, matched by realpath,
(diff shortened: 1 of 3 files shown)


pmattos:
@swift-ci test

Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
```diff
@@ -3226,21 +3224,37 @@ public final class SwiftCompilerSpec : CompilerSpec, SpecIdentifierType, SwiftDi
...
+            // match the command's inputs and libSwiftDriver returns no command line -> `noPreviewInfos`.
+            //
+            // Use the spelling that actually appears in the command's inputs, matched by realpath,
+            // and key the output-file-map / VFS overlay on it. No-op when the paths already agree
+            // (eg, the legacy package PIF builder available in Xcode).
+            //
+            // See the related `previewXOJITThunkInfoResolvesSymlinkedSourcePath` test.
+            func resolvedPath(_ path: Path) -> Path { (try? fs.realpath(path)) ?? path }
+            let realSourceFile = resolvedPath(sourceFile)
+            selectedInputPath = originalInputs.first { resolvedPath($0) == realSourceFile } ?? sourceFile
```
jakepetroules: Suspicious of the `first(where:)` usage here. What happens if there are multiple matches? Is the order of `originalInputs` actually meaningful here?

pmattos: There's at most one match in practice, so the ordering doesn't really matter here.

The `originalInputs` is this compile's list of Swift source files, and a target doesn't list the same source *twice* — so at most one input can resolve to the same file as `sourceFile`. Multiple matches would require two distinct entries that point at the same file on disk (eg, a path *and* a symlink to it both present as inputs), which a normal Swift compilation doesn't produce AFAIK.

If that degenerate case ever did occur, the matches would all be the same file anyway, so `first` is still correct here. What we actually need is just the *spelling* of the requested source as it appears in the command's inputs, so libSwiftDriver's primary-input lookup and the output-file-map/VFS overlays keys all agree on the exact path.

jakepetroules (approved):

pmattos:
@swift-ci test

pmattos:
@swift-ci test

PR #1473, opened 2026-06-22
Load SDK extended info before publishing to the registry

rjmansfield:
Fixes a SWBBuildService crash where concurrent SDK lookups could observe an SDK published in sdksByCanonicalName before its caller had run loadExtendedInfo, tripping the wrapper's "accessed before it was initialized" precondition.

Adds a loadingWith: parameter to SDKRegistry.registerSDK that runs loadExtendedInfo before publishing.

rdar://180151783

### DIFF
--- a/Sources/SWBCore/SDKRegistry.swift
+++ b/Sources/SWBCore/SDKRegistry.swift
@@ -216,17 +216,33 @@ public final class SDK: Sendable {
     ]
 
     /// Perform late binding of the SDK data.
+    ///
+    /// Parses all settings tables first; only commits results to the wrappers if every
+    /// parse succeeded. On parse failure, every wrapper is set to nil so callers that
+    /// keep the SDK in the registry can safely query an empty SDK.
     fileprivate func loadExtendedInfo(_ namespace: MacroNamespace) throws {
+        let defaultTable: MacroValueAssignmentTable
+        let overrideTable: MacroValueAssignmentTable
+        var variantTables: [(SDKVariant, MacroValueAssignmentTable)] = []
         do {
-            _defaultSettingsTable.initialize(to: try namespace.parseTable(defaultSettings, allowUserDefined: true, associatedTypesForKeysMatching: associatedTypesForKeysMatching))
-            _overrideSettingsTable.initialize(to: try namespace.parseTable(overrideSettings, allowUserDefined: true, associatedTypesForKeysMatching: associatedTypesForKeysMatching))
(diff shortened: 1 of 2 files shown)


rjmansfield:
@swift-ci test

Sources/SWBCore/SDKRegistry.swift
```diff
@@ -579,7 +593,8 @@ public final class SDKRegistry: SDKRegistryLookup, CustomStringConvertible, Send
     /// - parameter path: The nominal (given) path to the SDK directory.
     /// - parameter pathResolved: The resolved (normalized) path to the SDK directory.
     /// - parameter platform: The platform in which to register the SDK, if any.
-    @discardableResult private func registerSDK(_ path: Path, _ pathResolved: Path, _ platform: Platform?) -> SDK? {
+    /// - parameter loadingWith: If non-nil, runs `loadExtendedInfo` before publishing the SDK and returns nil on parse failure. Required on any path that registers SDKs concurrently so lookups never observe a partially initialized SDK.
+    @discardableResult private func registerSDK(_ path: Path, _ pathResolved: Path, _ platform: Platform?, loadingWith namespace: MacroNamespace? = nil) -> SDK? {
```
mhrawdon: This is a nit, but I might remove the `loadingWith:` label and replace it with `_` for symmetry with the other parameters. If you think it's important to have a label, I'd just call it `namespace:`.

Sources/SWBCore/SDKRegistry.swift
```diff
@@ -579,7 +593,8 @@ public final class SDKRegistry: SDKRegistryLookup, CustomStringConvertible, Send
     /// - parameter path: The nominal (given) path to the SDK directory.
     /// - parameter pathResolved: The resolved (normalized) path to the SDK directory.
     /// - parameter platform: The platform in which to register the SDK, if any.
-    @discardableResult private func registerSDK(_ path: Path, _ pathResolved: Path, _ platform: Platform?) -> SDK? {
+    /// - parameter loadingWith: If non-nil, runs `loadExtendedInfo` before publishing the SDK and returns nil on parse failure. Required on any path that registers SDKs concurrently so lookups never observe a partially initialized SDK.
```
mhrawdon: It looks like this is passed when an SDK is loaded via `lookup()` from a path, and not otherwise? I.e., when the SDK is loaded at some post-initialization time rather than during the scan of directories on launch? That makes sense (if I'm right), but making the comment a little more specific might help make it clearer.

mhrawdon (approved):

rjmansfield:
@swift-ci test

rjmansfield:
@swift-ci test

PR #1477, opened 2026-06-23
Quote LinkFileList paths for non-Apple linkers (Issue #13)

### LINKED ISSUE #13: LinkFileList generation needs to use quoting
Non-Apple linkers require quoting in order to handle projects with spaces in their path.

This is handled in LdLinkerSpec when evaluating `__INPUT_FILE_LIST_PATH__`.

rdar://132107054

aanyabharti101:
## Why was this PR needed?

Issue #13 reports that LinkFileList entries are generated without quoting. When project paths contain spaces, non-Apple linkers may interpret those spaces as argument separators rather than part of the file path, causing linking failures.

During investigation, I traced LinkFileList generation through `LinkerTools.swift` and the existing response-file formatting infrastructure. I found that quoting support already existed in `ResponseFiles.swift`, but LinkFileList generation was not consistently selecting a quoted format for non-Apple linkers.

## What does this PR do?

This PR updates LinkFileList generation to use a quoted response-file format for non-Apple linkers while preserving the existing behavior for Apple's ld64 linker.

The implementation determines the linker type during task construction and selects `unixShellQuotedNewlineSeparated` when appropriate. Rather than introducing a new quoting mechanism, the change reuses the project's existing response-file formatting infrastructure.

The PR also adds regression coverage for LinkFileList generation and additional unit-test coverage for quoted newline-separated response-file formatting.

## What are the relevant issue numbers?

Closes #13

### Before

During reproduction, LinkFileList entries for project paths containing spaces were emitted without quoting, for example:

/tmp/Test/a Project/Object.o

### After

The implementation now selects the `unixShellQuotedNewlineSeparated` response-file format for non-Apple linkers.

Automated coverage was added through:

- `responseFileContentUnixNewlineQuotedPaths()`, which verifies quoted newline-separated output
- `linkFileListQuotingForNonAppleLinker()`, a regression test covering LinkFileList generation for non-Apple linkers

### Test Output

```text
✔ Test responseFileContentUnixNewlineQuotedPaths() passed
✔ Suite ResponseFileTests passed

➜ Test linkFileListQuotingForNonAppleLinker() skipped:
"This test requires a Linux host OS."
```

## Does this PR meet the acceptance criteria?

* [x] Tests added for new/changed behavior
* [x] All relevant tests passing
* [x] Follows project style guide
* [x] No breaking changes introduced
* [x] Documentation updated where applicable


aanyabharti101:
Hi @jakepetroules, I just wanted to follow up on this PR in case it was missed. This is my first contribution to Swift Build, and I'd appreciate any feedback when you have time. Thanks for taking a look!

jakepetroules:
@swift-ci test

aanyabharti101:
Hi @jakepetroules, thanks for running CI! I tracked down the Linux test failure, updated the regression test to account for the Linux-specific build directory, and pushed a follow-up commit. I'd appreciate another look when you have time. Thanks!

PR #1480, opened 2026-06-24
Serialize llbuild builds per database directory to prevent concurrent engines

daveinglis:
SWBBuildService could crash with heap corruption inside llbuild::core::BuildEngine::build when two builds drove the same on-disk build database with two llbuild engines at once. Mutual exclusion was provided by SystemCacheEntry.serializationQueue, which only serializes builds that share the same cache entry; that breaks down when the cached build system is evicted and recreated between builds (cycle detection or a clean, or any other eviction of the entry), letting a restart run concurrently with the build it replaced.

Serialize builds with a process-global AsyncOperationQueue (concurrency 1) keyed by the build-database directory, so mutual exclusion survives cache eviction/recreation and holds across caches. Remove the now-unused SystemCacheEntry.serializationQueue. Add a debug-only invariant assert that trips if two engines ever build one database, and a deterministic regression test.

Also mirror the AsyncOperationQueue concurrency-limit and deadlock fix from swift-package-manager (the file is maintained in both repos): once a parked task was promoted it was removed from waitingTasks, so the admission gate under-counted running tasks and admitted extra ones; the continuation re-check also counted the .creating task against itself and could deadlock. Promote tasks to .running in place, re-check against the running-task count only, and stop onCancel from tombstoning an already-running task.

rdar://176791560

### DIFF
--- a/Sources/SWBBuildSystem/BuildOperation.swift
+++ b/Sources/SWBBuildSystem/BuildOperation.swift
@@ -158,6 +158,60 @@ package extension BuildSystemOperation {
     }
 }
 
+/// Process-global serialization of builds by on-disk build-database directory.
+///
+/// The same database must never be driven by two llbuild engines concurrently.
+/// A per-`SystemCacheEntry` queue can only enforce this while two builds share
+/// the *same* cache entry; it breaks down when an entry is evicted (e.g.
+/// `clearCachedBuildSystem` on cycle detection or by `CleanOperation`) and a
+/// restart re-inserts a fresh entry, or when two builds come from different
+/// caches. Keying serialization on the database directory —
+/// independent of cache-entry lifetime and of which build-system cache drove the
+/// build — guarantees at most one engine operates a given database at a time,
+/// which two engines over one database would otherwise corrupt. rdar://176791560.
+private let buildDatabaseSerializationQueues = SWBMutex<[Path: AsyncOperationQueue]>([:])
+
+/// Returns the serialization queue (concurrency 1) for `databaseDirectory`,
+/// creating it on first use. Queues persist for the lifetime of the process so
+/// they survive cache-entry eviction/recreation.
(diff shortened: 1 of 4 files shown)


Sources/SWBBuildSystem/BuildOperation.swift
```diff
@@ -158,6 +158,62 @@ package extension BuildSystemOperation {
...
+/// The same database must never be driven by two llbuild engines concurrently.
+/// A per-`SystemCacheEntry` queue can only enforce this while two builds share
+/// the *same* cache entry; it breaks down when an entry is evicted (e.g.
+/// `clearCachedBuildSystem` on cycle detection or by `CleanOperation`) and a
+/// restart re-inserts a fresh entry, or when two builds come from different
+/// caches. Keying serialization on the database directory —
+/// independent of cache-entry lifetime and of which build-system cache drove the
+/// build — guarantees at most one engine operates a given database at a time,
+/// which two engines over one database would otherwise corrupt. rdar://176791560.
+private actor BuildDatabaseSerializer {
```
jakepetroules: actor seems overkill for this, just use a Mutex?

Sources/SWBBuildSystem/BuildOperation.swift
```diff
@@ -158,6 +158,62 @@ package extension BuildSystemOperation {
...
+        }
+        let queue = AsyncOperationQueue(concurrentTasks: 1)
+        queues[databaseDirectory] = queue
+        return queue
+    }
+}
+
+/// Diagnostic guard for the serialization invariant above: at most one llbuild
+/// engine should be actively building a given database directory at a time. This
+/// is a debug-only `assert` (compiled away in release) that turns a regression
```
jakepetroules: Why not make it a fatalError? If it's an invariant we should probably enforce in release too

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test

owenv (approved):

PR #1483, opened 2026-06-25
Only declare DocC symbol-graph inputs for cohort base archs

rjmansfield:
Cohort member archs fold into their base arch's compile via -target-arch-variant and emit no symbol graph, so iterate ARCHS_BASE rather than ARCHS when declaring DocC inputs to avoid referencing a *.symbols.json that nothing produces.


rjmansfield:
@swift-ci test

jakepetroules (approved):

PR #1486, opened 2026-06-25
#bundle should error when used from a static framework

matthewseaman:
Set `SWIFT_MODULE_RESOURCE_BUNDLE_UNAVAILABLE` for static frameworks so that `#bundle` produces an error rather than silently doing the wrong thing.

Unlike mergeable libraries, a regular framework with its `MACH_O_TYPE` set to `staticlib` does not provide a bundle hook that `#bundle` could use to get access to the resources.


matthewseaman:
@swift-ci please test

matthewseaman:
Does this seem unrisky enough to pull into whatever Swift branch is aligned with Xcode 27?

Tantalum73 (approved):

owenv (approved):

PR #1493, opened 2026-06-28
CustomTasks in AggregateTargets and alwaysOutOfData flag

dschaefer2:
Add support for CustomTasks in AggregateTargets. This allows us more flexibility for plugins to contribute to the build graph.

Add support for an always out of data flag to allow CustomTasks with no inputs to always run. This is used by SwiftPM's prototype external builder support to allow the builder to be added as a CustomTask and plug into the build graph. The builder handle's its own incremental builds.


Sources/SWBTaskConstruction/ProductPlanning/ProductPlanner.swift
```diff
@@ -457,7 +457,17 @@ extension AggregateTarget
...
     {
         // TODO: We should probably check that only build phases useful in an aggregate target are present here.
-        return super.buildPhaseTargetTaskProducers(taskProducerContext).taskProducers
+        let (buildPhaseTaskProducers, buildPhasesEndNode) = super.buildPhaseTargetTaskProducers(taskProducerContext)
+        var taskProducers = buildPhaseTaskProducers
+
+        // Custom tasks are ordered after the build phases.  The producer creates a phase end task which
+        // must precede the target end task, so that building the target forces the custom tasks to run.
+        if !customTasks.isEmpty {
+            let customTasksEndNode = taskProducerContext.createVirtualNode("\(phaseNodeRoot(taskProducerContext.configuredTarget))-CustomTaskProducer")
```
owenv, line 467: Any particular reason these depend on the build phases (in an aggregate target these will really only ever be script and copy phases). Ideally I think these would just run in parallel with everything else

dschaefer2: Right now it's just a hack to get Aggregate Targets to schedule Custom Tasks. It was suggested to me by a "friend" :). Back over in the External Packages prototype PR, I've integrated the external build plugin tools as Custom Tasks and hosted off of aggregate targets. That may end up being a bad choice and happy to be pointed in the right direction. So far, I've just been hacking to get something working.

dschaefer2: I think this also leads towards custom targets, where the product of the target is provided by a custom task. Although in these cases the plugins are currently providing libraries, similar to what we've done with binary targets, system targets and prebuilts. I'm somewhat following what we've done there.

dschaefer2: I'm going to spend more time on this. Since external builders currently only support library targets and executables, there may be a more natural SwiftBuild way to do this (e.g. a no sources targets with custom tasks producing the products).

dschaefer2: OK, I'm over thinking things a bit. I've pushed a change to run the custom tasks in parallel.

I may end up not using this for the external build, but I may experiment using this for custom targets...

Sources/SWBCore/ProjectModel/Target.swift
```diff
@@ -74,6 +76,7 @@ public final class CustomTask: ProjectModelItem, Sendable {
         self.outputFilePaths = try Self.parseValueForKeyAsArrayOfStrings(PIFKey_CustomTask_outputFilePaths, pifDict: pifDict).map { pifLoader.userNamespace.parseString($0) }
         self.enableSandboxing = try Self.parseValueForKeyAsBool(PIFKey_CustomTask_enableSandboxing, pifDict: pifDict)
         self.preparesForIndexing = try Self.parseValueForKeyAsBool(PIFKey_CustomTask_preparesForIndexing, pifDict: pifDict)
+        self.alwaysOutOfDate = try Self.parseValueForKeyAsBool(PIFKey_CustomTask_alwaysOutOfDate, pifDict: pifDict)
```
owenv, line 79: To support older clients, I believe this needs to parse an optional Bool so we can fill in a default value if this field isn't present

Sources/SWBProjectModel/PIFGenerationModel.swift
```diff
@@ -793,8 +793,9 @@ public enum PIF {
         public let outputFilePaths: [String]
         public let enableSandboxing: Bool
         public let preparesForIndexing: Bool
+        public let alwaysOutOfDate: Bool
 
-        public init(commandLine: [String], environment: [(String, String)], workingDirectory: String?, executionDescription: String, inputFilePaths: [String], outputFilePaths: [String], enableSandboxing: Bool, preparesForIndexing: Bool) {
+        public init(commandLine: [String], environment: [(String, String)], workingDirectory: String?, executionDescription: String, inputFilePaths: [String], outputFilePaths: [String], enableSandboxing: Bool, preparesForIndexing: Bool, alwaysOutOfDate: Bool = false) {
```
owenv: Since this is public ABI we should add a new init overload instead of an arg with default value

Sources/SWBProtocol/ProjectModel/CustomTask.swift
```diff
@@ -21,8 +21,9 @@ public struct CustomTask: SerializableCodable, Sendable {
     public let outputFilePaths: [MacroExpressionSource]
     public let enableSandboxing: Bool
     public let preparesForIndexing: Bool
+    public let alwaysOutOfDate: Bool
 
-    public init(commandLine: [MacroExpressionSource], environment: [(MacroExpressionSource, MacroExpressionSource)], workingDirectory: MacroExpressionSource, executionDescription: MacroExpressionSource, inputFilePaths: [MacroExpressionSource], outputFilePaths: [MacroExpressionSource], enableSandboxing: Bool, preparesForIndexing: Bool) {
+    public init(commandLine: [MacroExpressionSource], environment: [(MacroExpressionSource, MacroExpressionSource)], workingDirectory: MacroExpressionSource, executionDescription: MacroExpressionSource, inputFilePaths: [MacroExpressionSource], outputFilePaths: [MacroExpressionSource], enableSandboxing: Bool, preparesForIndexing: Bool, alwaysOutOfDate: Bool = false) {
```
owenv: Same here, we should add a new overload instead of default value

Sources/SWBTaskConstruction/ProductPlanning/ProductPlanner.swift
```diff
@@ -456,8 +456,21 @@ extension AggregateTarget
...
     {
+        // Shared start node so custom tasks run in parallel with the build phases.
+        let startNode = taskProducerContext.createVirtualNode("\(phaseNodeRoot(taskProducerContext.configuredTarget))-start")
+
         // TODO: We should probably check that only build phases useful in an aggregate target are present here.
-        return super.buildPhaseTargetTaskProducers(taskProducerContext).taskProducers
+        let (buildPhaseTaskProducers, _) = super.buildPhaseTargetTaskProducers(taskProducerContext, startPhaseNodes: [startNode])
+        var taskProducers = buildPhaseTaskProducers
+
+        // Custom tasks are ordered after the build phases.  The producer creates a phase end task which
```
owenv: I think this comment is no longer accurate

Sources/SwiftBuild/ProjectModel/CustomTask.swift
```diff
@@ -51,7 +52,8 @@ extension ProjectModel {
             inputFilePaths: [String],
             outputFilePaths: [String],
             enableSandboxing: Bool,
-            preparesForIndexing: Bool
+            preparesForIndexing: Bool,
+            alwaysOutOfDate: Bool = false
```
owenv: Same comment here about adding a new overload

owenv (commented):
lgtm aside from some minor comments

dschaefer2:
Thanks! I'll work those in. I'm still trying to get the tests passing on Windows.

Sources/SWBCore/ProjectModel/Target.swift
dschaefer2: parseValueforKeyAsBool already handles optionality and defaults to false. Feels like we should be fine?

Sources/SWBProjectModel/PIFGenerationModel.swift
dschaefer2: Done.

Sources/SWBTaskConstruction/ProductPlanning/ProductPlanner.swift
dschaefer2: Yeah, we fixed that didn't we :).

Sources/SWBProtocol/ProjectModel/CustomTask.swift
dschaefer2: Done

Sources/SwiftBuild/ProjectModel/CustomTask.swift
dschaefer2: Done.

dschaefer2:
@swift-ci please test

owenv (approved):

PR #1495, opened 2026-06-29
[6.4.x] Use Swift SDK manifest path as imposed SDKROOT during specialization

owenv:
This updates @MaxDesiatov's changes in https://github.com/swiftlang/swift-build/pull/1326 to tweak a few comments, and update the test so it can run on a non-macOS host OS. I've also tested it against @kateinoigakukun's updated reproducer in https://github.com/swiftlang/swift-package-manager/issues/10228.


owenv:
@swift-ci test

jakepetroules (approved):

PR #1497, opened 2026-06-30
Modernize simulator test run destinations to arm64

cyndyishida:
Flip the simulator run destinations to arm64 and update the affected test expectations accordingly.

Tests whose simulator arch derives from `ARCHS_STANDARD` (rather than the active arch) would otherwise break once x86_64 is clamped out of the simulators at newer deployment targets; pin an explicit deployment target for those.

Assisted by Claude Opus 4.8


rjmansfield (approved):

cyndyishida:
@swift-ci please test

cyndyishida:
@swift-ci please test

jakepetroules (approved):

PR #1499, opened 2026-07-01
Include WORKSPACE_DIR when configuring prefix mapping

owenv:
Swift packages use the shared WORKSPACE_DIR as a shared CWD for compiles to maximize module reuse. This path is a parent of PROJECT_DIR, so it need to be prefix mapped to ensure we can get cache hits when cloning a package + dependencies at two different paths. Identified while working on https://github.com/swiftlang/swift-package-manager/pull/10246


owenv:
@swift-ci test

cachemeifyoucan (approved):

rjmansfield (approved):

jakepetroules (approved):

owenv:
@swift-ci test

PR #1500, opened 2026-07-01
Honour SWIFT_DRIVER_SWIFT_FRONTEND_EXEC override in the in-process Swift driver

rjmansfield:


rjmansfield:
@swift-ci test

rjmansfield:
@swift-ci test

Sources/SWBCore/LibSwiftDriver/LibSwiftDriver.swift
```diff
@@ -563,11 +563,17 @@ public final class LibSwiftDriver {
         self.diagnosticsEngine = diagnosticsEngine
         var env = ProcessEnvironmentBlock()
         let compilerExecutableDir: TSCBasic.AbsolutePath?
+        // Allow SWIFT_DRIVER_SWIFT_FRONTEND_EXEC to override which swift-frontend the in-process driver execs.
```
owenv, line 566: SwiftCompiler.swift currently has:
```
 // Pass through SWIFT_DRIVER_SWIFT_FRONTEND_EXEC if set
        // This is consumed by the driver to select which swift-frontend binary to use.
        let frontendExecValue = cbc.scope.evaluate(cbc.scope.namespace.parseString("$(SWIFT_DRIVER_SWIFT_FRONTEND_EXEC)"))
        if !frontendExecValue.isEmpty {
            env.append(("SWIFT_DRIVER_SWIFT_FRONTEND_EXEC", frontendExecValue))
        }
```

Which should be putting the right value into the `environment` parameter to this function, which it looks like we're currently ignoring. Could we try to thread that value here instead? We usually try to avoid reading from the process-global environment since it can interfere with incremental builds

PR #1501, opened 2026-07-01
Add support for Amazon Linux 2023 in CI

jakepetroules:


jakepetroules:
@swift-ci test

.github/workflows/pull_request.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     needs: [soundness, space-format-check]
     with:
       enable_cross_pr_testing: true
-      linux_os_versions: '["amazonlinux2", "bookworm", "noble", "jammy", "rhel-ubi9"]'
+      linux_os_versions: '["amazonlinux2", "amazonlinux2023", "bookworm", "noble", "jammy", "rhel-ubi9"]'
```
bkhouri, line 21: suggestion:  According to https://forums.swift.org/t/dropping-support-for-amazon-linux-2/87920, we should also drop amazonlinux2 support.

.github/workflows/pull_request.yml
```diff
@@ -18,7 +18,7 @@ jobs:
...
       linux_pre_build_command: ./.github/scripts/prebuild.sh
       linux_build_command: 'swift test --no-parallel'
       linux_swift_versions: '["nightly-main", "nightly-6.3"]'
```
bkhouri, line 24: We should add support for AmazonLinux2023 for `nightly-main` and  `nightly-6.4.x`

```suggestion
      linux_swift_versions: '["nightly-main", "nightly-6.6.x"]'
```

.github/workflows/pull_request.yml
jakepetroules: I was going to do that in a separate PR but I guess I could roll it into this one...

.github/workflows/pull_request.yml
jakepetroules: You wrote 6.6 there :) But I'm handling that separately in #1502

PR #1502, opened 2026-07-01
Update minimum CI versions to 6.4

jakepetroules:
...and to Xcode 26.4 (Swift 6.3) on macOS. This drops support for Swift 6.2 on Apple platforms, and for 6.3 elsewhere.


jakepetroules:
@swift-ci test

Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:6.0
```
owenv, line 1: I think it's too early to bump this, some dependents still have an earlier tools version

jakepetroules: Packages with older tools versions can depend on packages with newer ones. The only thing that matters is the Swift version doing the build -- what's the oldest version in use that you're aware of?

jakepetroules: Actually, sourcekit-lsp has 6.3 on main, so I think we're good?

owenv: SwiftPM has some CI running 6.2

jakepetroules: OK, updated to 6.2

owenv (approved):

jakepetroules:
@swift-ci test

PR #1504, opened 2026-07-02
Introduce a new linux-musl platform and avoid linking relocatable objects when we target it

### LINKED ISSUE #10237: Swift Build 6.4: duplicate libc++abi symbols when statically linking two C++ targets
A package that links two vendored copies of BoringSSL (one from swift-nio-ssl, one from swift-crypto) fails to statically link on Swift 6.4 with duplicate C++ symbol errors. The same package links fine when using the legacy build system (`--build-system native`), which points at the new Swift Build system that became the default in 6.4. The bug is not specific to swift-crypto or swift-nio-ssl.

### The error

When statically linking, `ld.lld` reports duplicate symbols coming from both BoringSSL objects:

```
ld.lld: error: duplicate symbol: std::set_new_handler(void (*)())
>>>            CNIOBoringSSL.o:(std::set_new_handler(void (*)()))
>>>            CCryptoBoringSSL.o:(.text._ZSt15set_new_handlerPFvvE+0x0)
```

There are several such symbols (`set_terminate`, `__cxa_terminate_handler`, `__cxa_demangle`, `std::terminate()`, and more). These are C++ runtime symbols that end up in both objects. They are not BoringSSL's own API, which is correctly namespaced (`CNIOBoringSSL_*` vs `CCryptoBoringSSL_*`) and does not collide.

The default build system emits these symbols with external linkage into each object, so the two objects collide at link time. The legacy build system links the same objects without the collision.

### How to reproduce

A minimal package with the two dependencies is enough (reproduced with swift-nio-ssl 2.37.1 and swift-crypto 4.5.0):

```swift
// swift-tools-version:6.0
import PackageDescription
let package = Package(
    name: "DupSymRepro",
    dependencies: [
        .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.37.1"),
        .package(url: "https://github.com/apple/swift-crypto.git", from: "4.5.0"),
    ],
    targets: [.executableTarget(name: "DupSymRepro", dependencies: [
        .product(name: "NIOSSL", package: "swift-nio-ssl"),
        .product(name: "Crypto", package: "swift-crypto"),
    ])]
)
```

```swift
// Sources/DupSymRepro/main.swift
import NIOSSL
import Crypto
_ = TLSConfiguration.makeClientConfiguration()
_ = SHA256.hash(data: Array("hello".utf8))
```

Build statically with the static Linux SDK:

```bash
# Fails on Swift 6.4 (default build system)
swift build --swift-sdk <arch>-swift-linux-musl

# Succeeds on Swift 6.4
swift build --swift-sdk <arch>-swift-linux-musl --build-system native
```

### Results

| Toolchain | Build system | Outcome |
|---|---|---|
| 6.4 | default (Swift Build) | Fails with the duplicate symbol errors above |
| 6.4 | `--build-system native` | Succeeds (valid static binary) |
| 6.3.2 | default | Succeeds |

Swift 6.3.2 succeeds by default because the legacy build system is still the default there.

### Workaround

Build with `--build-system native`.

### Environment

Reproduced in `swiftlang/swift:nightly-6.4.x-jammy` targeting `aarch64-swift-linux-musl`. The same failure shows up on `x86_64` in CI, and on a macOS host cross-compiling to linux-musl. It first surfaced in CI for [awslabs/swift-aws-lambda-runtime#567](https://github.com/awslabs/swift-aws-lambda-runtime/pull/567).

owenv:
Resolves https://github.com/swiftlang/swift-package-manager/issues/10237

Use a distinct platform for targeting linux-musl triples, whose spec domain inherits from the existing linux platform. This allows us to override some of the product types, and should also make it easier to model linux -> linux cross compilation in some edge cases.

In the immediate term, this change is motivated by the fact that the linux-musl triple unconditionally configures clang to include -lc++abi on link lines, which is incompatible with linking relocatable objects. Instead, we build object libraries for SwiftPM targets like we're already doing on a few other platforms.

Targeting this to main for now, and I'll cherrypick to 6.4 later unless we come up with a better approach

### DIFF
--- a/Sources/SWBGenericUnixPlatform/Plugin.swift
+++ b/Sources/SWBGenericUnixPlatform/Plugin.swift
@@ -89,6 +89,7 @@ struct GenericUnixPlatformSpecsExtension: SpecificationsExtension {
     func specificationDomains() -> [String: [String]] {
         [
             "linux": ["generic-unix"],
+            "staticlinux": ["linux"],
             "freebsd": ["generic-unix"],
             "openbsd": ["generic-unix"],
         ]
@@ -97,7 +98,7 @@ struct GenericUnixPlatformSpecsExtension: SpecificationsExtension {
 
 struct GenericUnixPlatformInfoExtension: PlatformInfoExtension {
     func additionalPlatforms(context: any PlatformInfoExtensionAdditionalPlatformsContext) throws -> [(path: Path, data: [String: PropertyListItem])] {
-        return try OperatingSystem.createFallbackSystemToolchains.compactMap { operatingSystem in
+        var platforms: [(path: Path, data: [String: PropertyListItem])] = try OperatingSystem.createFallbackSystemToolchains.compactMap { operatingSystem in
             // Only create platforms if the host OS allows a fallback toolchain, or we're cross compiling.
             guard operatingSystem.createFallbackSystemToolchain || operatingSystem != context.hostOperatingSystem else {
                 return nil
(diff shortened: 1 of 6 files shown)


owenv:
@swift-ci test

jakepetroules:
Thoughts on "staticlinux" as the platform name instead?

> this change is motivated by the fact that the linux-musl triple unconditionally configures clang to include -lc++abi on link lines, which is incompatible with linking relocatable objects

This is more related to the staticness than the Musl-ness. In theory the Static Linux SDK could change to a different libc implementation in the future, like uClibc.

jakepetroules:
Also this would technically be the first platform with a "-" in the name and I'm slightly worried about edge cases in parsing :)

owenv:
@jakepetroules renamed it to `staticlinux`

owenv:
@swift-ci test

jakepetroules (approved):

owenv:
@swift-ci test

PR #1505, opened 2026-07-02
Drop x86_64 from simulator builds at deployment targets 27+

cyndyishida:
Mirrors the existing macOS x86_64 clamp for the iOS, tvOS, watchOS, and visionOS simulators. Adds a test covering the clamped behavior.

resolves: rdar://179972645

assisted by Claude Opus 4.8


owenv (approved):
LGTM but there may be further test failures that need to be fixed

cyndyishida:
@swift-ci please test

cyndyishida:
> LGTM but there may be further test failures that need to be fixed

👍 My hope is that https://github.com/swiftlang/swift-build/pull/1497 covered them

PR #1508, opened 2026-07-04
Add build service API for querying CAS size

owenv:
Add API for returning information about a compiler CAS instance given a set of config options. For the time being, it only returns the size in bytes. This will be adopted by SwiftPM in https://github.com/swiftlang/swift-package-manager/pull/10246.


owenv:
@swift-ci test

jakepetroules (approved):

PR #1510, opened 2026-07-05
Restrict fine grained compilation caching to supported object file formats

owenv:
Otherwise it gets enabled by default and the compiler errors out when passed a non-MachO triple


owenv:
@swift-ci test

rjmansfield (approved):

cachemeifyoucan (approved):

omochi:
@owenv @rconnell9 

Would it be possible to backport this fix to `release/6.4.x`?

Swift 6.4 development snapshots expose SwiftPM compilation caching, but
enabling Clang compilation caching on Linux currently causes swift-build
to pass the Mach-O-only `-fcas-backend` option.

I confirmed that the fix from this PR is present on `main`, but is not
present in `release/6.4.x` or
`swift-6.4.x-DEVELOPMENT-SNAPSHOT-2026-09-01-a`.

This currently prevents us from evaluating SwiftPM compilation caching
on Linux CI using the Swift 6.4 snapshots.

PR #1513, opened 2026-07-06
[6.4.x] Introduce a new linux-musl platform and avoid linking relocatable objects when we target it

owenv:
6.4.x. cherrypick of https://github.com/swiftlang/swift-build/pull/1504

Original description:

Resolves https://github.com/swiftlang/swift-package-manager/issues/10237

Use a distinct platform for targeting linux-musl triples, whose spec domain inherits from the existing linux platform. This allows us to override some of the product types, and should also make it easier to model linux -> linux cross compilation in some edge cases.

In the immediate term, this change is motivated by the fact that the linux-musl triple unconditionally configures clang to include -lc++abi on link lines, which is incompatible with linking relocatable objects. Instead, we build object libraries for SwiftPM targets like we're already doing on a few other platforms.

Targeting this to main for now, and I'll cherrypick to 6.4 later unless we come up with a better approach

Cherrypick this to the release branch now that a main-branch toolchain has been produced and I've verified it resolves the linking issues (including those seen in this repo's own CI jobs)

### DIFF
--- a/Sources/SWBGenericUnixPlatform/Plugin.swift
+++ b/Sources/SWBGenericUnixPlatform/Plugin.swift
@@ -89,6 +89,7 @@ struct GenericUnixPlatformSpecsExtension: SpecificationsExtension {
     func specificationDomains() -> [String: [String]] {
         [
             "linux": ["generic-unix"],
+            "staticlinux": ["linux"],
             "freebsd": ["generic-unix"],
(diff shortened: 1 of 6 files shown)


owenv:
@swift-ci test

rjmansfield (approved):

PR #1515, opened 2026-07-06
[6.4.x] Fix product type checks in package diamond resolution

owenv:
SwiftPM PIF generation now uses a product type which inherits from "com.apple.product-type.objfile" instead, but diamond resolution was checking for an exact match. Instead, this needs to be a conforms-to check to ensure we correctly use dynamic variant when needed. 

This was found during code review in https://github.com/swiftlang/swift-build/pull/1424. This backports a minimal version of the change to 6.4


daveinglis (approved):

owenv:
@swift-ci test

jakepetroules (approved):

owenv:
@swift-ci test windows

PR #1519, opened 2026-07-07
Iterate ARCHS_BASE for clang DocC symbol extraction

rjmansfield:
Cohort member archs build as a variant slice of their base arch's compile rather than as a separate task, so the clang symbol extractor and its DocC consumer should use ARCHS_BASE instead of ARCHS.

rdar://168173468


mhrawdon (approved):

rjmansfield:
@swift-ci test

PR #1520, opened 2026-07-08
pipe2/dup3 O_CLOEXEC

jakepetroules:
Add `safeOpen` / `safePipe` / `safeDuplicate` helpers on `FileDescriptor` that always request close-on-exec, so descriptors created by the build system are not leaked into the child processes it spawns. They route through the swift-system pipe2/dup3 wrappers where those are available:

- `safePipe` uses `FileDescriptor.pipe(options: .closeOnExec)` on non-Apple platforms (including Windows), and on Apple platforms running OS 27 or later via the SDK's `System` framework; older Apple OSes fall back to a plain `pipe()`.
- `safeDuplicate(as:)` uses `FileDescriptor.duplicate(as:options:)` on non-Apple platforms (dup3 is unavailable on Windows) and on Apple OS 27 or later; elsewhere it falls back to a plain `dup2()`.
- `safeDuplicate()` (no target) uses `fcntl(F_DUPFD_CLOEXEC)` to atomically obtain the lowest-numbered unused descriptor with close-on-exec set; on Windows, which has no `F_DUPFD_CLOEXEC`, it falls back to a plain `dup()`.
- `safeOpen` forces `OpenOptions.closeOnExec`.

Adopt these wrappers in `FSProxy`, `PbxCp`, `XCBuildDataArchive`, `LineReader`, and the build service entry point, and require swift-system 1.7.3 (which exposes `OpenOptions.closeOnExec` on Windows).

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -502,7 +502,7 @@ if useLocalDependencies {
 
     package.dependencies += [
         .package(url: "https://github.com/swiftlang/swift-driver.git", branch: relatedDependenciesBranch),
-        .package(url: "https://github.com/apple/swift-system.git", .upToNextMajor(from: "1.5.0")),
+        .package(url: "https://github.com/apple/swift-system.git", .upToNextMajor(from: "1.7.3")),
         .package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.0.3"),
         .package(url: "https://github.com/swiftlang/swift-tools-protocols.git", branch: relatedDependenciesBranch),
     ]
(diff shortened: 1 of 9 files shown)


Package.swift
```diff
@@ -502,7 +502,7 @@ if useLocalDependencies {
 
     package.dependencies += [
         .package(url: "https://github.com/swiftlang/swift-driver.git", branch: relatedDependenciesBranch),
-        .package(url: "https://github.com/apple/swift-system.git", .upToNextMajor(from: "1.5.0")),
+        .package(url: "https://github.com/apple/swift-system.git", .upToNextMajor(from: "1.7.3")),
```
owenv, line 505: The toolchain is still pinned to 1.5.0: https://github.com/swiftlang/swift/blob/a1d0266274502f1c1fadd1ee5652088795be0d96/utils/update_checkout/update-checkout-config.json#L168-L169

PR #1523, opened 2026-07-08
Macro-expand product reference names in target and linkage dependency resolvers

daveinglis:


daveinglis:
@swift-ci test

Sources/SWBCore/DependencyResolution.swift
```diff
@@ -14,6 +14,21 @@ public import SWBUtil
 import struct SWBProtocol.RunDestinationInfo
 import SWBMacro
 
+extension ProductReference {
+    /// The product reference's name, evaluated as a build setting expression only when it contains a
```
owenv: I'd maybe drop the duplicated comment here

Sources/SWBCore/DependencyResolution.swift
```diff
@@ -14,6 +14,21 @@ public import SWBUtil
...
+    /// macro reference. Most product references are constant basenames (e.g. `Foo.framework`), so
+    /// `settings` is computed lazily to avoid constructing settings in the common case.
+    ///
+    /// - Note: `settings` is an `@autoclosure`; the settings are only constructed when the name actually
+    ///   contains a macro reference. Do not "simplify" the call sites to pass an eagerly-computed value.
+    ///
+    /// See <rdar://problem/29410050>.
+    func evaluatedName(settings: @autoclosure () -> Settings) -> String {
```
owenv: I worry that using an autoclosure here t skip evaluating settings if they're not needed is a little too fragile. Could we just make this a regular computeSettings closure and make callers add the braces?

daveinglis: done

Sources/SWBCore/DependencyResolution.swift
daveinglis: updated

daveinglis:
@swift-ci test

Sources/SWBCore/DependencyResolution.swift
```diff
@@ -14,6 +14,16 @@ public import SWBUtil
 import struct SWBProtocol.RunDestinationInfo
 import SWBMacro
 
+extension ProductReference {
+    /// Like `evaluatedName(scope:)`, but only invokes `computeSettings` when the product reference name
+    /// actually contains a macro reference, so callers can pass a potentially expensive settings
+    /// computation without paying for it in the common constant-basename case. See <rdar://problem/29410050>.
+    func evaluatedName(computeSettings: () -> Settings) -> String {
+        guard name.contains("$") else { return name }
```
owenv, line 22: The call below also has the $ check so I think this one is redundant

owenv (approved):

Sources/SWBCore/DependencyResolution.swift
daveinglis: this is the early out before calling to expensive `computeSettings` closure

owenv: ah, right. This looks fine

PR #1526, opened 2026-07-09
Update build server logging to align with builtin SourceKitLSP implementation

owenv:
This structures the logging so it more closely matches the old implementation, which makes it a bit easier to test their behavior against each other.


owenv:
@swift-ci test

rjmansfield (approved):

PR #1527, opened 2026-07-09
Fix race between BSP initialization and build description generation

### LINKED ISSUE #1509: `SWBBuildServer` returns `-32001 "No build description"` for requests in the post-handshake window before async build description generation completes
### Description

`SWBBuildServer.initialize()` kicks off build description generation as fire-and-forget work, then immediately returns the `InitializeBuildResponse`. Any BSP request that requires `buildDescriptionID` and arrives *after* the handshake completes but *before* generation finishes fails with `ResponseError(-32001, "No build description")` instead of being queued or awaited.

### Steps to reproduce

We hit this with [`sourcekit-xcode-bsp`](https://github.com/tideline-studio/sourcekit-xcode-bsp) without pre-warminng the build description. The current workaround requests a build description before starting the actual build server.

1. Bring up the bridge so sourcekit-lsp completes the BSP handshake with `SWBBuildServer` (`build/initialize` -> `build/initialized`).
2. As soon as the handshake completes, sourcekit-lsp issues `workspace/buildTargets` (it does not call `workspace/waitForBuildSystemUpdates` first on this path).
3. Because `initialize()` only *scheduled* build description generation and `buildDescriptionID` is still `nil`, the request fails with `ResponseError(-32001, "No build description")`.
4. No recovery was possible after the initial failure

The race is timing-dependent on how long generation takes, so it reproduces more reliably on larger workspaces (longer generation window).

### Expected behavior

A request arriving in the natural post-handshake window should not fail.

### Notes

The generation duration depends on workspace size, which widens the window for larger projects.

somiljain2006:
Fixes #1509

Fix race when handling requests after BSP initialization

- Ensure requests that depend on the build description wait for pending build description generation instead of failing with "No build description".
- Add a regression test covering workspace/buildTargets immediately after the BSP handshake without an explicit workspace/waitForBuildSystemUpdates request.
- Add test coverage for C compiler arguments to verify clang arguments do not include Swift-specific flags.

### DIFF
--- a/Sources/SwiftBuild/SWBBuildServer.swift
+++ b/Sources/SwiftBuild/SWBBuildServer.swift
@@ -266,11 +266,9 @@ public actor SWBBuildServer: QueueBasedMessageHandler {
     }
 
     private func buildTargets(request: WorkspaceBuildTargetsRequest) async throws -> WorkspaceBuildTargetsResponse {
-        try await logTaskToClient(name: "Computing targets list") { _ in
-            guard let buildDescriptionID else {
-                throw ResponseError.unknown("No build description")
-            }
-            let targets = try await session.configuredTargets(
+            try await logTaskToClient(name: "Computing targets list") { _ in
+                let buildDescriptionID = try await waitForBuildDescriptionID()
+                let targets = try await session.configuredTargets(
                 buildDescription: buildDescriptionID,
                 buildRequest: buildRequest
             ).asyncMap { targetInfo in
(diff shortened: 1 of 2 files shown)


owenv (changes requested):
This is working as intended, requests fail fast when no build description is computed so that the client can choose to provide fallback options before reloading when it receives a notification that the list of targets has changed.

somiljain2006:
Thanks for the clarification. I had interpreted this as a server-side readiness race.

PR #1528, opened 2026-07-09
Reject stale cached build descriptions when the workspace has changed

daveinglis:
## Summary

`SWBBuildService` traps (`EXC_BREAKPOINT` in `Workspace.project(for:)`) when a `ConfiguredTarget` frozen into a `BuildDescription` is resolved against a workspace that no longer contains it. Observed via CrashTracer in build teardown (rdar://181149017), multiple-producers diagnostics (rdar://145534302), and an indexing header-info request (a `+1` occurrence on a newer build, macOS 26.6).

## Root cause

A build/index request can carry `buildRequest.buildDescriptionID`. When set, the description is fetched via the `.cachedOnly` path, which looks it up purely by that ID and **skips validation** — `isValidFor` short-circuits to `return true` "for index-query performance."

Because the cache is keyed only by the ID, a description built against an earlier workspace generation can be served against the current one: when a client reuses an ID after the session swapped to a reloaded workspace (e.g. an external `git pull` rewrote project files, or a PIF re-transfer), the old description's targets are resolved against the new workspace. If a target was **removed**, that resolution traps.

The normal `.newOrCached` path is unaffected — its signature re-keys on `workspace.signature`, so a workspace change forces a rebuild.

## Fix

Validate at the cache layer, once, covering all consumers (builds, teardown, and every indexing message type that shares `handleIndexingInfoRequest`):

- For a `.cachedOnly` request, `isValidFor` now accepts the description **only if the current workspace still contains all of its targets** (checked by identity or GUID via `Workspace.contains(target:)`), instead of accepting unconditionally.
- A rejected `.cachedOnly` request surfaces a clean thrown error (the construction path cannot rebuild without a plan); both callers already handle this — the build aborts, the index query returns a failure.

Reusing a cached description across a workspace change is a **supported contract** — index clients rely on it (see `ArenaIndexingInfoTests.useBuildDescriptionAfterPIFChange`, which reuses an index build description after a PIF change that modifies a still-present target). Target-existence preserves that reuse and rejects only the genuinely-unsafe case (a referenced target was removed).

This supersedes the earlier per-site `workspace.contains(target:)` guards (removed here), which only covered known call sites and missed the indexing path.

## Testing

`BuildDescriptionTests.cachedOnlyValidatesTargetsAgainstWorkspace`:
- a *different* workspace generation that still contains the target (same GUID) → the cached description is **reused** (not rejected);
- a workspace generation where the target was **removed** → the request is **rejected** rather than returning the stale description.

## Note on behavior

Target-existence rejects the whole cached description if **any** referenced target is absent from the current workspace, so an index query can trigger a re-prepare in that case. This is deliberate: a removed target means the description is genuinely stale, and re-preparing is correct. Reuse across all other workspace edits (targets still present) is unaffected.

### DIFF
--- a/Sources/SWBBuildSystem/BuildOperation.swift
+++ b/Sources/SWBBuildSystem/BuildOperation.swift
@@ -972,6 +972,14 @@ package final class BuildOperation: BuildSystemOperation {
     package func taskDiscoveredRequiredTargetDependency(target: ConfiguredTarget, antecedent: ConfiguredTarget, reason: RequiredTargetDependencyReason, warningLevel: BooleanWarningLevel) {
         if !transitiveDependencyExists(target: target, antecedent: antecedent) {
 
+            // The targets come from the (possibly cached) build description and may be stale relative to a workspace
+            // that was reloaded from changed on-disk project files mid-build. Resolving settings for a target the
+            // reloaded workspace no longer contains would trap in `Workspace.project(for:)`, so skip the diagnostic in
+            // that case rather than crashing — it's moot once the target is gone. rdar://181149017
+            guard workspace.contains(target: target.target), workspace.contains(target: antecedent.target) else {
+                return
+            }
+
             // Ensure we only diagnose missing dependencies when platform and SDK variant match. We perform this check as late as possible since computing settings can be expensive.
             let targetSettings = requestContext.getCachedSettings(target.parameters, target: target.target)
             let antecedentSettings = requestContext.getCachedSettings(antecedent.parameters, target: antecedent.target)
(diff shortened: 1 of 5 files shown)


daveinglis:
@swift-ci test

Sources/SWBBuildSystem/BuildOperation.swift
```diff
@@ -972,6 +972,14 @@ package final class BuildOperation: BuildSystemOperation {
     package func taskDiscoveredRequiredTargetDependency(target: ConfiguredTarget, antecedent: ConfiguredTarget, reason: RequiredTargetDependencyReason, warningLevel: BooleanWarningLevel) {
         if !transitiveDependencyExists(target: target, antecedent: antecedent) {
 
+            // The targets come from the (possibly cached) build description and may be stale relative to a workspace
```
owenv: The settings computation below should go through the BuildRequestContext, which should have a fixed WorkspaceContext/Workspace throughout the whole build. Is that changing somehow, or is something accessing the WorkspaceContext via the Session object when it should be using the BuildRequestContext?

daveinglis: Hmm, ok, going to dig further to try and figure out what is happening here...

daveinglis: You were right — `BuildRequestContext` holds a fixed, immutable `WorkspaceContext`, and these sites use it correctly rather than reaching through the Session.

Digging further, the mismatch comes from the `.cachedOnly(buildDescriptionID)` path: it looks a description up purely by ID and skips `isValidFor` entirely ("for index-query performance"), so a description built against an earlier workspace generation gets served against the current one when a client reuses a stale ID after a workspace reload. The `buildDescriptionID` already encodes `workspace.signature`, so the ID is fine — it's the reuse pairing that's unvalidated. The normal `.newOrCached` path is unaffected because its signature re-keys on the workspace.

I've reworked the PR to fix it there: `BuildDescription` now records its originating `workspaceSignature`, and `isValidFor` rejects a `.cachedOnly` description whose signature doesn't match the request's workspace (rejection surfaces as a clean thrown error, which the build/index callers already handle). The per-site guards are dropped, since this covers all consumers including the indexing path — which is where a `+1` crash on a newer build (macOS 26.6) actually landed.

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test

owenv (approved):

PR #1529, opened 2026-07-09
Failures in AppShortcutsStringsValidationTests.appShortcutsStringsValidation()

quinntaylor:
Switch from comparing the entire string to regex for the required token, to work both before/after the diagnostic string was changed.

Resolves rdar://177031942.


quinntaylor:
@swift-ci test

owenv (approved):

quinntaylor:
@swift-ci test windows

PR #1530, opened 2026-07-09
Adopt -experimental-lazy-typecheck and -experimental-skip-non-exportable-decls when preparing for indexing

owenv:
SourceKit-LSP's builtin SwiftPM prepare-for-indexing operation has been using these options for a while, introduce an on-by-default setting to match that behavior and roll them out more widely.


owenv:
@swift-ci test

rjmansfield (approved):

PR #1531, opened 2026-07-09
Remove stale .swiftmodule bundles across build descriptions

### LINKED ISSUE #1171: Stale .swiftmodule files in Products cause cross-build-description compilation failures
When the dependency graph changes between builds (different PIF), swbuild creates a new build description but writes to the same `Build/Products/` directory. Stale `.swiftmodule` files from prior builds accumulate and can cause compilation failures when the Swift compiler loads them from the module search path (`-I`) and chases their embedded transitive dependency chains into modules that no longer exist for the platform.

## Error

```
error: Unable to find module dependency: 'LeafLib'
note: a dependency of Swift module 'MiddleLib': '.../Products/Debug-watchsimulator/MiddleLib.swiftmodule/arm64-apple-watchos-simulator.swiftmodule'
note: a dependency of main module 'AppLib'
```

## Root cause

The `Build/Products/Debug-<platform>/` directory is shared across all build descriptions. When the PIF changes (target added/removed, dependency graph restructured), a new xcbuilddata directory is created with a new build description. But nobody cleans up `.swiftmodule` outputs from prior build descriptions.

The existing stale file removal (SFR) infrastructure (`StaleFileRemovalContext.swift` + llbuild `stale-file-removal` tool) operates within a single build description — it tracks outputs for the current build's targets only. When the PIF changes, the new build description has no visibility into artifacts left by the old one. SFR literally cannot see cross-build-description stale files because they belong to a different xcbuilddata directory.

## Mechanism

1. **Build 1** (PIF A): `AppLib → MiddleLib → LeafLib`, all build for watchOS. Products contains `AppLib.swiftmodule`, `MiddleLib.swiftmodule`, `LeafLib.swiftmodule`.

2. **PIF changes**: `MiddleLib` and `LeafLib` are removed from the package. Only `AppLib` remains. `LeafLib.swiftmodule` is removed (e.g., by SFR from a prior matched build description, or it was never built for this platform).

3. **Build 2** (PIF B): New build description. `AppLib` compiles for watchOS. The compiler's `-I` search path includes `Products/Debug-watchsimulator/` where stale `MiddleLib.swiftmodule` still exists.

4. `AppLib` source has `#if canImport(MiddleLib)`. The compiler finds stale `MiddleLib.swiftmodule` in the search path → `canImport` returns `true` → compiler loads `MiddleLib.swiftmodule` → its metadata says it depends on `LeafLib` → `LeafLib.swiftmodule` doesn't exist → **error**.

On a **clean build** (no stale modules), `#if canImport(MiddleLib)` correctly evaluates to `false` and the build succeeds.

## Reproduction

Repro files: https://gist.github.com/jverkoey/6115f88ed45f1a0bac73252120d42526

```bash
# Download and run the reproduction case
mkdir stale-swiftmodule-repro && cd stale-swiftmodule-repro
curl -sL https://gist.githubusercontent.com/jverkoey/6115f88ed45f1a0bac73252120d42526/raw/reproduce.sh -o reproduce.sh
curl -sL https://gist.githubusercontent.com/jverkoey/6115f88ed45f1a0bac73252120d42526/raw/build-params.json -o build-params.json
mkdir -p StaleModuleRepro/Sources/{AppLib,MiddleLib,LeafLib}
curl -sL https://gist.githubusercontent.com/jverkoey/6115f88ed45f1a0bac73252120d42526/raw/Package.swift -o StaleModuleRepro/Package.swift
curl -sL https://gist.githubusercontent.com/jverkoey/6115f88ed45f1a0bac73252120d42526/raw/App.swift -o StaleModuleRepro/Sources/AppLib/App.swift
curl -sL https://gist.githubusercontent.com/jverkoey/6115f88ed45f1a0bac73252120d42526/raw/Middle.swift -o StaleModuleRepro/Sources/MiddleLib/Middle.swift
curl -sL https://gist.githubusercontent.com/jverkoey/6115f88ed45f1a0bac73252120d42526/raw/Leaf.swift -o StaleModuleRepro/Sources/LeafLib/Leaf.swift
bash reproduce.sh
```

The script:
1. Clones and builds swift-build (one-time, ~5 minutes)
2. Creates a Swift package with `AppLib → MiddleLib → LeafLib`
3. Builds for watchOS with swbuild (creates `.swiftmodule` files)
4. Deletes `LeafLib.swiftmodule` (simulating cross-build-description cleanup inconsistency)
5. Modifies `Package.swift` to remove `MiddleLib` and `LeafLib` targets
6. Rebuilds with swbuild using same derived data → **fails**
7. Removes stale `MiddleLib.swiftmodule` and rebuilds → **succeeds** (control)

Requires macOS with Xcode and the watchOS simulator SDK.

## Real-world impact

We hit this on CI where multiple builds with different PIFs share derived data across slots. The failure is intermittent — it depends on which slot a build lands on and what stale state exists from prior builds. Clean builds always succeed; the failure requires stale derived data from a prior build with a different dependency graph.

Typical real-world error chain (3 levels deep):
```
error: Unable to find module dependency: 'DecayingValue'
  note: a dependency of Swift module 'JourneySample'
  note: a dependency of Swift module 'TripLogger'
  note: a dependency of main module 'SettingsPage'
```

## Suggested fix

After `ConstructBuildDescription` completes, swbuild has perfect knowledge of which `.swiftmodule` files will be produced for each platform. A cleanup step before llbuild starts could:

1. Collect all `.swiftmodule` output paths from the build description's task store
2. Group by platform directory
3. Enumerate actual `.swiftmodule` directories in each `Products/Debug-<platform>/`
4. Remove any `.swiftmodule` directory not in the expected set

This extends the existing SFR concept to operate across build descriptions, specifically for `.swiftmodule` files in the shared Products directory.

somiljain2006:
Fixes #1171 

Clean up stale .swiftmodule bundles between build descriptions by recording the set of expected module output directories during build description construction and removing obsolete module bundles before starting a new build. This prevents stale cross-build artifacts from remaining in shared product directories when the target graph changes, while treating cleanup failures as non-fatal. Add a regression test covering consecutive builds with different dependency graphs to verify that obsolete .swiftmodule bundles are removed correctly.

### DIFF
--- a/Sources/SWBBuildSystem/BuildOperation.swift
+++ b/Sources/SWBBuildSystem/BuildOperation.swift
@@ -501,6 +501,39 @@ package final class BuildOperation: BuildSystemOperation {
         let serializationQueue = buildDatabaseSerializationQueue(for: databaseDirectory)
         do {
             return try await serializationQueue.withOperation { [buildEnvironment] in
+
+                if request.enableStaleFileRemoval && request.buildCommand.shouldEnableStaleFileRemoval {
+                    // Group the modules expected by this build description by their platform directory.
+                    var expectedByPlatformDir: [Path: Set<Path>] = [:]
+                    for moduleDir in buildDescription.swiftModuleOutputDirectories {
+                        let platformDir = moduleDir.dirname
+                        expectedByPlatformDir[platformDir, default: []].insert(moduleDir)
+                    }
+
+                    // Remove any .swiftmodule bundles that exist on disk but are no longer
+                    // produced by the current build description.
+                    for (platformDir, expectedModules) in expectedByPlatformDir {
+                        // Skip directories that haven't been created yet 
+                        guard fs.exists(platformDir) else { continue }
+
+                        do {
+                            let contents = try fs.listdir(platformDir)
+                            for item in contents {
+                                guard item.hasSuffix(".swiftmodule") else { continue }
+
+                                let foundPath = platformDir.join(item)
+
+                                if !expectedModules.contains(foundPath) {
+                                    try fs.remove(foundPath)
+                                    buildOutputDelegate.note("Removed stale cross-build swiftmodule: \(foundPath.str)")
+                                }
+                            }
+                        } catch {
+                            buildOutputDelegate.note("Skipped stale swiftmodule cleanup in \(platformDir.str): \(error)")
+                        }
+                    }
+                }
+
                 // If we use a cached build system, be sure to release it on build completion.
                 if userPreferences.enableBuildSystemCaching {
                     // Get the build system to use, keyed by the directory containing the (sole) database.
--- a/Sources/SWBTaskExecution/BuildDescription.swift
+++ b/Sources/SWBTaskExecution/BuildDescription.swift
@@ -143,6 +143,9 @@ package final class BuildDescription: Serializable, Sendable, Encodable, Cacheab
     /// The set of all (non-virtual) paths produced by tasks in the build description.
     private let allOutputPaths: Set<Path>
 
+    /// The set of planned top-level `.swiftmodule` directory paths.
+    package let swiftModuleOutputDirectories: Set<Path>
+
     private let rootPathsPerTarget: [ConfiguredTarget: [Path]]
     private let moduleCachePathsPerTarget: [ConfiguredTarget: [Path]]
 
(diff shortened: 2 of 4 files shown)


owenv (changes requested):
Thanks for the PR. Deleting only the swiftmodules here doesn't properly account for bundle layouts and is likely to cause further problems when a product also exposes clang modules, is a framework, etc. We'd also likely want to reuse the scoping mechanism introduced in https://github.com/swiftlang/swift-build/pull/1183/changes rather than injecting this into the BuildOperation as a special case. There are also existing workflows which depend on the products being persistent, which would require a migration path to adopt the stricter model. I think this needs more design work before we'd be able to accept an implementation

PR #1534, opened 2026-07-10
Preserve Android deployment target in Swift module filenames

### LINKED ISSUE #1167: SPM Build for Android fails with --build-system swiftbuild: module not found due to target triple mismatch
When building a Swift package for Android using `--build-system swiftbuild`, the build fails because the Android API level is stripped from the target triple during module lookup.

The compiler searches for a module targeting `aarch64-unknown-linux-android` but only finds one compiled for `aarch64-unknown-linux-android28.0.0`. The default build system handles this correctly.

### Reproduction

**Works (default build system):**
```
swiftly run swift build --product LoggingExtensionDynamic --swift-sdk aarch64-unknown-linux-android28 +main-snapshot-2026-03-02
```

**Fails (swiftbuild):**
```
swiftly run swift build --product LoggingExtensionDynamic --swift-sdk aarch64-unknown-linux-android28 --build-system swiftbuild +main-snapshot-2026-03-02
```

### Error

```
could not find module 'Logging' for target 'aarch64-unknown-linux-android';
found: aarch64-unknown-linux-android28.0.0
```

Full verbose output: [ErrorOutput.txt](https://github.com/user-attachments/files/25798342/ErrorOutput.txt)

### Sample package

Minimal reproduction with 1 target and 1 file: [FoundationExtension.zip](https://github.com/user-attachments/files/25798403/FoundationExtension.zip)

### Environment

Reproduced on both configurations:

| | Toolchain | Android SDK | API level |
|---|---|---|---|
| 1 | Swift 6.2.3 (swift-6.2.3-RELEASE), arm64-apple-macosx15.0 | [finagolfin/swift-android-sdk](https://github.com/finagolfin/swift-android-sdk) | 24 |
| 2 | Swift 6.3-dev (LLVM c483661d4d6aac5, Swift 34b17dd3e44b2a6), arm64-apple-macosx26.0, +assertions | Official swift.org nightly (main-snapshot-2026-03-02) | 28 |

somiljain2006:
Fixes #1167 

Preserve deployment target in Android module filenames. Android module filenames must retain the deployment target because it is part of the target triple used for module lookup. Other platforms continue to use the unversioned triple. Adds a regression test covering Android, Apple, and Linux triples.

### DIFF
--- a/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
@@ -879,9 +879,9 @@ public final class SwiftCompilerSpec : CompilerSpec, SpecIdentifierType, SwiftDi
         func destinationModuleFileName(_ triple: LLVMTriple) -> String {
             switch self {
             case .generateModule(let moduleTriple, _):
-                return moduleTriple.unversioned.description
+                return moduleTriple.moduleFileNameDescription
             default:
-                return triple.unversioned.description
+                return triple.moduleFileNameDescription
             }
         }
 
@@ -3743,6 +3743,15 @@ extension SwiftDiscoveredCommandLineToolSpecInfo {
 
         // Get the info from the global cache.
         return try await discoveredSwiftCompilerInfo(producer, delegate, at: toolPath, blocklistsPathOverride: userSpecifiedBlocklists)
+    }
+}
+
+extension LLVMTriple {
+    @_spi(Testing) public var moduleFileNameDescription: String {
+        if system == "linux" && environment?.hasPrefix("android") == true {
+            return description
+        }
+        return unversioned.description
     }
 }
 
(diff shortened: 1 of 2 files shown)


jakepetroules:
Why hasn't this surfaced as an issue so far?

somiljain2006:
My understanding is that this is a fairly specific code path. The existing behavior is correct on Apple platforms, and the issue only manifests on Android, where the API level is encoded in the target triple used for module lookup. That likely limits the number of users affected.

jakepetroules:
#1167 was filed some time ago. Are you actually still seeing this issue with nightly?

This change doesn't seem correct, unless someone is actively making changes recently to change the behavior. @finagolfin are you aware of anything that has changed the swiftmodule naming scheme on Android lately?

finagolfin:
My guess is the linked issue was back when Android support wasn't done yet in this new build system back in March, but ever [since Jake and Owen added that support in the next month](https://github.com/swiftlang/swift-build/commits/main/Sources/SWBAndroidPlatform), I have not seen a single issue related to Android versioning, and we now have several repos building with versioned target triples for Android.

I suggest that these users try the latest 6.4 snapshot and report any issues they see with some detailed build output. I did just make a change in trunk today, swiftlang/swift#90650, but that shouldn't affect anyone and these reports long predate that anyway.

finagolfin:
Closing since the linked issue was closed, please reopen if you see any problems that this would help with.

PR #1537, opened 2026-07-13
Allow extending explicit modulemap contents with the Swift generated header

owenv:
This is needed to support SwiftPM targets which mix Swift/C sources


owenv:
@swift-ci test

jakepetroules (approved):

PR #1541, opened 2026-07-14
Port ModuleVerifierModuleMap to use Swift's Regex

sudanchapagain:

The `ModuleVerifierModuleMap` used to use old `NSRegularExpression` API from `Foundation`. With this PR, it now uses the Swift's Regex type from Swift's stdlib. The foundation module is imported because `modulesDeclared` function in the file requires it.

The more important change is that the regex itself has been changed. Rather than relying on a lazy match with a negative lookbehind to find the closing quote, it now explicitly matches the contents of a quoted string by treating ordinary characters and escape sequences separately.

I ran the tests for it with `swift test --filter ModuleVerifierModuleMapTests` and all passed. But running all the tests with `swift test` failed in settings module. I am worried if i have done something wrong while setting up swift-build.

This was part of the swift mentorship program, where owen assigned this task for me to get familiar with the contribution process.

### DIFF
--- a/Sources/SWBCore/ClangModuleVerifier/ModuleVerifierModuleMap.swift
+++ b/Sources/SWBCore/ClangModuleVerifier/ModuleVerifierModuleMap.swift
@@ -12,9 +12,7 @@
 
 public import SWBUtil
 
-// FIXME: Port to Regex
-import struct Foundation.NSRange
-import class Foundation.NSRegularExpression
+import Foundation
 
 @_spi(Testing)
 public enum ModuleMapKind: String {
@@ -47,14 +45,11 @@ public struct ModuleVerifierModuleMap: Hashable {
         excludedHeaderNames = findHeaders(withSpecifierPattern: "exclude")
         privateHeaderNames = findHeaders(withSpecifierPattern: "private(?:\\s+textual)?")
 
-        // This is what you would call "quick and dirty". Real dirty. :(
         func findHeaders(withSpecifierPattern specifierPattern: String) -> [String] {
-            let pattern = specifierPattern + "\\s+header\\s+\"(.*?)(?<!\\\\)\""
-            let regularExpression = try! NSRegularExpression(pattern: pattern, options: .dotMatchesLineSeparators)
-            // The range conversions are all real awkward, <rdar://78800247&78800081&78800042>.
-            let wholeString = NSRange(moduleContents.startIndex..., in: moduleContents)
-            return regularExpression.matches(in: moduleContents, range: wholeString).map {match in
-                return String(moduleContents[Range(match.range(at: 1), in: moduleContents)!])
+            let pattern = specifierPattern + #"\s+header\s+"((?:[^"\\]|\\.)*)""#
+            let regex = try! Regex(pattern).dotMatchesNewlines()
+            return moduleContents.matches(of: regex).compactMap { match in
+                match.output[1].substring.map(String.init)
             }
         }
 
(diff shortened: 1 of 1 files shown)


sudanchapagain:
@swift-ci test

Sources/SWBCore/ClangModuleVerifier/ModuleVerifierModuleMap.swift
```diff
@@ -47,14 +45,11 @@ public struct ModuleVerifierModuleMap: Hashable {
...
-            let pattern = specifierPattern + "\\s+header\\s+\"(.*?)(?<!\\\\)\""
-            let regularExpression = try! NSRegularExpression(pattern: pattern, options: .dotMatchesLineSeparators)
-            // The range conversions are all real awkward, <rdar://78800247&78800081&78800042>.
-            let wholeString = NSRange(moduleContents.startIndex..., in: moduleContents)
-            return regularExpression.matches(in: moduleContents, range: wholeString).map {match in
-                return String(moduleContents[Range(match.range(at: 1), in: moduleContents)!])
+            let pattern = specifierPattern + #"\s+header\s+"((?:[^"\\]|\\.)*)""#
+            let regex = try! Regex(pattern).dotMatchesNewlines()
+            return moduleContents.matches(of: regex).compactMap { match in
+                match.output[1].substring.map(String.init)
```
jakepetroules: Can you use a named capture group instead of a subscript?

owenv:
Thanks!

> But running all the tests with swift test failed in settings module. I am worried if i have done something wrong while setting up swift-build.

These could be an issue with the setup, or just using a different version of the compiler/Xcode/tools compared to CI, so that aren't necessarily concerning.

owenv:
@swift-ci test

Sources/SWBCore/ClangModuleVerifier/ModuleVerifierModuleMap.swift
```diff
@@ -47,14 +45,11 @@ public struct ModuleVerifierModuleMap: Hashable {
...
 
-        // This is what you would call "quick and dirty". Real dirty. :(
         func findHeaders(withSpecifierPattern specifierPattern: String) -> [String] {
-            let pattern = specifierPattern + "\\s+header\\s+\"(.*?)(?<!\\\\)\""
-            let regularExpression = try! NSRegularExpression(pattern: pattern, options: .dotMatchesLineSeparators)
-            // The range conversions are all real awkward, <rdar://78800247&78800081&78800042>.
-            let wholeString = NSRange(moduleContents.startIndex..., in: moduleContents)
-            return regularExpression.matches(in: moduleContents, range: wholeString).map {match in
-                return String(moduleContents[Range(match.range(at: 1), in: moduleContents)!])
+            let pattern = specifierPattern + #"\s+header\s+"((?:[^"\\]|\\.)*)""#
```
owenv: Instead of appending specifierPattern and the regex string here, maybe we should just have one regex each for the whole `excludedHeaderNames` and `privateHeaderNames` patterns. That would allow writing them as regex literals and avoid the try! below

sudanchapagain:
> These could be an issue with the setup, or just using a different version of the compiler/Xcode/tools compared to CI, so that aren't necessarily concerning.

My operating system is Linux (Fedora Linux), with the following compiler spec: `Swift version 6.3.2 (swift-6.3.2-RELEASE)` `Target: x86_64-unknown-linux-gnu`. Could these be a problem?

owenv:
> My operating system is Linux (Fedora Linux), with the following compiler spec: Swift version 6.3.2 (swift-6.3.2-RELEASE) Target: x86_64-unknown-linux-gnu. Could these be a problem?

This should be generally fine, but a few tests here and there might require nightly snapshots of 6.4/main. So far the CI test results here look ok

owenv (approved):

owenv:
@swift-ci test

owenv:
@swift-ci test windows

PR #1543, opened 2026-07-14
Use top-level framework dir for ExtractAPI

snprajwal:
The documentation build passes each dependency's module map to Clang via `-fmodule-map-file=`. Clang only treats a module map as belonging to the framework if the `Modules` dir is present directly inside the `.frameworks` dir. However, it is possible that a framework nests the modules inside a versioned subdirectory. This patch updates the ExtractAPI task to normalise the path and use the top-level symlink (`Foo.framework/Modules/module.modulemap`) instead of the versioned path (`Foo.framework/Versions/A/Modules/module.modulemap`).

rdar://176811346

### DIFF
--- a/Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/TAPISymbolExtractorTaskProducer.swift
+++ b/Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/TAPISymbolExtractorTaskProducer.swift
@@ -71,15 +71,36 @@ final class TAPISymbolExtractorTaskProducer: PhasedTaskProducer, TaskProducer {
         }
 
         // For Objective-C targets that `@import` their dependencies we need to know the locations of the module maps for those targets.
-        // This search needs to be done recursively since, for example with SwiftPM the target will depend on the PACKAGE:PRODUCT target
+        // These are registered explicitly with `-fmodule-map-file=` because a module map is not always discoverable, e.g.:
+        //
+        // - Swift package dependencies generate their module map into build intermediates
+        // - A target may point `MODULEMAP_PATH` at a custom location
+        //
+        // This search needs to be done recursively since, for example, with SwiftPM, the target will depend on the PACKAGE:PRODUCT target
         // which in turn depends on the PACKAGE:TARGET target that generates the module map.
+        // rdar://93921653
         var dependenciesModuleMaps = OrderedSet<Path>()
         if let configuredTarget = context.configuredTarget {
             let transitiveClosure = transitiveClosure([configuredTarget], successors: { context.globalProductPlan.dependencies(of: $0) }).result
             for dependencyTarget in transitiveClosure {
                 let settings = context.globalProductPlan.getTargetSettings(dependencyTarget)
                 if let moduleInfo = context.globalProductPlan.getModuleInfo(dependencyTarget) {
-                    dependenciesModuleMaps.append(moduleInfo.moduleMapPaths.builtPath)
+                    var moduleMapPath = moduleInfo.moduleMapPaths.builtPath
+                    // A framework generates its module map at `TARGET_BUILD_DIR/MODULES_FOLDER_PATH`,
+                    // which is nested under `Foo.framework/Versions/A/Modules` for versioned frameworks.
+                    // Clang only treats a module map as belonging to a framework when the `Modules`
+                    // directory sits directly inside `.framework`, so point at the top-level
+                    // `Foo.framework/Modules` symlink instead if `MODULEMAP_PATH` is not provided.
+                    // rdar://176811346
+                    let depScope = settings.globalScope
+                    if settings.productType is FrameworkProductTypeSpec,
+                       depScope.evaluate(BuiltinMacros.MODULEMAP_PATH).isEmpty {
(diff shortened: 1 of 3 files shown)


Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/TAPISymbolExtractorTaskProducer.swift
```diff
@@ -82,7 +82,24 @@ final class TAPISymbolExtractorTaskProducer: PhasedTaskProducer, TaskProducer {
             for dependencyTarget in transitiveClosure {
                 let settings = context.globalProductPlan.getTargetSettings(dependencyTarget)
                 if let moduleInfo = context.globalProductPlan.getModuleInfo(dependencyTarget) {
-                    dependenciesModuleMaps.append(moduleInfo.moduleMapPaths.builtPath)
+                    // Clang only treats a module map as belonging to the framework if the `Modules` dir
```
owenv: Why is this manually registering the module map in the first place vs finding it in framework search paths? It's not clear to me if any of this code should be traversing the target graph

Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/TAPISymbolExtractorTaskProducer.swift
```diff
@@ -82,7 +82,24 @@ final class TAPISymbolExtractorTaskProducer: PhasedTaskProducer, TaskProducer {
...
+                    // is present directly inside the `.frameworks` dir. However, it is possible that
+                    // a framework nests the modules inside a versioned subdirectory. In such cases,
+                    // normalise the path to use the top-level symlink (`Foo.framework/Modules/module.modulemap`)
+                    // instead of the versioned path (`Foo.framework/Versions/A/Modules/module.modulemap`).
+                    // This mirrors the normalisation used for the Swift VFS overlay in `ModuleMapTaskProducer`.
+                    let builtPath = moduleInfo.moduleMapPaths.builtPath
+                    var moduleMapPath = builtPath
+                    var dir = builtPath.dirname
+                    while !dir.isRoot {
```
owenv: This directory traversal will be very expensive in builds with large numbers of targets

Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/TAPISymbolExtractorTaskProducer.swift
snprajwal: We can use `-F` to find it for frameworks, but there are cases where we would have to manually pass the module map, e.g. when an ObjC target depends on Swift packages, the generated module map file is present in the build intermediates, and is not discoverable by Clang (rdar://93921653). I'll add a comment explaining why we're doing this so it's clear going forward.

Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/TAPISymbolExtractorTaskProducer.swift
snprajwal: It's only string slicing here, it does not actually check the filesystem. However, I realised we can instead construct this path from existing info instead of walking up the dir tree, I've removed this.

Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/TAPISymbolExtractorTaskProducer.swift
owenv: When loading the module of package dependencies, I think it would be better to compute the compiler options used to compile this target's module, and then filter those to what's relevant to symbol graph extraction. This is a lot cheaper than traversing the dependency graph, and it ensures that the symbol extractor is able to load any module the compiler is able to

snprajwal: That's a better approach indeed, I'm on board. However, it would be a much bigger refactor than this fix. Would it be acceptable to land this fix now so it resolves the existing issue where Xcode is unable to build documentation in certain cases, and I can do the refactor in a follow-up PR?

snprajwal: P.S. I'm not familiar with the dev workflow of this project, typically I've merged into main and cherry-picked into `release/6.4.x` for other Swift repos. I see some past PRs that might point to the reverse happening here, so please let me know if I need to target the release branch in this PR instead.

owenv (approved):
I don't love this approach, but it seems narrow enough to be low-risk

Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/TAPISymbolExtractorTaskProducer.swift
owenv: There is an auto merger from 6.4 to main, but cherrypicks are ok too

snprajwal:
Thanks! I'll re-target this PR at `release.6.4.x` so the automerger can pull it into main.

snprajwal:
@swift-ci please test

Issue #1535, opened 2026-07-10
Feature Request: Display the resolved target dependency graph after package diamond resolution


yimajo:
## Motivation

Currently, the target dependency graph emitted by Swift Build only shows dependencies as they exist before the package diamond problem is resolved.

This feature request proposes updating the graph to show the targets and dependencies that are actually used after package diamond resolution. I’d appreciate feedback on whether this would be a useful addition to Swift Build.

When a Swift Package Target is replaced with a dynamic variant, the current dependency graph still shows the original Target.

For example, consider the following dependency graph:

```
App
├── DynamicLinkPackageProduct (dynamic)
│   └── PackageLib
│        └── Common
│
└── PackageLibProduct2 (static)
    └── PackageLib2
         └── Common
```

The build log currently shows the following:

```
note: Target dependency graph (6 targets)

    Target 'App' in project 'SampleProject'

        ➜ Explicit dependency on target 'DynamicLinkPackageProduct' in project 'Modules'

        ➜ Explicit dependency on target 'PackageLibProduct2' in project 'Modules'

    Target 'PackageLibProduct2' in project 'Modules'

        ➜ Explicit dependency on target 'PackageLib2' in project 'Modules'

        ➜ Explicit dependency on target 'Common' in project 'Modules'

    Target 'PackageLib2' in project 'Modules'

        ➜ Explicit dependency on target 'Common' in project 'Modules'

    Target 'DynamicLinkPackageProduct' in project 'Modules'

        ➜ Explicit dependency on target 'PackageLib' in project 'Modules'

        ➜ Explicit dependency on target 'Common' in project 'Modules'

    Target 'PackageLib' in project 'Modules'

        ➜ Explicit dependency on target 'Common' in project 'Modules'

    Target 'Common' in project 'Modules' (no dependencies)
```

However, when `Common` is used through multiple dependency paths, resolving the package diamond problem may cause the build to use `Common-dynamic`.
The current dependency graph does not show that `Common-dynamic` is being used.

The linker log emitted later in the build may show the Targets that are dynamically linked.
This may confirm that `Common-dynamic` was dynamically linked.

However, the linker log alone does not show:

- that `Common-dynamic` is a dynamic variant of `Common`
- which Targets in the dependency graph were replaced with dynamic variants
- how dependencies were reconnected to the replacement Targets
- which Targets remained static instead of being replaced with dynamic variants

[...]

PR #1549, opened 2026-07-19
Display dynamic targets in the target dependency graph

yimajo:
Implements #1535.

## Motivation

The target dependency graph currently emitted by Swift Build in the build log and `target-graph.txt` does not indicate whether each target will be built dynamically.

In particular, the existing output does not identify targets that are replaced with dynamic variants as a result of resolving the diamond dependency problem.

This change appends `(Dynamic)` to targets that will be built dynamically.
This applies both to targets that are explicitly configured as dynamic and to targets that become dynamic as a result of resolving the diamond dependency problem.

This makes it possible to verify at a relatively early stage of the build, and at low cost, whether each target will be built with the intended linkage.

### Change where the dependency graph diagnostic is computed

The existing implementation computes the dependency graph diagnostic before resolving the diamond dependency problem.
At that point, Swift Build cannot determine which targets will be replaced with dynamic variants.

This change makes the following updates:

- Remove the existing log emission from `PlanningOperation`
- Remove the dependency graph diagnostic computation from `TargetDependencyResolver`
- Compute `dependencyGraphDiagnostic` in `GlobalProductPlan`
  - `GlobalProductPlan` holds the result of resolving the diamond dependency problem, allowing it to determine which targets will be built dynamically

### Share the diagnostic between log and file output

- Add `dependencyGraphDiagnostic` to `BuildDescription`
  - This carries the diagnostic computed by `GlobalProductPlan` to the existing file output point
- Serialize `dependencyGraphDiagnostic` together with `BuildDescription`
  - This allows the same diagnostic to be used when restoring a `BuildDescription` from the on-disk cache

### Log output

- Add a `ReportTargetDependencyGraph` activity and associate the dependency graph diagnostic with it
  - A separate activity is used because using the existing `CreateBuildDescription` activity would require substantial changes to where that activity begins and ends
- Emit the diagnostic only for `.newOrCached` requests
  - This preserves the behavior of the existing log output

### File output

- Preserve the existing output timing for `target-graph.txt`
- Change the diagnostic used by `BuildDescriptionManager.serializeBuildDescription` from the `TargetBuildGraph` diagnostic to `BuildDescription.dependencyGraphDiagnostic`

### Tests

- Add a unit test for dependency graph log output with a single target
- Add a unit test for the Dynamic annotation when an existing dynamic framework is present and another target becomes dynamic as a result of resolving the diamond dependency problem
- Add a unit test for dependency graph log output with non-explicit dependencies
- Make minor adjustments to the existing `BuildDescription` tests

### DIFF
--- a/Sources/SWBTaskConstruction/ProductPlanning/ProductPlan.swift
+++ b/Sources/SWBTaskConstruction/ProductPlanning/ProductPlan.swift
@@ -144,6 +144,74 @@ package final class GlobalProductPlan: GlobalTargetInfoProvider
         .sorted(by: \.target.guid)
     }
 
+    /// Computes and returns a hierarchical diagnostic representing the build graph.
+    package var dependencyGraphDiagnostic: Diagnostic {
+        .init(
+            behavior: .note,
+            location: .unknown,
+            data: .init(
+                "Target dependency graph (\(planRequest.buildGraph.allTargets.count) target" + (
+                    planRequest.buildGraph.allTargets.count > 1 ? "s" : ""
+                ) + ")"
+            ),
+            childDiagnostics: planRequest.buildGraph.allTargets.reversed().map {
+                let project = planRequest.workspaceContext.workspace.project(
+                    for: $0.target
+                )
+                let resolvedDependencies = planRequest.buildGraph.resolvedDependencies(
+                    of: $0
+                )
+                let isDynamic = getTargetSettings($0).globalScope.evaluate(BuiltinMacros.MACH_O_TYPE) == "mh_dylib"
+                    || dynamicallyBuildingTargetsWithDiamondLinkage[$0.target] != nil
+
+                return .init(
+                    behavior: .note,
+                    location: .unknown,
+                    data: .init(
+                        [
+                            "Target '\($0.target.name)' in project '\(project.name)'",
+                            planRequest.buildRequest.shouldSkipExecution(
+                                target: $0.target
+                            ) ? " (skipped due to 'Skip Dependencies' scheme option)" : "",
+                            {
+                                switch (resolvedDependencies.isEmpty, isDynamic) {
+                                case (true, true): " (no dependencies, Dynamic)"
+                                case (true, false): " (no dependencies)"
+                                case (false, true): " (Dynamic)"
+                                case (false, false): ""
+                                }
+                            }(),
+                        ]
+                        .joined()
+                    ),
+                    childDiagnostics: resolvedDependencies.map {
+                        let project = planRequest.workspaceContext.workspace
+                            .project(for: $0.target.target)
+
+                        return .init(
+                            behavior: .note,
+                            location: .unknown,
+                            data: .init(
+                                "➜ " + {
+                                    switch $0.reason {
+                                    case .explicit:
+                                        "Explicit dependency on \($1)"
+                                    case .implicitBuildPhaseLinkage(filename: let filename, buildableItem: _, buildPhase: let buildPhase):
+                                        "Implicit dependency on \($1) via file '\(filename)' in build phase '\(buildPhase)'"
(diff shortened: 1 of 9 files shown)


yimajo:
@swift-ci test

Sources/SWBTaskConstruction/ProductPlanning/ProductPlan.swift
```diff
@@ -144,6 +144,74 @@ package final class GlobalProductPlan: GlobalTargetInfoProvider
...
+                ) + ")"
+            ),
+            childDiagnostics: planRequest.buildGraph.allTargets.reversed().map {
+                let project = planRequest.workspaceContext.workspace.project(
+                    for: $0.target
+                )
+                let resolvedDependencies = planRequest.buildGraph.resolvedDependencies(
+                    of: $0
+                )
+                let isDynamic = getTargetSettings($0).globalScope.evaluate(BuiltinMacros.MACH_O_TYPE) == "mh_dylib"
```
owenv, line 164: This will print the (Dynamic) suffix of any target that builds an mh_dylib, even it it's a target that could only ever build an mh_dylib, like a framework in an Xcode project. I wonder if this should only include the dynamicallyBuildingTargetsWithDiamondLinkage so that the text only appears if a target could build with either static or dynamic linkage. What do you think?

owenv:
@swift-ci test

Sources/SWBTaskConstruction/ProductPlanning/ProductPlan.swift
yimajo: Thank you for the comment.

Initially, I thought it was important to know only the result of resolving the diamond problem.
However, if this replaces the existing target dependency graph log, I have come to think that (Dynamic) should represent the final state—that the target will be dynamically linked in this build—rather than the reason that a dynamic variant was selected as a result of resolving the diamond problem. I believe this would make the log more consistent.

This is because, after considering how to simplify the output in response to [the comment on the issue](https://github.com/swiftlang/swift-build/issues/1535#issuecomment-4936466047), [I came to think that it would be unnatural for a target explicitly configured by the developer for dynamic linking not to be shown as (Dynamic)](https://github.com/swiftlang/swift-build/issues/1535#issuecomment-4956200532).

To elaborate on the reasoning, if (Dynamic) were added only to targets made dynamic as a result of resolving the diamond problem, targets explicitly configured from the outset to be built with dynamic linkage would not be marked as (Dynamic). I am concerned that the absence of the (Dynamic) annotation could cause confusion.

However, if we emitted a separate log and file in addition to the existing output, I think developers would not be confused if that output explicitly described the result of diamond resolution using dynamicallyBuildingTargetsWithDiamondLinkage. On the other hand, this would mean emitting two dependency graphs, which also seems redundant. Am I overthinking this? 

I also think it is useful for targets that are built as mh_dylib by default, such as frameworks in Xcode projects, to be marked as (Dynamic). Developers who use frameworks may not necessarily be aware of the linkage selected by default.  Even if a developer wants to [build a target as a static framework](https://developer.apple.com/documentation/xcode/creating-a-static-framework), the developer can confirm that (Dynamic) disappears from the log after making the change.

For this discussion, I think it may help to group the cases into roughly three categories:

- The result of resolving the diamond problem
  - A dynamic variant is selected by Swift Build
- `.library(type: .dynamic, ...)` in `Package.swift`
  - The developer explicitly selects dynamic linkage
- A framework target in an Xcode project
  - Dynamic linkage is used by default

I think displaying (Dynamic) whenever a target is dynamically linked, regardless of which of the three categories it falls into, would make the representation simpler and more consistent.

owenv: ok, that reasoning makes sense to me

owenv (approved):

PR #1557, opened 2026-07-22
Fix timeout configuration in GitHub workflows

owenv:
Fix an incorrect key which caused this to fail to parse


owenv:
@swift-ci test

jakepetroules (approved):

PR #1558, opened 2026-07-22
Run build-phase storyboardc bundles through the postprocessor (rdar://50701007)

jakepetroules:
## Summary

Fixes rdar://50701007: storyboard postprocessing was completely ignored by the build system.

The storyboard **linker** and **postprocessor** both synthesized build rules on the same `wrapper.storyboardc` input type, producing an unresolvable match tie. The build-rule engine hard-coded that pair (`DisambiguatingBuildRuleSet`) to suppress the ambiguity diagnostic and always pick the linker, so the postprocessor rule never produced a task.

## Approach

Rather than threading file provenance into the matcher, this gives compiled storyboard bundles a refined file type and lets the **existing** exact-vs-conformance match priority do the disambiguation:

- New file type `wrapper.storyboardc.compiled` (`BasedOn` `wrapper.storyboardc`, **no extension** — so on-disk `.storyboardc` bundles still infer the base type and the refined type is only ever stamped programmatically).
- The compiler and postprocessor stamp their `.storyboardc` outputs with the refined type, via a new overridable `declaredOutputFileType(forOutputAt:_:)` hook on `CommandLineToolSpec` (default returns `nil` → infer, so every other tool is unchanged).
- The postprocessor now writes to `$(TempResourcesDir)` instead of `$(ProductResourcesDir)` so its stripped output feeds the linker, and is backed by a small `IBStoryboardPostprocessorSpec` class.
- The linker accepts `wrapper.storyboardc.compiled`.
- The hard-coded workaround in `BuildRuleSet.swift` is removed.

Routing now falls out of priority matching with no collisions:

- on-disk `.storyboardc` added to a build phase (base type, exact match) → **postprocess** → temp → **link**
- `.storyboard` source → **compile** → temp → **link**
- compiler/postprocessor outputs (`.compiled`, exact match) → **link**

## Testing

- New `interfaceBuilderStoryboardPostprocessing` test: a pre-compiled `.storyboardc` added to a build phase is now stripped into the temp resources dir and then linked into the product (previously it was linked directly, with no postprocessing).
- All existing storyboard/localization/watch/installloc task-construction tests pass unchanged (the source→compile→link path is unaffected).
- Full `SWBTaskConstructionTests` suite passes except one pre-existing, unrelated Android build-directory-naming failure.

Also adds a `.storyboardc` → `wrapper.storyboardc` mapping to the test workspace helper so such bundles can be referenced in tests.

### DIFF
--- a/Sources/SWBApplePlatform/CMakeLists.txt
+++ b/Sources/SWBApplePlatform/CMakeLists.txt
@@ -34,6 +34,7 @@ add_library(SWBApplePlatform
   ResMergerLinkerSpec.swift
   SceneKitToolSpec.swift
   StoryboardLinker.swift
+  StoryboardPostprocessor.swift
   XCStringsCompiler.swift
   ActoolInputFileGroupingStrategy.swift
   AssetCatalogCompilerOutputParser.swift
--- a/Sources/SWBApplePlatform/InterfaceBuilderCompiler.swift
+++ b/Sources/SWBApplePlatform/InterfaceBuilderCompiler.swift
@@ -10,7 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-import SWBUtil
+public import SWBUtil
 public import SWBCore
 public import SWBMacro
 import Synchronization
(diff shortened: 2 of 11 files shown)


jakepetroules:
@swift-ci test

owenv (approved):

jakepetroules:
@swift-ci test

PR #1560, opened 2026-07-22
[6.4]Always use clang when linking a preview shim

owenv:
The preview shim will never contain code compiled with swiftc. Use clang to link it so we never force load the Swift compatibility libraries, which would fail to resolve symbols in the standard library.

rdar://181404607


owenv:
@swift-ci test

owenv:
* Explanation: Prefer clang when linking the preview shim, even when the linker driver would otherwise resolve to swiftc. The preview shim never contains code compiled with swiftc, so we want to avoid force loading compatibility libs when targeting an older OS.
* Scope: Previews-enabled builds of Playgrounds App projects
* Issues: rdar://181404607
* Risk:
Low, this is a focused change covered by existing tests
* Testing: Updated existing tests

rconnell9 (approved):

jakepetroules (approved):

PR #1561, opened 2026-07-23
Fix unused withLock result warnings

yimajo:
## Summary

Explicitly discard the results of `Set.insert` calls inside `withLock` closures.

The closures implicitly returned the results of `Set.insert`, triggering the following warning:

```
result of call to 'withLock' is unused
```

## Changes

- Discard the `Set.insert` result in `IbtoolCompilerSpec.BuildPhaseInfo.addToContext(_:)`
- Discard the `Set.insert` result in `TaskProducerContext.access(path:)`

This change does not affect behavior.


owenv (approved):
thanks

owenv:
@swift-ci test

jakepetroules (approved):

PR #1562, opened 2026-07-23
Build LinkageDependencyResolver product-name maps lazily

daveinglis:
### Summary

`LinkageDependencyResolver.init` eagerly builds `targetsByProductName` / `targetsByProductNameStem` over **every** workspace target, calling `evaluatedName` on each. Since every SwiftPM module target's product-reference name is a build setting expression (`$(EXECUTABLE_NAME)` / `$(WRAPPER_NAME)`), that forces `getCachedSettings` — and thus full `Settings` construction — for every target on **every** build, including no-change builds.

Those maps are only consulted during **implicit**-dependency resolution (`implicitDependency(forProductName:)` / `forProductNameStem:`). Packages that use only **explicit** dependencies — i.e. all SwiftPM packages — never trigger that path, so building the maps eagerly is pure overhead.

This PR builds the maps **lazily on first use**. They are byte-identical when built, so behavior is unchanged; explicit-dependency projects simply never pay to build them.

### Motivation

`rdar://182737643` — large SwiftPM packages regressed ~13% on no-change ("null") builds. The delta is entirely in the out-of-process `Build operation` phase. Root-causing with a compiler-fixed A/B (swap only `SWBBuildService` via `XCBBUILDSERVICE_PATH`, keep the toolchain constant) attributed the whole regression to the eager map build.

### DIFF
--- a/Sources/SWBCore/LinkageDependencyResolver.swift
+++ b/Sources/SWBCore/LinkageDependencyResolver.swift
@@ -73,35 +73,49 @@ fileprivate extension LinkageDependencyResolver {
 }
 
 actor LinkageDependencyResolver {
-    /// Sets of targets mapped by product name.
-    private let targetsByProductName: [String: Set<StandardTarget>]
-
-    /// Sets of targets mapped by product name stem.
-    private let targetsByProductNameStem: [String: Set<StandardTarget>]
+    /// Lazily-built target lookup maps, keyed by (evaluated) product name and by product-name stem.
+    ///
+    /// Building these forces `getCachedSettings` — and thus `Settings` construction — for every workspace
+    /// target whose product-reference name is a build setting expression, which for SwiftPM packages is
+    /// *every* module target (`$(EXECUTABLE_NAME)`/`$(WRAPPER_NAME)`). The maps are only consulted during
+    /// implicit-dependency resolution (`implicitDependency(forProductName:)` / `forProductNameStem:`), so
+    /// projects that use only explicit dependencies (all SwiftPM packages) never need them. Build them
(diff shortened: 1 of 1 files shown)


daveinglis:
@swift-ci test

Sources/SWBCore/LinkageDependencyResolver.swift
```diff
@@ -73,35 +73,49 @@ fileprivate extension LinkageDependencyResolver {
...
+    /// *every* module target (`$(EXECUTABLE_NAME)`/`$(WRAPPER_NAME)`). The maps are only consulted during
+    /// implicit-dependency resolution (`implicitDependency(forProductName:)` / `forProductNameStem:`), so
+    /// projects that use only explicit dependencies (all SwiftPM packages) never need them. Build them
+    /// lazily on first use to avoid that per-target `Settings` construction on every build — it was the
+    /// dominant cost of a no-change build for large packages. See <rdar://182737643>.
+    private struct ProductNameMaps: Sendable {
+        let byProductName: [String: Set<StandardTarget>]
+        let byProductNameStem: [String: Set<StandardTarget>]
+    }
+    private let productNameMapsCache = SWBMutex<ProductNameMaps?>(nil)
```
jakepetroules: Use a `Lazy<ProductNameMaps>` instead of a `SWBMutex<ProductNameMaps?>`, it handles the "already cached" ceremony for you.

daveinglis: found LazyCache that i think is a bit better

owenv (approved):

daveinglis:
@swift-ci test

PR #1545, opened 2026-07-15
Variant dylibs should have unsuffixed install names

ndrewh:
Variant dylibs (i.e. built using BUILD_VARIANTS) should have the same install name as the original. This (among other things) prevents dyld from ever loading both the variant and the original dylib.

rdar://79156379

PR #1563, opened 2026-07-23
Add EXECUTABLE_INSTALL_NAME to BuildSettings

ndrewh:
This is split off of #1545, which will provide a default for this setting and will switch to the new setting as the default for LD_DYLIB_INSTALL_NAME.

This change needs to land first, followed by adoption of an override in the swift-pm PIF-builder, and only then can #1545 be merged.


owenv (approved):

jakepetroules (approved):

ndrewh:
@swift-ci test

PR #1567, opened 2026-07-24
Fix SWBBuildService when Swift incremental cleanup diagnostics emitted after build request completes

rconnell9:
Fix SWBBuildService when Swift incremental cleanup diagnostics emitted after build request completes

SWBBuildService could crash with a precondition failure in Request.send at the end of a build. The crash originated on the SwiftModuleDependencyGraph queue when emitting a swift-driver incremental-build diagnostic after the Request had already been completed.

### DIFF
--- a/Sources/SWBBuildSystem/BuildOperation.swift
+++ b/Sources/SWBBuildSystem/BuildOperation.swift
@@ -1601,7 +1601,7 @@ internal final class OperationSystemAdaptor: SWBLLBuild.BuildSystemDelegate, Act
 
     /// Wait for all status activity to be complete before returning.
     func waitForCompletion(buildSucceeded: Bool) async {
-        let completionToken = await dynamicOperationContext.waitForCompletion()
+        _ = await dynamicOperationContext.waitForCompletion()
         if await validateCompilationCachePostBuild() == .succeeded {
             cleanupCompilationCache()
         }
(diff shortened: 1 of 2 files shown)


rconnell9:
@swift-ci test

Sources/SWBBuildSystem/BuildOperation.swift
```diff
@@ -1625,7 +1625,14 @@ internal final class OperationSystemAdaptor: SWBLLBuild.BuildSystemDelegate, Act
                     }
                 }
             }
+        }
 
+        // cleanUpForAllKeys above tears down Swift driver state asynchronously and emits
+        // incremental-build diagnostics.  We have to drain those diagnostics while the build's request
+        // is still open.
+        let completionToken = await dynamicOperationContext.waitForCompletion()
```
owenv: Instead of calling this twice, I wonder if we should make SwiftModuleDependencyGraph.cleanUpForAllKeys run the cleanup synchronously instead of async, since we want to ensure the incremental records are written before the next build anyway.

rconnell9: That seems reasonable.  I have a reproducer test case I can try your suggestion in, it is timing dependent and a bit long running

rconnell9:
@swift-ci test

owenv (approved):

PR #1568, opened 2026-07-24
[CAS] Serialize remote service path into .cas-config and unify plugin options

cachemeifyoucan:
Centralize CAS plugin-option computation on `CASOptions` so the clang and
Swift caching paths, the `BuildCacheInfo` message handler, and the
`.cas-config` file all share a single source of truth:

- Add `CASOptions.pluginOptions(useRemote:)` and the static
  `getRemoteServicePluginOption(remoteServicePath:)` helper, replacing the
  duplicated remote-service-path logic in ClangModuleDependencyGraph,
  LibSwiftDriver, and Messages.
- Write `PluginOptions` (including the remote service path) into
  `.cas-config`, matching the format LLVM's CASConfiguration reader expects,
  so external tools (dsymutil, lldb) can reach the same remote cache.
- Emit a warning from `CASOptions.create` when a remote service path is set
  but the CAS plugin is not enabled.

Add task-construction tests covering the serialized plugin options and the
new warning.


owenv (approved):

cachemeifyoucan:
@swift-ci please test

PR #1569, opened 2026-07-26
fix(build): ignore compiler verbosity in task signatures

### LINKED ISSUE #9299: [Parity] `--very-verbose` causes a rebuild with SwiftBuild
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

Running `swift build --very-verbose --build-system swiftbuild` should not cause a rebuild. However, it does.

### Expected behavior

a rebuild does not occur

### Actual behavior

providing `--very-verbose` causes a rebuild to occur.

### Steps to reproduce

1. Clone the SwiftPM repository
2. Run the following
```
P=$(swift run swift-build --build-system swiftbuild --package-path Fixtures/ValidLayouts/SingleModule/ExecutableNew --show-bin-path) 

swift run swift-package update
swift run swift-package clean --package-path Fixtures/ValidLayouts/SingleModule/ExecutableNew
ls -lrt ${P}/ExecutableNew  # this should return "No such file or directory

swift run swift-build --build-system swiftbuild --package-path Fixtures/ValidLayouts/SingleModule/ExecutableNew
ls -lrt ${P}/ExecutableNew  # the file should exists. Note the time

sleep 120
swift run swift-build --build-system swiftbuild --package-path Fixtures/ValidLayouts/SingleModule/ExecutableNew --very-verbose
ls -lrt ${P}/ExecutableNew  # the file should exists. The previous command only added `--very-verbose`.  The executable should have the same timestamp, but it doesn't.
```

### Swift Package Manager version/commit hash

`dc2e33fb6b2b1ab58d9170976dc5ce97afc8722b`

### Swift & OS version (output of `swift --version ; uname -a`)

Swift version `main-snapshot-2025-10-16`

```
❯ swift --version                                                                                                          
Apple Swift version 6.3-dev (LLVM 3a427639f2a06cf, Swift 5417f00a58578de)
Target: arm64-apple-macosx26.0
Build config: +assertions

❯ uname -a
Darwin BKs-Work-MBP-M4-Pro.local 25.1.0 Darwin Kernel Version 25.1.0: Tue Oct 14 21:57:14 PDT 2025; root:xnu_development-12377.41.5~4/DEVELOPMENT_ARM64_T6041 arm64
```

agisilaos:
Fixes swiftlang/swift-package-manager#9299

## Summary

Prevent compiler verbosity flags from invalidating Clang and Swift compiler task signatures.

SwiftPM forwards `-v` to the compiler when `--very-verbose` logging is enabled. Although this flag only changes diagnostic output, SwiftBuild included it when computing compiler task signatures. Adding `--very-verbose` to an otherwise unchanged build therefore caused compiler tasks and their downstream dependents to run again.

SwiftBuild already treats linker verbosity as output-agnostic. This change extends that behavior to Clang and Swift compiler invocations.

## Changes

- Treat Clang’s `-v` argument as output-agnostic when computing compiler task signatures.
- Treat Swift’s `-v` argument as output-agnostic when computing compiler task signatures.
- Add persistent-build regression coverage for `OTHER_CFLAGS=-v` and `OTHER_SWIFT_FLAGS=-v`.
- Verify that changing only compiler verbosity produces a null build.

## Alternatives considered

Removing `-v` from SwiftPM’s very-verbose build settings was rejected because it would prevent genuinely executed compiler tasks from producing the requested verbose output.

Filtering `-v` globally for every command-line tool was also considered. The change remains compiler-specific because `-v` is not necessarily output-agnostic for every tool, and the existing compiler specifications already provide the appropriate signature-normalization mechanism.

Handling only the linker flags is insufficient because Clang and Swift compiler task signatures still change and cause their downstream linker tasks to run.

## Notes

The flag remains present on the actual compiler command line. It is removed only from the command-line representation used to compute the task signature, so verbose compiler output continues to work whenever compilation is otherwise required.

After this change lands, SwiftPM will need to update its pinned SwiftBuild revision before removing the known-issue quarantine for the corresponding regression test.

## Testing

- `swift test --filter outputAgnosticsFlagsRebuild` — passed.
- `bash .github/scripts/format-check.sh` — passed.
- Verified that the regression test fails without the compiler normalization, reporting signature changes for both the Clang and Swift compilation tasks.

### DIFF
--- a/Sources/SWBCore/SpecImplementations/Tools/CCompiler.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/CCompiler.swift
@@ -823,6 +823,7 @@ public class ClangCompilerSpec : CompilerSpec, SpecIdentifierType, GCCCompatible
         "-fdiagnostics-parseable-fixits",
         "-fno-elide-type",
         "-fdiagnostics-show-template-tree",
+        "-v",
 
         // https://clang.llvm.org/docs/ClangCommandLineReference.html
         "-fdiagnostics-show-note-include-stack",
--- a/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
@@ -722,6 +722,10 @@ public final class SwiftCompilerSpec : CompilerSpec, SpecIdentifierType, SwiftDi
         case invalid
     }
 
+    static let outputAgnosticCompilerArguments = Set<ByteString>([
+        "-v"
+    ])
+
     static let outputAgnosticCompilerArgumentsWithValues = Set<ByteString>([
         "-index-store-path",
         "-index-unit-output-path",
@@ -732,7 +736,8 @@ public final class SwiftCompilerSpec : CompilerSpec, SpecIdentifierType, SwiftDi
     ])
 
     static func isOutputAgnosticCommandLineArgument(_ argument: ByteString, prevArgument: ByteString?) -> Bool {
-        if SwiftCompilerSpec.outputAgnosticCompilerArgumentsWithValues.contains(argument) {
+        if SwiftCompilerSpec.outputAgnosticCompilerArguments.contains(argument) ||
+           SwiftCompilerSpec.outputAgnosticCompilerArgumentsWithValues.contains(argument) {
             return true
         }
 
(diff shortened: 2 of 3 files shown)


owenv (approved):

owenv:
@swift-ci test

Issue #1574, opened 2026-07-28
Swift SDKs won't compile for non-matching target arch with swiftbuild system in 6.4


xtremekforever:
If I do this on a Linux x86_64 host:

```bash
$ swiftly install --use 6.4.x-snapshot
$ cd swift-sdk-generator
$ swift run swift-sdk-generator make-linux-sdk --swift-branch swift-6.4.x-branch --swift-version 6.4.x-DEVELOPMENT-SNAPSHOT-2026-07-23-a --target aarch64-unknown-linux-gnu
$ swift sdk install ./Bundles/6.4.x-DEVELOPMENT-SNAPSHOT-2026-07-23-a_ubuntu_jammy_aarch64.artifactbundle
$ swift build --swift-sdk 6.4.x-DEVELOPMENT-SNAPSHOT-2026-07-23-a_ubuntu_jammy_aarch64
```

The build fails with some fun errors:

```
Building for debugging...
/usr/bin/x86_64-linux-gnu-ld.bfd: unrecognised emulation mode: aarch64linux
Supported emulations: elf_x86_64 elf_i386 elf32_x86_64 elf_iamcu i386pep i386pe
~/swift-sdk-generator/.build/checkouts/swift-atomics/Package.swift: _AtomicsShims: clang: error: linker command failed with exit code 1 (use -v to see invocation)
error: Ld ~/swift-sdk-generator/.build/out/Products/Debug-linux-aarch64/_AtomicsShims.o normal failed with a nonzero exit code. Command line:     cd ~/swift-sdk-generator/.build/checkouts/swift-atomics
[...]

PR #1575, opened 2026-07-28
[6.4]Respect -use-ld in toolset swift options as an ALTERNATE_LINKER override

owenv:
* Explanation: Some existing Swift SDK toolsets specify -use-ld= in _Swift compiler_ options without specifying a linker path override, assuming that only swiftc will ever be used to link. Map these options to an ALTERNATE_LINKER override so they also apply to linking of C-family targets.
* Scope: Cross compilation using toolsets with -use-ld
* Issues: https://github.com/swiftlang/swift-build/issues/1574
* Risk:
      Low, this is a narrowly targeted change to toolset settings
* Testing: Added unit test, manual testing of the example from the issue

Closes https://github.com/swiftlang/swift-build/issues/1574


owenv:
@swift-ci test

owenv:
@swift-ci test windows

rconnell9 (approved):

jakepetroules (approved):

PR #1576, opened 2026-07-29
Fix indexStorePath/indexDatabasePath being transposed in BSP response

### LINKED ISSUE #10350: `experimental-build-server` reports an `indexStorePath` that doesn't match where `swift build` writes the index store
### Summary

With the `swiftbuild` backend, `swift package experimental-build-server` advertises
`indexStorePath: <scratch>/index-store` in its `build/initialize` response, but the
build itself compiles with `-index-store-path <scratch>/out`. The advertised store is
therefore always empty, and any BSP client that trusts it — notably SourceKit-LSP with
`swiftPM.buildSystem: "swiftbuild"` — gets an empty index and returns no results for
cross-file queries.

It also reports `indexDatabasePath: <scratch>/out`, i.e. the path where the index store
actually lives. The two values look transposed.

### Environment

- Swift 6.4 (`swiftlang-6.4.0.27.1`), Xcode 27.0 beta 4
- BSP server self-reports `"version": "6.4.0-dev"`
- macOS 27.0, arm64
- SwiftPM package with 30 targets; default `swiftbuild` build system

### Reproduction

Drive the BSP server directly over stdio:

```jsonc
// -> build/initialize
{"jsonrpc":"2.0","id":1,"method":"build/initialize","params":{
  "displayName":"probe","version":"1.0","bspVersion":"2.2.0",
  "rootUri":"file:///path/to/Package-dir",
  "capabilities":{"languageIds":["swift"]}}}
```

Launched as:

```
swift package experimental-build-server --build-system swiftbuild --package-path <dir>
```

Response:

```json
{
  "displayName": "SwiftPM Build Server",
  "version": "6.4.0-dev",
  "bspVersion": "2.2.0",
  "capabilities": {},
  "dataKind": "sourceKit",
  "data": {
    "prepareProvider": true,
    "watchers": [],
    "indexDatabasePath": "<pkg>/.build/out",
    "sourceKitOptionsProvider": true,
    "outputPathsProvider": true,
    "indexStorePath": "<pkg>/.build/index-store"
  }
}
```

Now compare against where the build actually puts index data. A `swift build`
compile command (captured from `swift build -v`) contains:

```
-index-store-path <pkg>/.build/out -index-system-modules
```

and on disk:

```
.build/out/v5/units      2748 files
.build/out/v5/records    1281 files
.build/index-store       absent (or empty — created by the client opening it)
```

### Expected

`indexStorePath` should point at the directory the build writes its index store to
(`.build/out` here), and `indexDatabasePath` should point somewhere the client may
create its IndexStoreDB — not the same directory as the store.

### Actual

`indexStorePath` names a directory nothing writes to. A client that opens it sees zero
units.

### Impact

SourceKit-LSP's `swiftPM.buildSystem: "swiftbuild"` path is effectively non-functional:
`textDocument/references` and friends return nothing, with no error surfaced — the
client opens a valid-but-empty store, so nothing looks wrong. Measured on the package
above, with everything else identical:

| Symbol | refs, as advertised | refs, with `.build/index-store` symlinked to `out` |
|---|---|---|
| a type used in 8 files | 0 | 45 |
| a second, older type | 0 | 19 |

### Workaround

```bash
ln -s out .build/index-store
```

Must be recreated whenever `.build` is cleaned.

AnupamKumar-1:
Fixes swiftlang/swift-package-manager#10350

**Note:** the linked issue was filed against `swift-package-manager`, but the root cause is in `swift-build`'s BSP server implementation, so the fix lives here.

## Summary
`indexStorePath` and `indexDatabasePath` in `SWBBuildServer`'s BSP `build/initialize` response were transposed:

- `indexStorePath` derived a sibling directory (`dirname(indexDataStoreFolderPath)/index-store`) that nothing ever writes to.
- `indexDatabasePath` returned `indexDataStoreFolderPath` directly — i.e. the *real* index store location — mislabeled as a database path.

Any BSP client that trusts `indexStorePath` (e.g. SourceKit-LSP with `swiftPM.buildSystem: "swiftbuild"`) opens a valid-but-empty store and silently returns no results for cross-file queries.

## Fix
- `indexStorePath` now returns `arenaInfo.indexDataStoreFolderPath` directly — the actual directory the build writes index data into.
- `indexDatabasePath` now derives a sibling directory (`DataStoreDB`, following the naming convention already used for adjacent Index.noindex paths in `SWBTestSupport/Misc.swift`: `DataStore`, `PrecompiledHeaders`, `Build`) so a client can safely create its own IndexStoreDB without colliding with the raw store.

## Verification
Reproduced and confirmed with a real package, using the `swiftbuild` backend end-to-end:
1. Captured the real `-index-store-path` from a verbose `swift build --build-system swiftbuild -v` compile invocation.
2. Queried `swift package experimental-build-server --build-system swiftbuild` directly over stdio with a raw `build/initialize` request and confirmed the advertised paths didn't match the real one (before fix).
3. After the fix, rebuilt and re-queried: `indexStorePath` now matches the real store path, and `.build/out/v5/{units,records}` are confirmed present exactly where `indexStorePath` claims.
4. `indexDatabasePath` now points at a distinct, non-colliding sibling path.

## Testing
Added `indexStoreAndDatabasePathsAreNotTransposed` in `BuildServerTests.swift`, which constructs a build request with an explicit `SWBArenaInfo` and asserts `indexStorePath` matches the real derived-data path while `indexDatabasePath` remains distinct from it. Verified the test fails without the fix and passes with it.

### DIFF
--- a/Sources/SwiftBuild/SWBBuildServer.swift
+++ b/Sources/SwiftBuild/SWBBuildServer.swift
@@ -36,12 +36,14 @@ public actor SWBBuildServer: QueueBasedMessageHandler {
     private var buildDescriptionID: SWBBuildDescriptionID? = nil
 
     private var indexStorePath: String? {
-        buildRequest.parameters.arenaInfo?.indexDataStoreFolderPath.map {
-            Path($0).dirname.join("index-store").str
-        }
+        // The directory containing index store info written by the compilers.
+        buildRequest.parameters.arenaInfo?.indexDataStoreFolderPath
     }
     private var indexDatabasePath: String? {
-        buildRequest.parameters.arenaInfo?.indexDataStoreFolderPath
+        // This is where a BSP client (e.g. SourceKit-LSP) may create its own IndexStoreDB.
+        buildRequest.parameters.arenaInfo?.indexDataStoreFolderPath.map {
+            Path($0).dirname.join("IndexStoreDB").str
+        }
     }
 
     public let messageHandlingHelper = QueueBasedMessageHandlerHelper(
--- a/Tests/SwiftBuildTests/BuildServerTests.swift
+++ b/Tests/SwiftBuildTests/BuildServerTests.swift
@@ -53,7 +53,7 @@ extension Connection {
     }
 }
 
-fileprivate func withBuildServerConnection(setup: (Path) async throws -> (TestWorkspace, SWBBuildRequest), body: (any Connection, CollectingMessageHandler, Path) async throws -> Void) async throws {
+fileprivate func withBuildServerConnection(setup: (Path) async throws -> (TestWorkspace, SWBBuildRequest), body: (any Connection, CollectingMessageHandler, Path, InitializeBuildResponse) async throws -> Void) async throws {
     try await withTemporaryDirectory { (temporaryDirectory: NamedTemporaryDirectory) in
         try await withAsyncDeferrable { deferrable in
             let tmpDir = temporaryDirectory.path
@@ -74,7 +74,7 @@ fileprivate func withBuildServerConnection(setup: (Path) async throws -> (TestWo
 
             connectionToServer.start(handler: buildServer)
             connectionToClient.start(handler: collectingMessageHandler)
-            _ = try await connectionToServer.send(
+            let initializeResponse = try await connectionToServer.send(
                 InitializeBuildRequest(
                     displayName: "test-bsp-client",
                     version: "1.0.0",
@@ -86,7 +86,7 @@ fileprivate func withBuildServerConnection(setup: (Path) async throws -> (TestWo
             connectionToServer.send(OnBuildInitializedNotification())
             _ = try await connectionToServer.send(WorkspaceWaitForBuildSystemUpdatesRequest())
 
-            try await body(connectionToServer, collectingMessageHandler, tmpDir)
+            try await body(connectionToServer, collectingMessageHandler, tmpDir, initializeResponse)
 
             _ = try await connectionToServer.send(BuildShutdownRequest())
             connectionToServer.send(OnBuildExitNotification())
@@ -185,7 +185,7 @@ fileprivate struct BuildServerTests: CoreBasedTests {
             request.parameters.activeRunDestination = .host
 
             return (testWorkspace, request)
-        }) { connection, handler, _ in
+        }) { connection, handler, _, _ in
             #expect(handler.notifications.withLock { notifications in
                 notifications.contains { notification in
                     notification is OnBuildTargetDidChangeNotification
@@ -257,7 +257,7 @@ fileprivate struct BuildServerTests: CoreBasedTests {
             request.parameters.activeRunDestination = .host
 
             return (testWorkspace, request)
-        }) { connection, _, tmpDir in
+        }) { connection, _, tmpDir, _ in
             let targetsResponse = try await connection.send(WorkspaceBuildTargetsRequest())
             let target = try #require(targetsResponse.targets.only)
             let sourcesResponse = try await connection.send(BuildTargetSourcesRequest(targets: [target.id]))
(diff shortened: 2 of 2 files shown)


AnupamKumar-1:
@swift-ci test

Sources/SwiftBuild/SWBBuildServer.swift
```diff
@@ -36,12 +36,19 @@ public actor SWBBuildServer: QueueBasedMessageHandler {
     private var buildDescriptionID: SWBBuildDescriptionID? = nil
 
     private var indexStorePath: String? {
-        buildRequest.parameters.arenaInfo?.indexDataStoreFolderPath.map {
-            Path($0).dirname.join("index-store").str
-        }
+        // `arenaInfo.indexDataStoreFolderPath` is the directory `swift build`/swiftc actually
```
owenv: This comment doesn't really make sense without the context of the original issue, IMO it should only keep the first half, "the directory containing index store info written by the compilers"

Sources/SwiftBuild/SWBBuildServer.swift
```diff
@@ -36,12 +36,19 @@ public actor SWBBuildServer: QueueBasedMessageHandler {
...
+        // writes index data into, so report it directly rather than deriving a sibling path
+        // that nothing writes to.
+        buildRequest.parameters.arenaInfo?.indexDataStoreFolderPath
     }
     private var indexDatabasePath: String? {
-        buildRequest.parameters.arenaInfo?.indexDataStoreFolderPath
+        // This is where a BSP client (e.g. SourceKit-LSP) may create its own IndexStoreDB.
```
owenv, line 43: Same here, I would drop all but the first line of the comment

Sources/SwiftBuild/SWBBuildServer.swift
```diff
@@ -36,12 +36,19 @@ public actor SWBBuildServer: QueueBasedMessageHandler {
...
+        // It must not be the same directory as the index store itself, so derive a sibling
+        // location, following the repo's `Index.noindex/<Name>` convention (see
+        // `SWBTestSupport/Misc.swift`: DataStore, PrecompiledHeaders, Build).
+        buildRequest.parameters.arenaInfo?.indexDataStoreFolderPath.map {
+            Path($0).dirname.join("DataStoreDB").str
```
owenv: IndexStoreDB might match existing conventions a bit better than DataStoreDB

Tests/SwiftBuildTests/BuildServerTests.swift
```diff
@@ -516,5 +516,83 @@ fileprivate struct BuildServerTests: CoreBasedTests {
...
+        }) { _, _, tmpDir, initializeResponse in
+            let data = try #require(SourceKitInitializeBuildResponseData(fromLSPAny: initializeResponse.data))
+
+            // `indexStorePath` must be the real store location the build writes to, not a
+            // derived sibling directory nothing writes to.
+            #expect(data.indexStorePath == tmpDir.join("out").str)
+
+            // `indexDatabasePath` must be distinct from the store itself, so a client can
+            // safely create its own IndexStoreDB there without colliding with the raw store.
+            #expect(data.indexDatabasePath != data.indexStorePath)
```
owenv: Specifically trying to detect these being transposed seems fragile since it's unlikely a future regression would break things in exactly the same way. It might be preferable to just check for the exact expected path here so it has to be manually updated whenever that changes.

owenv (commented):
Thanks, I have a few comments but this seems reasonable

owenv:
@swift-ci test

Sources/SwiftBuild/SWBBuildServer.swift
AnupamKumar-1: Done, simplified the comment.

Sources/SwiftBuild/SWBBuildServer.swift
AnupamKumar-1: Done, trimmed the comment to a single line.

Sources/SwiftBuild/SWBBuildServer.swift
AnupamKumar-1: Renamed to IndexStoreDB.

Tests/SwiftBuildTests/BuildServerTests.swift
AnupamKumar-1: Replaced the != check with an exact expected-path assertion.

owenv (approved):

owenv:
@swift-ci test

PR #1577, opened 2026-07-29
[6.4.x] Fix link command line filtering when using dynamic replacement previews with swiftc as the linker driver

owenv:
* Explanation: Dynamic replacement based preview info relies on being able to filter various arguments from the linker command line. Some of these rules needed updates to support using swiftc as the linker driver for package targets containing swift sources.
* Scope: Dynamic replacement preview info in packages
* Risk:
Low, this is narrowly scoped to only affect dynamic replacement previews
* Testing: Updated existing tests


owenv:
@swift-ci test

jakepetroules (approved):

PR #1580, opened 2026-07-29
Move Mig filetype to AppleFileTypes.xcspec

owenv:
Otherwise MiG.xcspec mixes specs from the Darwin and empty domains, and fails validation when loaded in Xcode.


owenv:
@swift-ci test

mhrawdon (approved):

PR #1589, opened 2026-08-04
Add a build setting which allows overriding the display name reported by the build server for a target

owenv:
SourceKit-LSP surfaces these display names to the user, so we don't want to use the PIF target name which may include disambiguating suffixes. Follow the model used by tags to provide a build setting PIF clients can use to override what's reported


owenv:
@swift-ci test

rjmansfield (approved):

PR #1590, opened 2026-08-04
Fix ipi clang module derivation logic

sepy97:
Currently, clang module is considered ipi if the modulemap is located in the srcroot. It can be still installed and imported from the install location (e.g. XCTest), and, therefore, should not always be considered project internal. The path-based heuristic should be applied only when the install path is empty (to verify the module is not installed in the sdk).

rdar://184041007


sepy97:
@swift-ci please test

sepy97:
@swift-ci please test

owenv (approved):

PR #1595, opened 2026-08-05
Add a build setting for passing `-register-module-dependency` options to the Swift compiler driver

abertelrud:
## Summary

Adds a new string-list build setting called `SWIFT_REGISTERED_MODULE_DEPENDENCIES` that causes one or more `-register-module-dependency` options to be passed to the Swift compiler driver (one for each string in the value).

## Motivation

The Swift compiler driver has a `-register-module-dependency` flag that was introduced in Swift 6.3 in https://github.com/swiftlang/swift/pull/85394. It allows the caller to explicitly specify the name of a module that should be added to the set of discovered names of module dependencies. This is useful for edge cases in which, for example, the caller knows that a transient dependency (such as `Playgrounds`) will be needed for a later incremental build, and it wants to make sure that this module is treated as if it had already been found during an earlier dependency scan.

Because `-register-module-dependency` is an official flag of the compiler driver, it would be cleaner to have an official build setting for controlling it, than to require callers to use `OTHER_SWIFT_FLAGS` to pass `-register-module-dependency`.

## Specific changes

- SWBCore: declare the `SWIFT_REGISTERED_MODULE_DEPENDENCIES` macro.

- Swift.xcspec: map it to `-register-module-dependency,` The StringList `CommandLineFlag` entry causes the flag to be emitted once per element; no SwiftCompiler.swift change is needed.

- SWBProjectModel: expose it as a typed `[String]?` PIF setting so that build system clients can pass it down.

- SWBTaskContructionTests: add a new test `SwiftTaskConstructionTests.registeredModuleDependencies()` for this functionality.

## Comments

There is no change in behaviour unless the build setting `SWIFT_ADDITIONAL_IMPORTABLE_MODULES` is set, and so this is safe to land at any time.

I considered naming the new build setting `SWIFT_ADDITIONAL_IMPORTABLE_MODULES`, which better captures the effect of adding the name of a module to this build setting, but it seemed better to match the existing name of the option in the Swift compiler driver.


owenv (approved):

abertelrud:
@swift-ci test

artemcm (approved):

abertelrud:
The many failing tests do look unrelated.  Is this OK to land even so?

abertelrud:
@owenv  Thank you for the review and approval.  It looks as if the required builds have passed, but there are several failures that look unrelated as far as I can tell.  Is this ok to merge based on required builds passing or should I investigate the other failures?  Thanks!

abertelrud:
(by the way, I did look into many of the tests and a lot seem to be failing with `fatalError`, which did look unrelated)

owenv:
@abertelrud if you rebase the PR a lot of the linux failures should resolve themselves, a handful of others are known issues

abertelrud:
> @abertelrud if you rebase the PR a lot of the linux failures should resolve themselves, a handful of others are known issues

Great, thank you.  I will do that.

abertelrud:
@swift-ci test

PR #1599, opened 2026-08-05
 Add Opt-in Clang Explicitly Built Modules Support for C++/ObjC++ Sources

qiongsiwu:
Enables explicitly built modules (EBM) for C++ and Objective-C++ sources, staged behind `_CLANG_EXPLICIT_MODULES_ALLOW_CXX` (default `NO`) so this change is a no-op unless deliberately turned on. This PR also teaches `swift-build` to look for a block list of projects so it does not enable EBM for C++/Objective-C++ sources for projects on the block list.


qiongsiwu:
@swift-ci please test

owenv (approved):

cyndyishida:
Why do we need separate settings? My feeling is that it adds more staging steps than we likely need.  I think we should have a blocklist just for `CLANG_ENABLE_EXPLICIT_MODULES` and opt objective-c++/c++ sources in the same setting.

qiongsiwu:
> Why do we need separate settings? My feeling is that it adds more staging steps than we likely need.

My understanding of the staging process is that we want to turn on `CLANG_ENABLE_EXPLICIT_MODULES` for C++/ObjC++ sources, and run world builds to figure out the content of the block list. I checked with @owenv and it seems to be rather cumbersome to run world builds with a custom `swift-build`. The suggestion is to land the enablement with the gating option default to off, and then run world build using a submitted `swift-build` with the gating option turned on. 

>  I think we should have a blocklist just for `CLANG_ENABLE_EXPLICIT_MODULES` and opt objective-c++/c++ sources in the same setting.

What do we mean exactly by `opt objective-c++/c++ sources in the same setting`? Does this mean we want to flip `CLANG_ENABLE_EXPLICIT_MODULES` on only conditioning on EBM on and the block list? I don't quite see how that can work with an empty block list.

qiongsiwu:
@swift-ci please test.

Issue #1479, opened 2026-06-24
'Create Build Description' performance regression in Xcode 27 beta 2


stjernegard:
When building our app with Xcode 27 beta 2 (27A5209h), the build hangs for a very long on the "Create build description" step. 
I've narrowed it down to a specific dependency we use, which has some other C dependencies. 
This does not happen when building with Xcode 27 beta 1.

To reproduce, you can check out https://github.com/SDWebImage/SDWebImageSVGNativeCoder, open it in Xcode 27b2 and simply press Build.

<img width="1572" height="548" alt="Image" src="https://github.com/user-attachments/assets/6d99a57f-3d35-4d29-a134-acf4ac7aadb1" />

stjernegard:
This is still an issue with the release of Xcode 27 beta 3.

Attaching a sample of SWBBuildService during the build description step in case that's helpful.
[Sample of SWBBuildService.txt](https://github.com/user-attachments/files/29735186/Sample.of.SWBBuildService.txt)

<img width="1200" height="520" alt="Image" src="https://github.com/user-attachments/assets/ad0a03e7-17da-444a-9267-a81855fa43b7" />

gsapienza:
Also experienced this. 'Create Build Description' would take minutes to complete on even a small app. Had to remove SDWebImageSVGNativeCoder to get build times back to normal.

PR #1624, opened 2026-08-12
Fix build file ordering assignment

stjernegard:
This array concat and write was being done for every loop iteration. As the array is being overwritten every time, significant amounts of time can be saved by only writing after the loop.

This fixes the performance regression reported in #1479 and during a build of [SDWebImage/SDWebImageSVGNativeCoder](https://github.com/SDWebImage/SDWebImageSVGNativeCoder) reduces the time taken during "Create build description" from minutes to 1-2 seconds.

A trace from Instruments helped me track down the issue, which showed 99.7% of the entire build time was spent assigning and re-assigning this array:
Screenshot 2026-08-12 at 11 46 29

Running the same trace after this change reduces the amount of time spent in that function to almost nothing.


rjmansfield:
@swift-ci please test

rjmansfield (approved):

owenv (approved):
lgtm thanks. I think we should consider cherrypicking to release/6.4.x as well

jakepetroules (approved):

PR #1625, opened 2026-08-12
Only resolve xcframework paths for actual xcframeworks

rjmansfield:
XCFrameworkTaskProducer.prepare() resolved every build file reachable through package products to find the xcframeworks. Since a file reference's type is known without resolving its absolute path, resolve the path only for xcframeworks.

rdar://184265424


rjmansfield:
@swift-ci please test

owenv (approved):

PR #1628, opened 2026-08-12
[6.4] Fix build file ordering assignment

owenv:
* Explanation: From original PR by @stjernegard:
This array concat and write was being done for every loop iteration. As the array is being overwritten every time, significant amounts of time can be saved by only writing after the loop.

This fixes the performance regression reported in #1479 and during a build of SDWebImage/SDWebImageSVGNativeCoder reduces the time taken during "Create build description" from minutes to 1-2 seconds.
* Scope: Addresses performance issue in task construction
* Risk:
Low, this is a targeted change covered by existing tests
* Testing: Covered by existing unit tests


owenv:
@swift-ci test

rjmansfield (approved):

jakepetroules (approved):

rconnell9 (approved):

PR #1631, opened 2026-08-12
Fix Android platform filter matching

owenv:
It's important to use SWIFT_TARGET_TRIPLE instead of CURRENT_TARGET_TRIPLE when stripping the version number from the environment of an Android triple during platform filter matching, because it permits an unbound architecture.


owenv:
@swift-ci test

mhrawdon (approved):

PR #1632, opened 2026-08-12
Add SWIFT_EXPLAIN_IPI_LIBRARY_LEVEL to explain IPI classification dec…

sepy97:
…isions

Adds a debug-only build setting that emits notes explaining why a Swift or Clang module was classified project-internal.

rdar://184747937

### DIFF
--- a/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
+++ b/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
@@ -1633,6 +1633,23 @@ public final class SwiftCompilerSpec : CompilerSpec, SpecIdentifierType, SwiftDi
             if toolSpecInfo.toolFeatures.has(.libraryLevel),
                let libraryLevel = cbc.scope.evaluateAsString(BuiltinMacros.SWIFT_LIBRARY_LEVEL).nilIfEmpty {
                 args += ["-library-level", libraryLevel]
+                if libraryLevel == "ipi", cbc.scope.evaluate(BuiltinMacros.SWIFT_EXPLAIN_IPI_LIBRARY_LEVEL) {
+                    // Explained here rather than where SWIFT_LIBRARY_LEVEL is derived, so it is
+                    // reported only for targets that actually compile Swift (this task only runs
+                    // for those).
+                    // Re-checks the same condition Settings.swift uses to infer the level (search
+                    // "infer the default SWIFT_LIBRARY_LEVEL") to distinguish derived from explicit.
+                    if cbc.scope.evaluate(BuiltinMacros.SWIFT_ENABLE_IPI_LIBRARY_LEVEL) && cbc.scope.evaluate(BuiltinMacros.SKIP_INSTALL) {
+                        delegate.note("passing -library-level ipi (derived because SKIP_INSTALL=YES)")
+                    } else {
+                        delegate.note("passing -library-level ipi (SWIFT_LIBRARY_LEVEL set explicitly)")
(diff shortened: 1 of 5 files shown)


Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
```diff
@@ -1633,6 +1633,23 @@ public final class SwiftCompilerSpec : CompilerSpec, SpecIdentifierType, SwiftDi
             if toolSpecInfo.toolFeatures.has(.libraryLevel),
                let libraryLevel = cbc.scope.evaluateAsString(BuiltinMacros.SWIFT_LIBRARY_LEVEL).nilIfEmpty {
                 args += ["-library-level", libraryLevel]
+                if libraryLevel == "ipi", cbc.scope.evaluate(BuiltinMacros.SWIFT_EXPLAIN_IPI_LIBRARY_LEVEL) {
+                    // Explained here rather than where SWIFT_LIBRARY_LEVEL is derived, so it is
+                    // reported only for targets that actually compile Swift (this task only runs
+                    // for those).
+                    // Re-checks the same condition Settings.swift uses to infer the level (search
+                    // "infer the default SWIFT_LIBRARY_LEVEL") to distinguish derived from explicit.
+                    if cbc.scope.evaluate(BuiltinMacros.SWIFT_ENABLE_IPI_LIBRARY_LEVEL) && cbc.scope.evaluate(BuiltinMacros.SKIP_INSTALL) {
```
owenv, line 1642: Since this is a debug setting, I don't think it does a lot of harm if it's overly verbose. Would that let us avoid duplicating the logic here?

sepy97: It's not just gonna be overly verbose, but also confusing; without this verification we might emit same note for unrelated targets or emit multiple (conflicting) notes, e.g.:
```
Clang module 'ClangModulemapIPI' is IPI: SKIP_INSTALL=YES  
Clang module 'ClangModulemapIPI' is IPI: its module map in '…' is under SRCROOT '…'
```

Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
```diff
@@ -1648,6 +1665,10 @@ public final class SwiftCompilerSpec : CompilerSpec, SpecIdentifierType, SwiftDi
                     for name in ipiNames {
                         args += ["-Xfrontend", "-ipi-clang-module", "-Xfrontend", name]
                     }
+                } else if cbc.scope.evaluate(BuiltinMacros.SWIFT_EXPLAIN_IPI_LIBRARY_LEVEL) {
+                    // Just for the explanation purposes -- to cover the rare case when ipi modules are not passed
+                    // when neither driver not frontend supports ipi library level feature
+                    delegate.note("not passing -ipi-clang-module for \(ipiNames.joined(separator: ", ")): neither the driver nor the frontend supports it")
```
nkcsgexi, line 1671: What is the difference between this note and the previous one? aka `not passing -library-level ipi: the toolchain does not support it`

sepy97: One is relevant when we have -library-level ipi flag added but toolchain does not support it (the `not passing -library-level ipi: the toolchain does not support it`). The other is for ipi-clang-module flag support (by driver and frontend). I guess, phrasing of the note could be identical instead (besides the flag name logic remains the same)

PR #1633, opened 2026-08-12
[Previews] Fix XOJIT thunk builds when Swift caching prefix mapping is on

cachemeifyoucan:
With compilation caching prefix mapping enabled, swift-driver rewrites the paths in the frontend invocation it plans, so the preview thunk is asked to compile `/^src/main.swift`. Those pseudo-paths only resolve against the CAS file system, which the uncached preview build deliberately does not set up because `-module-import-from-cas` is meant to read sources from the real file system, so the thunk build failed with "error opening input file '/^src/...'". Remapped paths recorded inside the module artifacts loaded from the CAS, such as the headers a PCM refers to, fail the same way and cannot be reached by rewriting the command line at all.

Pass a `directory-remap` VFS overlay that maps every prefix back onto the directory it stands for, built by inverting the mapping the driver reports in `-cache-replay-prefix-map`. Those arguments are left in place because the compiler needs them to locate macro plugins, which are dlopen'd outside of any VFS. `PrefixMapper` mirrors `llvm::PrefixMapper` so the ordering and matching agree with what the compiler and driver do.

The thunk substitution overlay is now keyed on the prefix mapped spelling too, since that is the path the frontend looks up, and is passed last so it wins over the prefix map overlay. Getting that wrong does not fail the build, it silently compiles the original source instead of the thunk, so the new test asserts on the symbols in the resulting object file. For the same reason the prefix map overlay is deliberately not named after the thunk: Previews requests thunk info once using a placeholder variant suffix and then substitutes the real suffix into the command line for each variant it builds, regenerating only the thunk overlay, so a name derived from the thunk would point at a file nobody writes.

rdar://184756825

### DIFF
--- a/Sources/SWBUtil/PrefixMapper.swift
+++ b/Sources/SWBUtil/PrefixMapper.swift
@@ -1,0 +1,83 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Rewrites path prefixes, mirroring `llvm::PrefixMapper`.
+///
+/// This reproduces, in Swift Build, the path canonicalization the compiler and `swift-driver`
+/// perform for compilation caching, where real paths are replaced by stable pseudo-paths such as
+/// `/^src` so that cache keys don't depend on where the sources live.
+///
+/// Ordering matters: mappings are applied in order and the first match wins, so `sort()` must be
+/// called before mapping to produce the same result the compiler would.
+public struct PrefixMapper: Sendable {
+    /// A single prefix replacement, rewriting the prefix `from` to `to`.
+    public struct Mapping: Sendable, Equatable {
+        public let from: Path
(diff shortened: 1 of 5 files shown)


cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:
@swift-ci please test

Sources/SWBUtil/PrefixMapper.swift
```diff
@@ -0,0 +1,83 @@
...
+        }
+        return path
+    }
+
+    /// The inverse mapper, exchanging `from` and `to` in every mapping.
+    ///
+    /// The result is sorted, because the order that was correct for the forward direction says
+    /// nothing about the order the inverse direction needs.
+    public func inverted() -> PrefixMapper {
+        PrefixMapper(mappings: mappings.map { Mapping(from: $0.to, to: $0.from) }).sorted()
```
owenv, line 81: Is it ever correct for one forward mapping to map to a prefix of the result of another forward mapping? If that never happens I'm not sure it's necessary to sort when constructing the inverted mapping, though I suppose it doesn't hurt

owenv (approved):

Sources/SWBUtil/PrefixMapper.swift
cachemeifyoucan: It is to avoid common prefixes and produce a deterministic ordering for that. E.g., /var/tmp -> /^tmp and /var/tmp/out -> /^out. But I guess when going for inverse mapping, this might not be needed since usually we don't do compounded path for mapped path but just in case.

PR #1636, opened 2026-08-14
Support reporting task cache keys in a format clients can inspect

owenv:
Add new build event messages which report the CAS ID and CAS config corresponding to the cache key of cached tasks. The intent is that clients can build on this and use this data to diff cache keys across builds


owenv:
@swift-ci test

cachemeifyoucan (approved):

rjmansfield (approved):

PR #1637, opened 2026-08-14
Track Clang module directory-listing dependencies for correct incremental rebuilds

daveinglis:
Umbrella and framework modules depend on the set of headers in a directory, not just specific files. Adding or removing a header must rebuild any module that enumerated it, but Swift Build only tracked concrete header files and so could reuse a stale PCM.

Clang's scanner (swiftlang/llvm-project#13797) now reports these via clang_experimental_DepGraphModule_getDirectoryDeps. Consume them and register each directory as an llbuild directory-tree dependency on both the scan task (to force a rescan) and the module precompile task (so the PCM itself rebuilds). SDK/system directories are filtered out, and the .scan aggregate count bumps 10 -> 11.

Runtime capability-gated: active only when the loaded libclang exports the symbol, and a no-op on older toolchains. Adds serialization unit tests and an umbrella-directory integration test.

### DIFF
--- a/Sources/SWBTaskExecution/DynamicTaskSpecs/ClangModuleDependencyGraph.swift
+++ b/Sources/SWBTaskExecution/DynamicTaskSpecs/ClangModuleDependencyGraph.swift
@@ -70,6 +70,13 @@ package final class ClangModuleDependencyGraph {
         /// The list of file (paths) dependencies that are required by the translation unit or module.
         package let files: [Path]
 
+        /// The list of directories whose contents (directory listing) are required by the translation unit or module.
+        ///
+        /// If a file is added to or removed from one of these directories, the translation unit or module must be
+        /// rebuilt. These are reported by the Clang dependency scanner for umbrella and framework modules, and are
+        /// empty when the loaded libclang does not support directory dependencies.
+        package let directories: [Path]
+
         /// The CASID of the include-tree required by this translation unit or module, if any.
         package let includeTreeID: String?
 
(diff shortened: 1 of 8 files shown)


daveinglis:
@swift-ci test

daveinglis:
new tests pass locally with the new clang toolchain that supports the new API

Sources/SWBTaskExecution/DynamicTaskSpecs/ClangModuleDependencyGraph.swift
```diff
@@ -382,6 +398,20 @@ package final class ClangModuleDependencyGraph {
                                 $0.append(contentsOf: fileDependencies)
                             }
 
+                            let directoryDependencies = module.directory_deps.map(Path.init).filter {
+                                // Exclude SDK/system directories (e.g. system framework header directories) for the
```
owenv: not sure this comment is adding much

Sources/SWBTaskExecution/TaskActions/ClangScanTaskAction.swift
```diff
@@ -181,6 +182,9 @@ public final class ClangScanTaskAction: TaskAction, BuildValueValidatingTaskActi
         }
 
         var dependencyPaths = result.dependencyPaths
+        // Directory-listing dependencies (`result.directoryPaths`) are already filtered against the SDK/system
```
owenv: same here, I don't think the comment is needed

Tests/SWBTaskExecutionTests/ClangModuleDependencyGraphTests.swift
```diff
@@ -0,0 +1,91 @@
+//===----------------------------------------------------------------------===//
```
owenv: I'm not sure these serialization tests are adding much value

owenv (approved):

owenv:
lgtm, a few minor optional comments

daveinglis:
@swift-ci test

PR #1640, opened 2026-08-14
Annotate two ui test task construction tests as requiring Xcode 27

owenv:
These no longer check for copy tasks of the testing frameworks, so require Xcode 27 to pass


owenv:
@swift-ci test

mhrawdon (approved):

PR #1646, opened 2026-08-17
Expand build description signature env denylist to more session vars

rjmansfield:
Terminal/shell session variables vary between otherwise identical invocations and don't affect build output, but were hashed into the signature and invalidated the build description cache.

rdar://182103048


rjmansfield:
@swift-ci test

owenv (approved):

PR #1648, opened 2026-08-18
Fix `.when(platforms: [.wasi])` to match with `wasm32-unknown-wasip1-threads`

kateinoigakukun:
The build-time platform filter for a wasm destination comes out as ("wasip1", "threads"): the synthesized Swift SDK sets SWIFT_PLATFORM_TARGET_PREFIX from LLVMTargetTripleSys ("wasip1") and LLVM_TARGET_TRIPLE_SUFFIX from LLVMTargetTripleEnvironment ("-threads"). The .wasi platform condition previously emitted only environment-less filters ("wasi", "")/("wasip1", ""), and PlatformFilter.matches() demands exact equality on (platform, environment), so no .wasi condition could ever match a `*-threads` triple.

### DIFF
--- a/Sources/SWBProjectModel/PIFGenerationModel.swift
+++ b/Sources/SWBProjectModel/PIFGenerationModel.swift
@@ -1108,6 +1108,9 @@ public extension Array where Element == PIF.BuildSettings.Platform {
 
             case .wasi:
                 result.append(.init(platform: "wasi"))
+                result.append(.init(platform: "wasip1"))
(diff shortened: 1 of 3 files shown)


kateinoigakukun:
@swift-ci test

jakepetroules (approved):

jakepetroules:
This seems like a fix we'd definitely want in 6.4.x as well.

MaxDesiatov (approved):

rconnell9 (approved):

PR #1649, opened 2026-08-18
Fix explicit access of Foundation imports in resource bundle accessor

EhsanAzish80:
## Summary

- generate resource bundle accessor imports with explicit `internal` access
- cover both named and empty resource bundles

This is the SwiftBuild side of swiftlang/swift-package-manager#10163.

## Testing

- `swift test --filter PackageProductConstructionTests`
- reproduced the reported failure before the change
- verified the generated accessor builds with `-warnings-as-errors` after the change


rjmansfield:
@swift-ci test

owenv (changes requested):
This appears largely the same as https://github.com/swiftlang/swift-build/pull/1447, with the same issues as that PR

EhsanAzish80:
You’re right. The macOS CI failure confirms this also breaks targets using an implicit Foundation import in Swift 5 mode, so an unconditional `internal import` has the same compatibility problem as #1447. Thanks for pointing it out—I’m closing this.

PR #1654, opened 2026-08-19
Use per-file header inputs for framework InstallAPI dependency tracking

rjmansfield:
Depend on a node per installed public/private header instead of a directory-tree signature over the framework's installed header directories, whose contents listing can be captured before a directory settles and spuriously re-runs InstallAPI an incremental build.

rdar://182482020

### DIFF
--- a/Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/ProductPostprocessingTaskProducer.swift
+++ b/Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/ProductPostprocessingTaskProducer.swift
@@ -761,11 +761,15 @@ private extension ProductTypeSpec {
         }()
 
         if isFramework && !requiresSRCROOTSupport {
-            if !(targetHeaderInfo?.publicHeaders.filter({ !filteringContext.isExcluded(header: $0) }) ?? []).isEmpty {
-                tapiInputNodes.append(producer.context.createDirectoryTreeNode(TargetHeaderInfo.destDirPath(for: .public, scope: scope), excluding: []))
+            // Framework public/private headers are excluded from the `-filelist` above, so depend on a node per installed header instead of a directory-tree signature
+            // whose contents listing can be captured before the directory settles.
+            for entry in targetHeaderInfo?.publicHeaders ?? [] {
+                guard !filteringContext.isExcluded(header: entry), let path = filteringContext.path(header: entry) else { continue }
+                tapiInputNodes.append(producer.context.createNode(TargetHeaderInfo.outputPath(for: path, visibility: .public, scope: scope).normalize()))
             }
-            if !(targetHeaderInfo?.privateHeaders.filter({ !filteringContext.isExcluded(header: $0) }) ?? []).isEmpty {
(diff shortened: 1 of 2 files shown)


rjmansfield:
@swift-ci test

Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/ProductPostprocessingTaskProducer.swift
```diff
@@ -761,11 +761,15 @@ private extension ProductTypeSpec {
...
+            // whose contents listing can be captured before the directory settles.
+            for entry in targetHeaderInfo?.publicHeaders ?? [] {
+                guard !filteringContext.isExcluded(header: entry), let path = filteringContext.path(header: entry) else { continue }
+                tapiInputNodes.append(producer.context.createNode(TargetHeaderInfo.outputPath(for: path, visibility: .public, scope: scope).normalize()))
             }
-            if !(targetHeaderInfo?.privateHeaders.filter({ !filteringContext.isExcluded(header: $0) }) ?? []).isEmpty {
-                tapiInputNodes.append(producer.context.createDirectoryTreeNode(TargetHeaderInfo.destDirPath(for: .private, scope: scope), excluding: []))
+            for entry in targetHeaderInfo?.privateHeaders ?? [] {
+                guard !filteringContext.isExcluded(header: entry), let path = filteringContext.path(header: entry) else { continue }
+                tapiInputNodes.append(producer.context.createNode(TargetHeaderInfo.outputPath(for: path, visibility: .private, scope: scope).normalize()))
```
owenv, line 772: IIRC some of this is setup the way it is because a fair number of existing projects have scripts which copy headers into these directories without informing the build system. If we keep both the old directory inputs and the new per-file inputs, does it still order things correctly?

rjmansfield: Unfortunately keeping both doesn't work. The problem is that keeping the directory inputs introduces the churn in the next build. I tried must-scan-after on the directory node, which fixed small frameworks but still churned on frameworks with a larger number of headers.

Supporting the unknown header inputs is a bit tricky so I guess I'll have to explore other options.

owenv: hmm, all in all I wonder if this is still a bigger improvement

rjmansfield: I don't think the directory node gives reliable coverage there for undeclared copies, it’s an ungated scan, so it invalidates late, and re-runs on files tapi never reads.

What if we switched to invoking tapi to emit .d files for headers it parsed?  This would capture undeclared headers and track them. But the gap would be undeclared headers appearing in the incremental case..

owenv: I like that plan (alongside this change)

owenv (approved):

PR #1658, opened 2026-08-20
Gate -final_output on deployment location, not postprocessing

rjmansfield:
-final_output records the product's install path, so it applies whenever the product is deployed not only when postprocessing runs. Under a deployed layout a build and a build-for-install are the same link task, and using DEPLOYMENT_POSTPROCESSING results in churn.

rdar://182546393


rjmansfield:
@swift-ci test

owenv (approved):

rjmansfield:
@swift-ci test windows

PR #1660, opened 2026-08-21
Track InstallAPI header dependencies via tapi's discovered deps

rjmansfield:
Complement the per-file header nodes: tapi emits the headers it parsed as a discovered-dependency file, so GenerateTAPI reruns when any consumed header changes.

rdar://182482020


rjmansfield:
@swift-ci test

owenv (approved):

rjmansfield:
@swift-ci test

PR #1661, opened 2026-08-21
Avoid computing product names in the wrong context when resolving implicit dependencies

owenv:
When computing implicit dependencies, product reference names were being resolved in the context of the dependent rather than the dependency. This caused them to be computed incorrectly, effectively causing a target to depend on itself. This false dependency would usually be dropped, except where multiple targets in a workspace shared a product name, in which case we would incorrectly infer a dependency between them.

Add a resolveParameterizedProductName parameter to control product name resolution behavior. Currently, parameterized product names can only appear in packages, and packages are never allowed as an implicit dependency, so we can just skip resolving the product name altogether to avoid the incorrect dependency.


owenv:
@swift-ci test

rjmansfield (approved):

owenv:
@swift-ci test windows

daveinglis (approved):

owenv:
rebasing

owenv:
@swift-ci test

PR #1666, opened 2026-08-24
Deduplicate Swift -add_ast_path arguments when linking

rjmansfield:
Each Swift module emits a response file listing `-add_ast_path` for its whole transitive module closure. A product linking many libraries that share a closure repeats each AST path once per dependent, which can exceed ARG_MAX.

rdar://185323523


rjmansfield:
@swift-ci test

owenv:
When using a current compiler we generally shouldn't hit this code path because we'll drop the args in favor of explicit modules based debug info, but this doesn't add a ton of complexity so I think it's ok if you think it'll help in the fallback case

owenv (approved):

rjmansfield:
> When using a current compiler we generally shouldn't hit this code path because we'll drop the args in favor of explicit modules based debug info, but this doesn't add a ton of complexity so I think it's ok if you think it'll help in the fallback

Yeah, I noticed that. The reporter was using a 6.3 toolchain and hitting ARG_MAX exhaustion, and I thought it was a recent enough toolchain that we might want to handle it in swift-build.

PR #1669, opened 2026-08-25
Update automerge workflow for new release branches

owenv:
https://forums.swift.org/t/swift-6-4-release-process/85421/20

I expect these branches will mostly use cherry-picks, but the workflow is helpful to ensure anything picked to an earlier release makes it to all subsequent releases as well


owenv:
@swift-ci test

owenv:
@swift-ci test

rjmansfield (approved):

PR #1670, opened 2026-08-25
Stop asserting specific recommended deployment target values

owenv:
The recommended value is a property of the SDK, not the build system version. Just assert that we're setting it correctly, but don't check for a specific value which will vary from SDK to SDK


owenv:
@swift-ci test

rjmansfield (approved):

PR #1671, opened 2026-08-26
Don't link a host macro's object into a cross-platform test bundle

### LINKED ISSUE #1642: Xcode links a same-package macro's macOS testable object into an iOS test bundle when SwiftSyntax prebuilts are enabled
## Summary

When a Swift package contains all three targets below in the same package,
Xcode links the macro target's host-only testable object into the iOS Simulator
test bundle when SwiftSyntax prebuilts are enabled:

```text
DemoKitTests (.testTarget) -> DemoKit (.target) -> DemoKitMacros (.macro)
```

`DemoKitTests` has no direct dependency on `DemoKitMacros`. The failure occurs
even when the test does not invoke or reference the macro.

This reproduces with Xcode 26.6 and Xcode 27 beta 4. Moving the macro into a
separate local child package avoids the failure while keeping prebuilts enabled.

## Expected behavior

The macro executable and its SwiftSyntax dependencies should be built for and
loaded on the macOS host. Its relocatable/testable macOS object should not be a
link input of the iOS Simulator test bundle.

## Actual behavior

The generated target dependency graph makes `DemoKitTests-product` depend on
`DemoKitMacros`, and the test bundle's link file list contains the macro's
testable host object:

```text
Target 'DemoKitTests-product' in project 'DemoKit'
    -> Explicit dependency on target 'DemoKitMacros' in project 'DemoKit'
    -> Explicit dependency on target 'DemoKitMacros' in project 'DemoKit'
    -> Explicit dependency on target 'DemoKit' in project 'DemoKit'
```

The iOS Simulator link then fails:

```text
ld: building for 'iOS-simulator', but linking in object file
(.../Build/Products/Debug/ExecutableModules/
DemoKitMacros--228C090CAA851E1-testable.o) built for 'macOS'
```

With Xcode 27 beta 4, the log confirms that the SwiftSyntax MacroSupport
prebuilt is selected before the failure:

```text
Downloading package prebuilt
https://download.swift.org/prebuilts/swift-syntax/603.0.2/
swiftlang-6.4.0.27.1-macosx27.0-MacroSupport.zip
```

## Steps to reproduce

1. Download and extract [DemoKit.zip](https://github.com/user-attachments/files/31116100/DemoKit.zip).
2. From the extracted `DemoKit` directory, run against any available iOS
   Simulator, starting with clean DerivedData:

   ```shell
   xcodebuild test \
     -scheme DemoKit \
     -destination 'platform=iOS Simulator,name=iPhone 16 Pro' \
     -derivedDataPath /tmp/DemoKitDerivedData \
     -IDEPackageEnablePrebuilts=YES
   ```

## Controls and workarounds

- Passing `-IDEPackageEnablePrebuilts=NO` makes the same-package layout build
  and test successfully, but requires building SwiftSyntax from source.
- Moving `DemoKitMacros` into a local child package makes the test pass with
  `-IDEPackageEnablePrebuilts=YES`.
- Building the same target graph with the SwiftPM CLI succeeds, including an
  iOS Simulator cross-build of the test targets:

  ```shell
  swift build --build-tests \
    --triple arm64-apple-ios18.0-simulator \
    --sdk "$(xcrun --sdk iphonesimulator --show-sdk-path)" \
    --enable-experimental-prebuilts
  ```

This suggests the incorrect testable-object linkage is in the Xcode/Swift
Build package build plan rather than the package's declared dependency graph.

## Environment

- macOS 26.5.2 (25F84), Apple silicon
- Xcode 26.6 (17F113), Swift 6.3.3
- Xcode 27 beta 4 (27A5228h), Swift 6.4
- swift-syntax 603.0.2

## Related reports

- Swift Forums: [Reposting Feedback for visibility: In certain configuration,
  the compiler is unable to use the pre-compiled SwiftCompilerPlugin and
  SwiftSyntaxMacros modules](https://forums.swift.org/t/84633), FB21914833. It
  reports the same target shape and `IDEPackageEnablePrebuilts=NO` workaround.
- Swift Forums: [Xcodebuild attempts to build macro package with wrong
  arch](https://forums.swift.org/t/76340).
- swift-build #73 fixed missing macro loading for dynamic package target
  variants during build-for-testing. This issue has a different symptom: the
  macro is loaded, but its macOS testable object is also linked into the iOS
  test bundle.

rjmansfield:
Fixes #1642.

### DIFF
--- a/Tests/SWBTaskConstructionTests/PackageProductConstructionTests.swift
+++ b/Tests/SWBTaskConstructionTests/PackageProductConstructionTests.swift
@@ -1392,4 +1392,67 @@ fileprivate struct PackageProductConstructionTests: CoreBasedTests {
             try await TaskConstructionTester(getCore(), testProject)
         }
     }
+
+    @Test(.requireSDKs(.macOS, .iOS), .requireXcode26())
+    func hostMacroObjectNotLinkedIntoCrossPlatformTestBundle() async throws {
+        let testProject = try await TestPackageProject(
+            "aProject",
+            groupTree: TestGroup(
+                "SomeFiles",
+                children: [
+                    TestFile("Macro.swift"),
+                    TestFile("Lib.swift"),
+                    TestFile("Tests.swift"),
+                ]),
+            buildConfigurations: [
+                TestBuildConfiguration("Debug", buildSettings: [
+                    "SWIFT_EXEC": swiftCompilerPath.str,
+                    "SWIFT_VERSION": swiftVersion,
+                    "GENERATE_INFOPLIST_FILE": "YES",
+                    "PRODUCT_NAME": "$(TARGET_NAME)",
+                    "CODE_SIGNING_ALLOWED": "NO",
+                    "SDKROOT": "auto",
+                    "SDK_VARIANT": "auto",
+                    "SUPPORTED_PLATFORMS": "macosx iphoneos iphonesimulator",
+                    "SUPPORTS_MACCATALYST": "YES",
+                    // Workaround for CI which have Intel hosts.
+                    "MACOSX_DEPLOYMENT_TARGET": "26.0",
+                ]),
+            ],
+            targets: [
+                TestStandardTarget(
+                    "LibTests", type: .unitTest,
+                    buildPhases: [
+                        TestSourcesBuildPhase(["Tests.swift"]),
+                        TestFrameworksBuildPhase([
+                            TestBuildFile(.target("Macro-testable")),
+                            TestBuildFile(.target("Lib")),
+                        ]),
+                    ],
+                    dependencies: ["Macro-testable", "Lib"]),
+                TestStandardTarget(
+                    "Lib", type: .objectFile,
+                    buildPhases: [TestSourcesBuildPhase(["Lib.swift"])]),
+                TestStandardTarget(
+                    "Macro-testable", type: .commonObject,
+                    buildConfigurations: [
+                        TestBuildConfiguration("Debug", buildSettings: ["SUPPORTED_PLATFORMS": "$(HOST_PLATFORM)"]),
+                    ],
+                    buildPhases: [TestSourcesBuildPhase(["Macro.swift"])]),
+            ])
+        let tester = try await TaskConstructionTester(getCore(), testProject)
+
+        for (destination, macroObjectLinked) in [(RunDestinationInfo.iOS, false), (.macCatalyst, true)] {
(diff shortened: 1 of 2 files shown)


rjmansfield:
@swift-ci test

Tests/SWBTaskConstructionTests/PackageProductConstructionTests.swift
```diff
@@ -1392,4 +1392,67 @@ fileprivate struct PackageProductConstructionTests: CoreBasedTests {
...
+                        ]),
+                    ],
+                    dependencies: ["Macro-testable", "Lib"]),
+                TestStandardTarget(
+                    "Lib", type: .objectFile,
+                    buildPhases: [TestSourcesBuildPhase(["Lib.swift"])]),
+                TestStandardTarget(
+                    "Macro-testable", type: .commonObject,
+                    buildConfigurations: [
+                        TestBuildConfiguration("Debug", buildSettings: ["SUPPORTED_PLATFORMS": "$(HOST_PLATFORM)"]),
```
owenv, line 1439: Does this match what you're seeing in PIF? I would not expect the testable variant of a macro to get this HOST_PLATFORM override. Based on the bug report it sounds like something in the SwiftSyntax prebuilt infrastructure might be applying it?

rjmansfield: Yes, in the PIF demoKitMacros has two targets and they both have SUPPORTED_PLATFORMS = ['$(HOST_PLATFORM)']

owenv: I think that may be the underlying bug. The macro itself should be built for the host platform, but the testable variant shouldn't have the override because it should be able to specialize for whatever platform the tests build for

rjmansfield: Actually, it looks like you've already fixed this in swiftpm. After https://github.com/swiftlang/swift-package-manager/pull/10320 the testable object is dropped.  I was using a 6.3.2 toolchain to reproduce. Do you think there's any value in keeping this check in swift-build or should I just abandon this PR?

owenv: I'm not a huge fan of this change if we don't absolutely need it, because I think it could hide real underlying problems if we introduce similar bugs in the future

rjmansfield: Ok, works for me.

PR #1675, opened 2026-08-26
Fix activeRunDestination_MultiPlatform_Target on Xcode 27

duow310:
The test asserted that a watchOS device run destination resolves ARCHS to `arm64_32`. Since the arm64_32 architecture spec in watchOSDevice.xcspec gained `DeploymentTargetRange = ( "0", "27" )`,
`Settings.addStandardArchitecturesOverride` constrains arm64_32 out of ARCHS_STANDARD at watchOS deployment targets of 27 or later, leaving its compatibility architecture `arm64`.

Derive the expected architecture from the watchOS SDK's default deployment target so the test asserts the exact value on either side of that cutoff.


duow310:
@swift-ci please test

owenv (approved):

PR #1678, opened 2026-08-27
Clear failed PIF transfer so it doesn't block later transfers

rjmansfield:
A PIF transfer that failed on the no-cache retransmission or audit-mismatch paths left the session's operation set so every later transfer failed with "unable to initiate PIF transfer session (operation in progress?)"

rdar://185918980


rjmansfield:
@swift-ci test

owenv (approved):

PR #1682, opened 2026-08-28
Skip sandbox violation scan for tasks that fail during setup

rjmansfield:
A task that fails during setup never launches its sandboxed subprocess, so scanning the log for sandboxing violations is unnecessary and expensive.

rdar://177373700


owenv (approved):

rjmansfield:
@swift-ci test

PR #1684, opened 2026-08-28
Fix host tool index tests

rjmansfield:
.anyMac became arm64-only in the macOS 27 so host tools VALID_ARCHS=$(HOST_ARCH) fails on Intel. Use .host, and set the deployment target under a macOS 27 SDK.


rjmansfield:
@swift-ci test

rjmansfield:
@swift-ci test

mhrawdon (approved):

PR #1695, opened 2026-09-02
Diagnose missing target gate nodes instead of crashing

rjmansfield:
Replace the force unwraps of `targetGateNodes[dependency]` in the target-ordering and index-preparation task producers with a guard that emits an error naming the offending edge.


rjmansfield:
@swift-ci test

owenv (approved):

PR #1696, opened 2026-09-02
Targets built with BUILD_VARIANTS should link against per-variant static dependencies

ndrewh:
I noticed that static library and object-file products always use the normal variant when linked into (non-normal-variant) dependents. The most prominent example of this is swift-package products, which appear to be typically object files.

The cause appears to be that `BuildFileResolution.resolveBuildFileReference` does not have any awareness of the caller's variant scope. This PR plumbs the caller's `MacroEvaluationScope` into `resolveBuildFileReference`, and then uses the caller's `CURRENT_VARIANT` when evaluating absolute paths to certain types of products.

This PR tries to deal with two types of products:
- Static archives, which are typically linked with `-l` options. These are covered by ld's `-image_suffix` option.
- Object files, which are typically passed via a link-list that contains absolute paths, and which aren't subject to any `ld` search logic. These require `swift-build` to pass the correct variant-aware absolute path to `ld`.

`ENABLE_VARIANT_AWARE_STATIC_LINKING` is a new (default-OFF) option that adds two new behaviors
1. Pass `-image_suffix` to ld when linking non-normal-variant products
2. Evaluate object file absolute paths in a variant-aware scope

Because `-image_suffix` silently falls back to the normal variant, this PR also introduces a warning (when ENABLE_VARIANT_AWARE_STATIC_LINKING is enabled) when a dependency is not building a variant that is built by a dependent. This warning can be disabled with `DISABLE_VARIANT_AWARE_STATIC_LINKING_FALLBACK_WARNING`.

Assisted-by: Claude

rdar://186490598


ndrewh:
@swift-ci please test

ndrewh:
@swift-ci please test

PR #1697, opened 2026-09-03
Don't pass -rdynamic when linking an object file

### LINKED ISSUE #573: Eliminate extraneous warnings when generating a test entry point with no tests
Needed to pass discovery_whenNoTests in SwiftPM when using Swift Build

tirdesh:
Fixes #573.

A package with no tests builds with one warning under `--build-system swiftbuild`, and none under `--build-system native`:

```
clang: warning: argument unused during compilation: '-rdynamic' [-Wunused-command-line-argument]
```

`-rdynamic` is passed to the partial link that produces the target's object file, where it does nothing: an `ld -r` link has no dynamic symbol table. This adds the `$(MACH_O_TYPE) != mh_object` guard already used elsewhere in `Ld.xcspec` to both the clang and swiftc variants.

`getTargetTaskOverrides` also suppresses this for objfile products, but not for SwiftPM package targets; guarding the spec covers every path that sets it.

The test covers both linker drivers and fails before the change.


owenv:
Closing as the PR description doesn't appear to match the contents

PR #1706, opened 2026-09-05
Limit concurrent nonblocking dependency graph construction

lynnswap:
## Summary

`NonBlockingComputeDependencyGraphMsg` currently starts an independent task for every request, allowing overlapping computations to retain multiple request contexts and graph states. This PR limits concurrent graph construction while keeping foreground requests independent of indexing work.

## Changes

- Limit index graph construction to one operation across the service process.
- Limit foreground graph construction to one operation per session, using queues independent of indexing and other sessions.
- Track each request through its terminal response. Session deletion requests cancellation and awaits owned tasks; quiescence also waits for their responses.
- Send synchronous cancellation responses for outstanding requests during teardown.

Only the nonblocking request path is limited; other graph-construction paths and normal build-task scheduling are unchanged. Queue depth and total memory usage are not capped.

## Measured impact — Release builds

In one 900-second run per variant with six concurrent Preview requests, observed peak physical footprint fell from **43.49 GiB to 7.14 GiB (83.6% lower)**. Both variants completed all six requests successfully.

```mermaid
xychart-beta
    title "Observed peak physical footprint — 900-second window"
    x-axis ["Unpatched", "Patched"]
    y-axis "GiB" 0 --> 50
    bar [43.49, 7.14]
```

| Metric | Unpatched | Patched | Observed change |
|---|---:|---:|---:|
| Peak physical footprint over 900 s | 43.49 GiB | 7.14 GiB | 83.6% lower |
| Physical footprint at 900 s | 42.85 GiB | 1.06 GiB | 97.5% lower |
| Time until all six Preview responses completed | 111.91 s | 98.19 s | 12.3% shorter |
| Service CPU time over 900 s | 2,594.25 CPU-s | 985.28 CPU-s | 62.0% lower |
| Average service CPU usage during the Preview request interval | 250.10% | 179.49% | 28.2% lower |

Compared Release builds (`-O`, whole-module optimization) of upstream `b2e24c9a875bc38370c9da36ea25bc2877522540` and this PR's production sources on Xcode 27.0 (27A5252f), Swift 6.4, macOS 26.6.2 (25G83), and 64 GiB RAM. Each run used fresh Xcode/service processes, the same workload, and restored DerivedData and UI state. Requests were submitted at 60 s; `proc_pid_rusage` was sampled every 100 ms for 900 s.

The patched run came first, and the observation window includes startup and background work. OS page cache and Simulator state were not reset; another Xcode instance remained open. CPU figures are service-only (100% = one core). Preview duration covers request submission through the final response, not graph construction alone.

## Testing

Validated locally on macOS with Swift 6.4:

- `swift test --disable-sandbox --force-resolved-versions`: passed with pinned dependencies, including existing known issues and SDK-conditional skips.
- All eight coordinator tests passed 100 repetitions each with Thread Sanitizer and `LIBDISPATCH_COOPERATIVE_POOL_STRICT=1`, with no sanitizer reports.
- Added coverage for cross-session admission, independent foreground progress, cancellation, exactly-once responses, and response completion during shutdown. The dependency-request integration test covers both `build` and `indexbuild` actions.
- `bash .github/scripts/format-check.sh` and diff-whitespace checks passed.

### DIFF
--- a/Sources/SWBBuildService/CMakeLists.txt
+++ b/Sources/SWBBuildService/CMakeLists.txt
@@ -16,6 +16,7 @@ add_library(SWBBuildService
   BuildServiceEntryPoint.swift
   ClientExchangeDelegate.swift
   DependencyGraphMessages.swift
+  DependencyGraphRequestCoordinator.swift
   DocumentationInfo.swift
   LocalizationInfo.swift
   Messages.swift
--- a/Sources/SWBBuildService/DependencyGraphMessages.swift
+++ b/Sources/SWBBuildService/DependencyGraphMessages.swift
@@ -121,16 +121,23 @@ struct NonBlockingComputeDependencyGraphMsg: MessageHandler {
             priority = .userInitiated
         }
 
-        _Concurrency.Task<Void, Never>(priority: priority) {
-            do {
-                let buildGraph = try await constructTargetBuildGraph(for: message.targetGUIDs, in: workspaceContext, buildParameters: message.buildParameters, includeImplicitDependencies: message.includeImplicitDependencies, dependencyScope: message.dependencyScope)
-                var adjacencyList: [TargetGUID: [TargetGUID]] = [:]
-                for configuredTarget in buildGraph.allTargets {
-                    adjacencyList[TargetGUID(rawValue: configuredTarget.target.guid), default: []].append(contentsOf: buildGraph.dependencies(of: configuredTarget).map { TargetGUID(rawValue: $0.target.guid) })
-                }
-                requestForReply.reply(DependencyGraphResponse(adjacencyList: adjacencyList))
-            } catch {
-                return requestForReply.reply(ErrorResponse("unable to compute dependency graph: \(error)"))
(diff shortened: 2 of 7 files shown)


owenv (changes requested):
In general one session should not block work requested of another session, and even within a session clients rely on being able to dispatch parallel requests. If the primary concern is memory use, it would be better to address that client-side. Also, the testing description in the PR description doesn't appear to be relevant to the code changes

lynnswap:
Thanks for taking the time to review this.

I agree that sessions should remain independent and that clients rely on parallel requests within a session. This memory issue should be addressed on the client side.
To clarify, the memory and CPU measurements in the PR description were collected while displaying #Previews in Xcode in our app project, which has a large dependency graph. I should have made that context clearer.

I'll close this PR for now and submit feedback to Apple, including a standalone reproduction project and the measurements, to request improvements to Xcode's handling of dependency graph requests.

lynnswap:
I’ve submitted feedback to Apple (FB24696968), proposing improvements to how Xcode handles dependency graph requests.

Thank you for the review and guidance.

PR #1707, opened 2026-09-05
[SwiftCompiler] When caching is enabled pass `-disable-cache-replay` to frontend

akyrtzi:
Cache replay is handled via SwiftScan APIs, the frontend is only invoked when there is a cache miss. Pass `-disable-cache-replay` so that the frontend just runs the compilation without checking the cache redundantly.

This also avoids the frontend emitting a warning if a compiler output ended up missing.


akyrtzi:
@swift-ci Please test

rjmansfield (approved):

akyrtzi:
@swift-ci Please test

PR #1710, opened 2026-09-08
Narrow ONLY_ACTIVE_ARCH for cross-platform simulator builds

rjmansfield:
Simulators run the host arch so don't build all archs.

rdar://182170059

### DIFF
--- a/Sources/SWBCore/Settings/Settings.swift
+++ b/Sources/SWBCore/Settings/Settings.swift
@@ -3950,7 +3950,8 @@ private class SettingsBuilder: ProjectMatchLookup, TripleLookup {
         let targetSupportedPlatforms = SettingsBuilder.targetSupportedPlatforms(scope: scope, core: core, runDestinationPlatform: destinationPlatform)
         let targetSupportsDestinationPlatform: Bool = targetSupportedPlatforms.contains { $0 === destinationPlatform }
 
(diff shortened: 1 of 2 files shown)


rjmansfield:
@swift-ci test

Sources/SWBCore/Settings/Settings.swift
```diff
@@ -3950,7 +3950,7 @@ private class SettingsBuilder: ProjectMatchLookup, TripleLookup {
         let targetSupportedPlatforms = SettingsBuilder.targetSupportedPlatforms(scope: scope, core: core, runDestinationPlatform: destinationPlatform)
         let targetSupportsDestinationPlatform: Bool = targetSupportedPlatforms.contains { $0 === destinationPlatform }
 
-        if !targetSupportsDestinationPlatform || runDestination.disableOnlyActiveArch {
+        if (!targetSupportsDestinationPlatform && !destinationPlatform.isSimulator) || runDestination.disableOnlyActiveArch {
```
owenv: Instead of checking only if the destination platform is a simulator, I think we need to check if it's the simulator platform corresponding to the destination platform. Normally SUPPORTED_PLATFORMS should have explicit entries for simulator platforms, but maybe there are some exceptions

rjmansfield:
@swift-ci test

PR #1711, opened 2026-09-08
Drop armv7k as a Swift module-only arch on watchOS

cyndyishida:
`armv7k` is retired on watchOS, so opt armv7k out via IncludeAsSwiftModuleOnlyArch, matching armv7/armv7s-ios.

Resolves: rdar://182837743


cyndyishida:
@swift-ci please test

owenv (approved):

PR #1713, opened 2026-09-08
Enable index-while-building with clang from a Swift.org toolchain

owenv:
Previously when enabling index-while-building for Clang we checked for an Apple Clang to determine compatibility. Also check for the indexUnitOutputPath feature in features.json, which will be present in an open source clang with the functionality.


owenv:
@swift-ci test

rjmansfield (approved):

PR #1725, opened 2026-09-10
Don't infer SWIFT_LIBRARY_LEVEL from SKIP_INSTALL over an explicit -l…

sepy97:
…ibrary-level

A target that passed -library-level spi itself compiled as ipi whenever SKIP_INSTALL was YES. That affects multiple projects in B&I.

Skip the inference when OTHER_SWIFT_FLAGS already sets the level. An assigned SWIFT_LIBRARY_LEVEL still takes precedence over the raw flag.

rdar://187128757


sepy97:
@swift-ci please test

owenv (approved):

PR #1726, opened 2026-09-10
Fix indexing preparation on windows

owenv:
We were computing the node name correctly when requesting index preparation, it should always use POSIX slashes


owenv:
@swift-ci test

owenv:
@swift-ci test

rjmansfield (approved):

PR #1727, opened 2026-09-10
Prefer a known framework install path over SKIP_INSTALL when inferrin…

sepy97:
…g SWIFT_LIBRARY_LEVEL

SKIP_INSTALL is automatically set to YES when INSTALL_PATH is empty. It doesn't work in reverse. Meaning, we should first check whether the module installed in known public/private location, only if not we perform IPI derivation

rdar://186946943


sepy97:
@swift-ci please test

nkcsgexi (approved):
LGTM!

owenv (approved):

PR #1728, opened 2026-09-10
Cherry picks PR 1465 to branch release/6.4.x

pmattos:
## Summary

Cherry picks PR #1465 (ie, rdar://176386125) to the `release/6.4.x` branch.

Without it, `SwiftCompilerSpec.generatePreviewInfo` matches the requested source file against the compile command's inputs by *exact* string. For standalone Swift packages whose sources live under a symlinked directory (eg, `/tmp` → `/private/tmp`), the new package PIF builder records the symlink-resolved spelling while the previews client asks to thunk the unresolved one. Nothing matches, `libSwiftDriver` returns no command line, and the client reports `noPreviewInfos`, so the preview never renders.

## Notes

The cherry pick did not apply cleanly though. Branch `release/6.4.x` carries the compilation caching prefix mapping work, so the thunk `vfsoverlay` is no longer emitted inside the output-file-map block; it moved down to the post-driver block, where it already keys the substitution on `selectedInputPath`. That hunk of the original commit is therefore already covered, and only the `selectedInputPath` computation and the output-file-map entry need to be applied.

## Tests 

Verified on `release/6.4.x`: `previewXOJITThunkInfoResolvesSymlinkedSourcePath` returns zero preview infos without the fix and exactly one with it. The full` PreviewsBuildOperationTests` suite passes, including `previewXOJITThunkInfoResolvesSymlinkedSourcePath` test from the original PR.

## Fixed Radar

- rdar://187133016

## Related Radars

- rdar://176386125
- rdar://179887248
- rdar://180730041


pmattos:
@swift-ci test

mhrawdon (approved):

PR #1729, opened 2026-09-11
Support injecting extra build variants, with opt-out setting

ndrewh:
I want to be able to easily build all (or as many as possible) of any project's products with a sanitizer enabled. It would be nice if there was a single setting that could be used to inject additional variants into any existing build, and a way for projects to opt-out from this mechanism (i.e. because they know that specific targets are incompatible with a specific sanitizer).

This PR adds a generic mechanism for injecting extra BUILD_VARIANTS into a build, and an opt-out mechanism so that individual targets can declare that they do not support building specific variants.

- BUILD_VARIANTS is concatenated with a new setting EXTRA_BUILD_VARIANTS to produce the effective list of variants.
- Support for individual variants will be indicated by the SUPPORTS_VARIANT_<variant name> boolean setting.
  - XCBuild will not inject the variant unless the setting evaluates to boolean true.
  - This will be defaulted to true by SDKSettings for some SDKs and some sanitizers. This does not mean that the variant is automatically injected, but makes it eligible for injection with EXTRA_BUILD_VARIANTS.

rdar://187161036


ndrewh:
@swift-ci please test

owenv (approved):
I think this is reasonable, and it seems relatively self-contained

PR #1732, opened 2026-09-11
Track xcconfig override files as settings inputs

rjmansfield:
Command line and environment override files weren't added to inputPathsAffectingSettings, so editing their contents didn't invalidate the settings cache.


rjmansfield:
@swift-ci test

owenv (approved):

PR #1733, opened 2026-09-11
Simplify logic for copying XCTest frameworks

harlanhaskins:
This simplifies the logic for the "Copy XCTest Products" phase resolving public/private frameworks.

The previous logic hard-coded the locations of each of the set of frameworks, but for extra flexibility (and since this is called so infrequently as of recent Xcodes) let's just apply the public-and-then-private check for all frameworks.

rdar://187211213


harlanhaskins:
@swift-ci test

owenv (approved):

stmontgomery (approved):

PR #1737, opened 2026-09-14
Depend only on actual copied headers for InstallAPI

rjmansfield:
Track the headers the Copy Headers phase actually produces and drop installapi dependencies on public/private destinations that aren't produced.

rdar://186950689


rjmansfield:
@swift-ci test

Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/ProductPostprocessingTaskProducer.swift
```diff
@@ -802,6 +802,17 @@ func addCommonInstallAPITasks(_ producer: PhasedTaskProducer, _ scope: MacroEval
                               tapiOutputNode: PlannedPathNode, tapiOrderingNode: PlannedVirtualNode, phaseStartNodes: [any PlannedNode],
                               phaseEndTask: any PlannedTask, jsonPath: Path?, destination: InstallAPIDestination) async -> [any PlannedTask] {
     let buildComponents = scope.evaluate(BuiltinMacros.BUILD_COMPONENTS)
+
+    // Drop installapi dependencies on public/private destinations that aren't produced.
+    let producedHeaders = producer.context.producedHeaderPaths()
+    let publicHeaderDir = TargetHeaderInfo.destDirPath(for: .public, scope: scope).normalize()
+    let privateHeaderDir = TargetHeaderInfo.destDirPath(for: .private, scope: scope).normalize()
+    let headerDependencyInputs = headerDependencyInputs.filter { node in
```
owenv, line 810: I think headerDependencyInputs may contain the json file list passed to tapi, which wouldn't be in any of the header directories, is this going to filter it out?

rjmansfield: The json is passed separately as fileListPath not in headerDependencyInputs, so it won't get filtered out. Locally I added an assertion to the new tests to confirm the json doesn't get filtered out. I can amended that to the PR if you'd prefer

rjmansfield:
@swift-ci test

owenv (approved):

PR #1739, opened 2026-09-15
Skip module-only swiftmodules when a target is skipped for installAPI

rjmansfield:
During installAPI, if a target that doesn't produce a module has its compile skipped so the module-only path should do the same.

rdar://187177000


rjmansfield:
@swift-ci test

owenv (approved):

PR #1743, opened 2026-09-16
Isolate stateful task actions per build engine to prevent data races

nkcsgexi:
Build descriptions memoize task-action instances and are cached and shared across llbuild build engines that execute concurrently. SwiftDriverJobSchedulingTaskAction (and its only concrete subclass, SwiftDriverCompilationRequirementTaskAction) carries a mutable `state` backed by Set<UInt>. When the same memoized instance is driven by two engines at once, concurrent mutation races on the set's copy-on-write buffer refcount and over-releases it: the buffer is freed while still referenced, its slot is recycled, and dereferencing the stale pointer crashes with EXC_BAD_ACCESS.

Locking would not suffice, since one engine's taskSetup/reset would logically clobber another engine's in-flight scheduling state. Instead, isolate: add copyForConcurrentExecution() on TaskAction (default nil, meaning safe to share) and have the stateful action vend a fresh instance. InProcessTool.createCommand takes that per-engine copy for static tasks; dynamic tasks already build fresh actions and were unaffected.


nkcsgexi:
@swift-ci test

mhrawdon (approved):

PR #1745, opened 2026-09-16
Emit framework symlinks to Versions/Current/ per-variant

ndrewh:
Currently, a symlink is produced from `X.framework/X` to `X.framework/Versions/Current/X`. However, a symlink is not produced for variant dylibs e.g. `X.framework/X_debug` to `X.framework/Versions/Current/X_debug`.

This is problematic, since some code (e.g. [`[NSBundle executablePath]`](https://developer.apple.com/documentation/foundation/bundle/executablepath?language=objc)) references `X.framework/X` -- and then some users pass this path to `dlopen`. When `DYLD_IMAGE_SUFFIX=_debug` is set, `dyld` looks for `X.framework/X_debug`, which does not exist. It then falls back to `X.framework/X`, which is not the desired behavior.

This change emits symlinks for every variant.

rdar://179734460


ndrewh:
@swift-ci please test

mhrawdon:
I'm skeptical that this is desirable. Support for build variants has been largely unchanged for decades - why is this an issue now? Has it been broken for all this time? My understanding is that `dyld` transparently looks up the variant based on the suffix during normal location and loading of framework binaries.

But that isn't the case with `dlopen()`. If clients are using `dlopen()` then it's incumbent on them to handle `DYLD_IMAGE_SUFFIX` manually.

Basically, build variants are not a mainline workflow, and using it for tricky behaviors puts the onus on the code implementing those behaviors.

ndrewh:
> But that isn't the case with dlopen(). If clients are using dlopen() then it's incumbent on them to handle DYLD_IMAGE_SUFFIX manually.

This is not true. DYLD_IMAGE_SUFFIX is applied to libraries that are loaded with dlopen.

> Has it been broken for all this time?

I believe so. The particular caller of dlopen recently changed.

ndrewh:
I think we are going to try to fix this a different way.

PR #1751, opened 2026-09-18
Indexing: record explicit-module inputs during index-build preparation

nkcsgexi:
Add an EnableSwiftExplicitModulesInIndexBuild feature flag (off by default) that enables Swift explicit modules in the index/preparation build arena. When enabled, the compilation-requirements task action persists a sidecar recording the resolved invocation so that background indexing can later reuse the modules preparation built instead of re-scanning and rebuilding them.


nkcsgexi:
@swift-ci please test

Sources/SWBCore/Settings/Settings.swift
```diff
@@ -4206,8 +4206,14 @@ private class SettingsBuilder: ProjectMatchLookup, TripleLookup {
             table.push(BuiltinMacros.EFFECTIVE_PLATFORM_NAME, literal: MacCatalystInfo.publicSDKBuiltProductsDirSuffix)
         }
 
-        table.push(BuiltinMacros.SWIFT_ENABLE_EXPLICIT_MODULES, literal: .disabled)
-        table.push(BuiltinMacros._EXPERIMENTAL_SWIFT_EXPLICIT_MODULES, literal: .disabled)
+        if SWBFeatureFlag.enableSwiftExplicitModulesInIndexBuild.value {
+            table.push(BuiltinMacros.SWIFT_ENABLE_EXPLICIT_MODULES, literal: .enabled)
```
owenv, line 4210: I think I'd prefer to not force-enable in this case, and instead just skip the disablement and allow the project config to take over

Sources/SWBTaskExecution/TaskActions/SwiftDriverCompilationRequirementTaskAction.swift
```diff
@@ -38,6 +39,42 @@ final public class SwiftDriverCompilationRequirementTaskAction: SwiftDriverJobSc
         !driverPayload.eagerCompilationEnabled
     }
 
+    public override func performTaskAction(_ task: any ExecutableTask, dynamicExecutionDelegate: any DynamicTaskExecutionDelegate, executionDelegate: any TaskExecutionDelegate, clientDelegate: any TaskExecutionClientDelegate, outputDelegate: any TaskOutputDelegate) async -> CommandResult {
+        let result = await super.performTaskAction(task, dynamicExecutionDelegate: dynamicExecutionDelegate, executionDelegate: executionDelegate, clientDelegate: clientDelegate, outputDelegate: outputDelegate)
+        if result == .succeeded {
+            writeIndexExplicitModuleInfo(task, dynamicExecutionDelegate: dynamicExecutionDelegate, executionDelegate: executionDelegate, outputDelegate: outputDelegate)
```
owenv, line 45: I wonder if SwiftDriverTaskAction should be responsible for this since that's where the dependency scan is happening, WDYT?

Sources/SWBCore/Settings/Settings.swift
nkcsgexi: Makes sense! Just updated this.

Sources/SWBTaskExecution/TaskActions/SwiftDriverCompilationRequirementTaskAction.swift
nkcsgexi: I think it's a reasonable move since we don't need anything particularly from the requirement task action for this. I've just moved it.

nkcsgexi:
@swift-ci please test

Sources/SWBTaskExecution/TaskActions/SwiftDriverTaskAction.swift
```diff
@@ -158,4 +162,31 @@ final public class SwiftDriverTaskAction: TaskAction, BuildValueValidatingTaskAc
...
+            guard let job = plannedBuild.compilationRequirementsPlannedDriverJobs().first else { return }
+            let commandLine = job.driverJob.commandLine.map { $0.asString }
+            // Only record when the scan actually resolved an explicit module map, i.e. there is a handoff worth making.
+            guard commandLine.contains("-explicit-swift-module-map-file") else { return }
+            let info = IndexExplicitModuleInfo(uniqueID: driverPayload.uniqueID, resolvedArguments: commandLine)
+
+            let encoder = JSONEncoder()
+            encoder.outputFormatting = [.sortedKeys, .prettyPrinted]
+            let contents = try ByteString(encoder.encode(info))
+            let path = driverPayload.indexExplicitModuleInfoPath
```
owenv: Can this just be nil if not using explicit modules, instead of the ` guard commandLine.contains("-explicit-swift-module-map-file") else { return }` check above?

Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift
```diff
@@ -465,6 +465,39 @@ public struct SwiftDriverPayload: Serializable, TaskPayload, Encodable {
...
+    /// background indexing can reuse prep's already-built modules instead of re-scanning and rebuilding them.
+    ///
+    /// Derived purely from fields shared by the writer (the prep task action) and the reader
+    /// (`generateIndexingInfo`), so both agree on the path without any additional plumbing.
+    public var indexExplicitModuleInfoPath: Path {
+        explicitModulesTempDirPath.join("\(moduleName)-\(slice).index-explicit-modules.json")
+    }
+}
+
+/// A faithful record of the compilation-requirements invocation preparation resolved, consumed by background indexing.
```
owenv: most of this comment doesn't seem very useful

owenv (approved):

nkcsgexi:
@swift-ci please test

PR #1752, opened 2026-09-18
Refactor how caching settings on Xcode targets are imposed on package dependencies

owenv:
Previously, if an Xcode target enabled Swift caching, it would impose this preference on package dependencies. Refactor this mechanism and extend it to cover all caching settings. If two dependents of a package disagree, the boolean settings are OR-ed, if the path/prefix mapping settings disagree we pick a winner and warn.


owenv:
@swift-ci test

owenv:
@swift-ci test

rjmansfield (approved):

PR #1756, opened 2026-09-20
[MockToolchainCASPlugin] Make it loadable from the tests when built as an Xcode project target

akyrtzi:


akyrtzi:
@swift-ci Please test

owenv (approved):

